PR #19, opened 2022-12-06
Correction on readme for development setup

jubalm:
The `vendor` directory is not present on clone and errors out during `Load unpacked` step during setup on Chrome. Running `npm run vendor` before build fixes the issue.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -29,7 +29,7 @@ Install:
 `npm ci --ignore-scripts`
 
 Build:
-`npm run build`
+`npm run vendor && npm run build`
 
 Then depending on your browser:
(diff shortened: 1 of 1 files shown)


MicahZoltu (approved):
It is also missing the build step for the inpage.  We can merge this as is though as it is strictly better than current instructions!

KillariDev:
> 

this is not a mandatory step, as the js code's are in the repo. Its mandatory if you change those files thought.

MicahZoltu:
`vendor` folder isn't checked in, it is in the [.gitignore](https://github.com/DarkFlorist/TheInterceptor/blob/main/extension/.gitignore)

Issue #3, opened 2022-12-05
Clicking one side of the Simulating/Signing toggle shouldn't toggle, but set.


MicahZoltu:
If you click on the left side of the Simulating | Signing toggle, it will switch you to Simulating mode.  If you click in the same spot again it will switch you back to Signing.  Given the way we have this element designed, I think clicking on one side should set the state to that side, rather than toggle.  So to switch back to Signing you would need to click on the Signing side.

Alternatively, I think if we change the design so the words are outside of the toggle and the toggle is a UI element between them, then it would be more intuitive to have the toggle flip left/right with each click.

KillariDev:
fixed

PR #20, opened 2022-12-06
Change simulation and signing controls UI behavior

jubalm:
Updated the simulation and signing controls to behave as per @MicahZoltu's description in issue #3. This change will require the user to deliberately click on the target mode to initiate the switch instead of toggling between the two when either side of the control is clicked.

Used built-in classes to get the design as close to the original without style attribute overrides

Screenshot 2022-12-07 at 5 08 53 AM

### DIFF
--- a/extension/app/ts/components/pages/Home.tsx
+++ b/extension/app/ts/components/pages/Home.tsx
@@ -34,17 +34,13 @@ function FirstCard(param: FirstCardParams) {
 						<img src = { param.tabIcon } />
 					</span>
 				</div>
-				<div class = 'card-header-title'>
-					<input type = 'checkbox' id = 'toggle' style = 'display: none;' checked = { !param.simulationMode } class = 'toggleCheckbox' onInput = { e => { if (e.target instanceof HTMLInputElement && e.target !== null) { param.enableSimulationMode(!e.target.checked) } } } />
-					<label for = 'toggle' class = 'toggleContainer'>
-						<div style = 'font-weight: normal' >Simulating</div>
-						<div style = 'font-weight: normal;' >
-							<SignerLogoText
-								signerName = { param.signerName }
-								text = { 'Signing' }
-							/>
-						</div>
-					</label>
+				<div class = 'card-header-title px-0 is-justify-content-center'>
+					<div class='buttons has-addons'>
+						<button class={ `button is-primary${ param.simulationMode ? '' : ' is-outlined'}` } onClick={ () => param.enableSimulationMode(true) }>Simulating</button>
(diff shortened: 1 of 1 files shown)


extension/app/ts/components/pages/Home.tsx
```diff
@@ -34,17 +34,13 @@ function FirstCard(param: FirstCardParams) {
...
-								text = { 'Signing' }
-							/>
-						</div>
-					</label>
+				<div class = 'card-header-title px-0 is-justify-content-center'>
+					<div class='buttons has-addons'>
+						<button class={ `button is-primary${ param.simulationMode ? '' : ' is-outlined'}` } onClick={ () => param.enableSimulationMode(true) }>Simulating</button>
+						<button class={ `button is-primary${ param.simulationMode ? ' is-outlined' : ''}` } onClick={ () => param.enableSimulationMode(false) }>
+							<SignerLogoText signerName = { param.signerName } text = { 'Signing' } />
+						</button>
```
MicahZoltu, line 42: Consider this just for consistency within this component:
```suggestion
						<button class={ `button is-primary${ param.simulationMode ? '' : ' is-outlined'}` } onClick={ () => param.enableSimulationMode(true) }>
							Simulating
						</button>
						<button class={ `button is-primary${ param.simulationMode ? ' is-outlined' : ''}` } onClick={ () => param.enableSimulationMode(false) }>
							<SignerLogoText signerName = { param.signerName } text = { 'Signing' } />
						</button>
```
Alternatively:
```suggestion
						<button class={ `button is-primary${ param.simulationMode ? '' : ' is-outlined'}` } onClick={ () => param.enableSimulationMode(true) }>Simulating</button>
						<button class={ `button is-primary${ param.simulationMode ? ' is-outlined' : ''}` } onClick={ () => param.enableSimulationMode(false) }><SignerLogoText signerName = { param.signerName } text = { 'Signing' } /></button>
```

MicahZoltu (commented):
Will leave it to Killari to decide on this.

KillariDev:
I aligned the style to be closer, removed some old css that is no longer needed and then I also think the button shouldn't do anything if it doesn't change things

KillariDev:
image

MicahZoltu:
FWIW I like Jubal's styling a bit better.  😬  However, I agree that we should be consistent everywhere.

----

We should also mark the currently selected option as "disabled" so it doesn't get a cursor that hints you can click it.

KillariDev:
I don't think this is a consistency issue as this is the only place we use this kind of element

KillariDev:
image

now looking like this

MicahZoltu (approved):

Issue #8, opened 2022-12-06
Transaction preview/simulation UI should be responsive.


MicahZoltu:
If I resize the popup window horizontally, the UI remains fixed in the center, with a lot of the UI elements crammed into a limited amount of space.

Just removing the width/max-width on the `body` element seems to resolve the issue for the most part.  We could do a bit better with layout, but it is an improvement over current.

PR #25, opened 2022-12-07
Remove preset window sizes

KillariDev:
You are right, things seem to work reasonably well with these changes. They didn't use to work that well before :)

Fixes https://github.com/DarkFlorist/TheInterceptor/issues/8

image

### DIFF
--- a/extension/app/css/interceptor.css
+++ b/extension/app/css/interceptor.css
@@ -126,7 +126,7 @@ input[type = 'checkbox']:checked::before {
 }
 
 html, body {
-	height: 100%;
+	height: fit-content
 }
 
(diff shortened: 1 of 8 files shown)


MicahZoltu:
Hmm, why don't the arrows line up?  I think that should be a `display: grid` with 7 columns:
```
<address><1fr><arrow><1fr><address><1fr><amount>
```
This should result in all of the arrows lining up, and space evenly distributed.  Alternatively, we could do:
```
<1fr><address><arrow><address><amount><1fr>
```
That should also make everything line up, but move the excess spacing to the outside.

MicahZoltu (approved):

PR #27, opened 2022-12-07
add -ignore-scripts flag and remove addressData command as its no longer used

KillariDev:
Fixes https://github.com/DarkFlorist/TheInterceptor/issues/22

### DIFF
--- a/extension/package.json
+++ b/extension/package.json
@@ -3,6 +3,7 @@
 	"private": true,
 	"type": "module",
 	"scripts": {
+		"setup": "npm run vendor && npm run inpage && npm run build",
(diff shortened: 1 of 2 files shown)


MicahZoltu (approved):
We may also want to consider adding a script like `setup: npm run vendor && npm run inpage && npm run build` to simplify the instructions in the readme.

KillariDev:
> We may also want to consider adding a script like `setup: npm run vendor && npm run inpage && npm run build` to simplify the instructions in the readme.

Running vendor and inpage scripts is often not necessary, just adds to the compile time. But I guess it could be useful to have one command that does everything

Issue #21, opened 2022-12-07
Readme instructions should include inpage build.


MicahZoltu:
#19 adds build step for vendor, but is still missing inpage build step.

I wonder if we should create a script for first build that runs all of this?

PR #28, opened 2022-12-07
Add instructions about inpage

KillariDev:
Fixes https://github.com/DarkFlorist/TheInterceptor/issues/21


MicahZoltu (changes requested):
I don't think we should be committing generated code.  Instead, we should be building it locally from whatever is committed.  It is *way* too easy to end up with out of sync source files and generated files.

See my suggestion in other issue about creating a `setup` NPM script that does all of the initial setup in one command so our readme can be simpler.

Issue #15, opened 2022-12-06
Brave - refused to execute inline script because it violates the following Content Security Policy directive


randomishwalk:
Brave version: [Version 1.46.134 Chromium: 108.0.5359.94 (Official Build) (64-bit)](https://brave.com/latest/)

> Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'self' blob: 'unsafe-eval'". Either the 'unsafe-inline' keyword, a hash ('sha256-Pno6VqOVhlRXSd0eVB22rNathycbNhTyHseqef92DaU='), or a nonce ('nonce-...') is required to enable inline execution.

```html

<!DOCTYPE html>
<html>
	<head>
		<meta charset = 'utf-8'>
		<link rel = 'icon' type = 'image/x-icon' href = 'favicon.ico'>
	</head>
	<body style = 'width: 520px; height: 400px;'>
		<script async type = 'module' src = '../vendor/es-module-shims/es-module-shims.wasm.js'></script>
		<script type = 'importmap'>
		{
			"imports": {
				"ethers": "../vendor/ethers/ethers.esm.js",
				"webextension-polyfill": "../vendor/webextension-polyfill/browser-polyfill.js",
				"es-module-shims": "../vendor/es-module-shims/es-module-shims.js",
				"preact": "../vendor/preact/preact.module.js",
				"preact/jsx-runtime": "../vendor/preact/jsx-runtime/jsxRuntime.module.js",
				"preact/hooks": "../vendor/preact/hooks/hooks.module.js",
				"funtypes": "../vendor/funtypes/index.mjs",
				"node-fetch": "../vendor/node-fetch/index.mjs",
				"@zoltu/ethereum-abi-encoder": "../vendor/@zoltu/ethereum-abi-encoder/index.js",
				"@zoltu/ethereum-crypto": "../vendor/@zoltu/ethereum-crypto/index.js",
				"@zoltu/rlp-encoder": "../vendor/@zoltu/rlp-encoder/index.js",
				"@darkflorist/address-metadata": "../vendor/@darkflorist/address-metadata/index.js"
			}
		}
	</script>

		<main>Loading...</main>

		<script type = 'module' src = '../js/background/background.js'></script>
	</body>
</html>

```
![image](https://user-images.githubusercontent.com/94845277/205971006-85b79ffb-4596-442b-9aeb-4295f8ab4752.png)

MicahZoltu:
[...]

PR #33, opened 2022-12-08
Fixes inline error by requiring hash and not using module type

KillariDev:
Fixes https://github.com/DarkFlorist/TheInterceptor/issues/15

The downside of this is, that when the vendor stuff is changed, the hash changes and developer needs to remember to update the hash to `content_security_policy`. Is there a better way for this?

### DIFF
--- a/extension/app/html/background.html
+++ b/extension/app/html/background.html
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
 	<body>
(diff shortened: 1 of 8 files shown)


MicahZoltu:
Can we have the vendor script generate that hash and write it into `manifest.json`?  The vendor script already writes into `index.html`, so we already have the infrastructure to do "mutate file on run".

Also, doesn't the es-module-shims page specifically say to do `async type = 'module'`?  I'm worried that this change may break something else.

Issue #17, opened 2022-12-06
Brave - Error: Extension context invalidated.


randomishwalk:
Brave version: [Version 1.46.134 Chromium: 108.0.5359.94 (Official Build) (64-bit)](https://brave.com/latest/)

![image](https://user-images.githubusercontent.com/94845277/205972545-2d9d4577-a76f-4fff-abda-06106d649b30.png)

```js
"use strict";
function listenInContentScript() {
    /**
     * this script executed within the context of the active tab when the user clicks the extension bar button
     * this script serves as a _very thin_ proxy between the page scripts (dapp) and the extension, simply forwarding messages between the two
    */
    // the content script is a very thin proxy between the background script and the page script
    const extensionPort = browser.runtime.connect();
    // forward all message events to the background script, which will then filter and process them
    window.addEventListener('message', messageEvent => {
        try {
            // we only want the data element, if it exists, and postMessage will fail if it can't clone the object fully (and it cannot clone a MessageEvent)
            if (!('data' in messageEvent))
                return;
            extensionPort.postMessage({ data: messageEvent.data });
        }
        catch (error) {
            // CONSIDER: should we catch data clone error and then do `extensionPort.postMessage({data:JSON.parse(JSON.stringify(messageEvent.data))})`?
            console.error(error);
        }
    });
    // forward all messages we get from the background script to the window so the page script can filter and process them
    extensionPort.onMessage.addListener(response => {
        try {
            window.postMessage(response, '*');
        }
        catch (error) {
            console.error(error);
        }
    });
}
function injectScript(content) {
    try {
        const container = document.head || document.documentElement;
        const scriptTag = document.createElement('script');
        scriptTag.setAttribute('async', 'false');
        scriptTag.textContent = content;
        container.insertBefore(scriptTag, container.children[0]);
        container.removeChild(scriptTag);
        listenInContentScript();
    }
    catch (error) {
        console.error('Interceptor: Provider injection failed.', error);
    }
}
const injected_ts = `"use strict";
const METAMASK_ERROR_USER_REJECTED_REQUEST = 4001;
const METAMASK_ERROR_CHAIN_NOT_ADDED_TO_METAMASK = 4902;
class InterceptorFuture {
    constructor() {
        this.then = (onfulfilled, onrejected) => {
            return this.promise.then(onfulfilled, onrejected);
        };
        this.resolve = (value) => {
            this.resolveFunction(value);
        };
        this.reject = (reason) => {
[...]

PR #34, opened 2022-12-08
Silence "Extension context invalidated" -error

KillariDev:
Fixes https://github.com/DarkFlorist/TheInterceptor/issues/17

### DIFF
--- a/extension/app/inpage/output/injected_document_start.js
+++ b/extension/app/inpage/output/injected_document_start.js
@@ -16,6 +16,12 @@ function listenInContentScript() {
         }
         catch (error) {
             // CONSIDER: should we catch data clone error and then do `extensionPort.postMessage({data:JSON.parse(JSON.stringify(messageEvent.data))})`?
+            if (error instanceof Error) {
+                if (error.message?.includes('Extension context invalidated.')) {
+                    // this error happens when the extension is refreshed and the page cannot reach The Interceptor anymore
+                    return;
+                }
+            }
             console.error(error);
         }
     });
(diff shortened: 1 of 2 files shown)


extension/app/inpage/output/injected_document_start.js
```diff
@@ -16,6 +16,12 @@ function listenInContentScript() {
         }
         catch (error) {
             // CONSIDER: should we catch data clone error and then do `extensionPort.postMessage({data:JSON.parse(JSON.stringify(messageEvent.data))})`?
+            if (error instanceof Error) {
+                if (error.message?.includes('Extension context invalidated.')) {
+                    // this error happens when the extension is refreshed and the page cannot reach The Interceptor anymore
+                    return;
+                }
+            }
             console.error(error);
```
MicahZoltu, line 25: Should this still be `console.error`, or should we rethrow it here?

MicahZoltu (approved):
Approved if swallowing the error is still the correct behavior instead of rethrowing.

extension/app/inpage/output/injected_document_start.js
KillariDev: I am not sure what other errors could occur. Maybe rethrow, and then it would be correctly logged as unhandled exception

MicahZoltu: My general exception handling policy is to handle exceptions anywhere you can fully recover regardless of failure.  This leads to my exception blocks always being incredibly generic, and I often have just a few in the whole application.  The idea is that you never know in advance what kinds of failures you will have, so it is best to handle all failures in as few places as possible, where you can spend a lot of time thinking about exactly how to deal with a particular situation.

In this case, we have one specific error that we intentionally want to ignore, but for all other possible errors we should fall back to our generalized exception handling (i.e., rethtrow).

TL;DR: I agree.

PR #39, opened 2022-12-09
Stop messaging after disconnect

KillariDev:


MicahZoltu (approved):
Having this state machine feels dirty to me.  Would be nice if we could encapsulate it somehow.  However, that isn't a blocker for this PR.

KillariDev:
> Having this state machine feels dirty to me. Would be nice if we could encapsulate it somehow. However, that isn't a blocker for this PR.

and I have been working on adding one similar one elsewhere :D. yah, I need to think if we can get rid of these

PR #41, opened 2022-12-09
Update typescript to 4.9.3 and @types/node

KillariDev:


extension/inpage/package.json
```diff
@@ -1,11 +1,11 @@
 {
 	"type": "module",
 	"devDependencies": {
-		"@types/node": "16.11.7",
+		"@types/node": "18.11.12",
```
MicahZoltu, line 4:
```suggestion
```
`inpage` shouldn't have a dependency on `@types/node` as it is a browser targeted project.

extension/package.json
```diff
@@ -23,10 +23,9 @@
 	},
 	"devDependencies": {
 		"@types/firefox-webext-browser": "94.0.1",
-		"@types/node": "16.11.40",
+		"@types/node": "18.11.12",
```
MicahZoltu, line 26:
```suggestion
```
This project shouldn't have a dependency on this as it is a web targeted project.

MicahZoltu (approved):
@types/node should be removed from two of the projects, but can be in a separate PR.

extension/inpage/package.json
KillariDev: It's node project that compiles JavaScript files for the main project

MicahZoltu: Oh, I see.  Ignore this one then.  Still should remove `@node/types` from the extension itself though.

PR #42, opened 2022-12-09
on port disconnect, stop sending from background

KillariDev:


extension/app/ts/background/background.ts
```diff
@@ -428,19 +433,23 @@ async function onContentScriptConnected(port: browser.runtime.Port) {
 		if (!window.interceptor.websiteTabIcons.has(tabId)) {
 			updateExtensionIcon(port)
 			port.onDisconnect.addListener(() => {
-				updateExtensionIcon(port)
```
MicahZoltu, line 431: Why aren't we doing this anymore?

extension/app/ts/background/background.ts
```diff
@@ -494,23 +503,23 @@ async function onContentScriptConnected(port: browser.runtime.Port) {
...
-			port.postMessage({
-				interceptorApproved: false,
-				requestId: request.requestId,
-				options: request.options,
-				error: {
-					code: 123456,
-					message: error
-				}
-			})
+			if (error instanceof Error) {
```
MicahZoltu, line 512: I feel like we should be propagating the error even if someone decided to throw a non-error.  Consider something like this for ensuring that we are always dealing with an `Error` type (adjusted as necessary to fit this use case).
```ts
error = (error instanceof Error) ? error
	: (typeof error === 'string') ? new Error(error)
	: new Error(JSON.stringify(error))
```

MicahZoltu (approved):
I would like to see the `if (error instanceof Error)` thing addressed.  I think swallowing errors like that is pretty dangerous.  At the least we should be logging them somewhere.

extension/app/ts/background/background.ts
KillariDev: this is because the port is closed, which means that the tab is closed, no need to update extension icon for such tab as it doesn't exist anymore

PR #50, opened 2022-12-15
Transaction view update

KillariDev:
- uses grids instead of texts to display things
- use blockies for unknown tokens
- approving 0 is now positive event (appears green)
- quite a bit refactoring for types and for how coins.tsx modules work (now they don't determine what is positive and what negative anymore)

image
image
image


extension/app/css/interceptor.css
```diff
@@ -450,3 +454,30 @@ svg.spinner > circle {
...
+
+.log-cell {
+	text-overflow: ellipsis;
+	overflow: hidden;
+	white-space: nowrap;
+	display: flex;
+}
+.ellipsis {
+	text-overflow: ellipsis;
+    overflow: hidden;
```
MicahZoltu, line 482: Weird indentation.

KillariDev: fixed

extension/app/ts/components/subcomponents/icons.tsx
```diff
@@ -0,0 +1,25 @@
+export function ArrowIcon(param: { color: string }) {
+	return <>
+		&nbsp;
```
MicahZoltu, line 3: Feels like we should be using styling here rather than `&nbsp;`?  e.g., `padding`

MicahZoltu: Same below and I think I saw another one potentially above.

MicahZoltu (approved):
This PR is huge, so approving.

PR #57, opened 2022-12-16
Fixes mew

KillariDev:
Fixes https://github.com/DarkFlorist/TheInterceptor/issues/51

- Fixes bug that Interceptor only supports single `window.ethereum.on()` callbacks, now it supports arbitrary many
- Add `window.ethereum.removeListener` to remove callback functions
- Adds `eth_getTransactionCount`
- When DApp asks to switch chain, and we are already on that chain, do not ask anything from user and just return

### DIFF
--- a/extension/app/inpage/ts/inpage.ts
+++ b/extension/app/inpage/ts/inpage.ts
@@ -87,17 +87,23 @@ interface ProviderMessage {
 	readonly data: unknown
 }
 
+type AnyCallBack =  ((message: ProviderMessage) => void)
+	| ( (connectInfo: ProviderConnectInfo) => void )
+	| ( (accounts: string[]) => void )
+	| ( (error: ProviderRpcError) => void )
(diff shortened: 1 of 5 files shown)


extension/app/inpage/ts/inpage.ts
```diff
@@ -331,25 +332,48 @@ function injectEthereumIntoWindow() {
...
+				window.interceptor.onChainChangedCallBacks.add(callback as (chainId: string) => void)
+				return
+			default:
+		}
+	}
+
+	const removeListener = async (kind: OnMessage, callback: any) => {
+		switch (kind) {
+			case 'accountsChanged':
+				window.interceptor.onAccountsChangedCallBacks.delete(callback)
```
MicahZoltu: Note: I'm still not a fan of using globals and would like to see something like dependency injection via function parameters used instead.

MicahZoltu (approved):

Issue #52, opened 2022-12-15
Arrow bounces when window is resized


KillariDev:
![image](https://user-images.githubusercontent.com/13102010/207903750-4f059f7a-8273-491a-8196-ea21e25eec40.png)

DecentralisedTech:
Was not able to reproduce in both Firefox and Brave. Is this still an issue with the arrow?
The resizing of the inner light grey card could be replaced with some flex positioning though.

KillariDev:
yah, seems like that is no longer issue. But the container is limited by some max size, I guess this is what you mean by the flex positioning?

DecentralisedTech:
Yes exactly

PR #67, opened 2022-12-20
don't limit container size

KillariDev:
Fixes https://github.com/DarkFlorist/TheInterceptor/issues/52

Now container can resize freely without being restricted to max-widths depending on the window size

image


MicahZoltu (approved):
💯

Issue #31, opened 2022-12-08
I should be able to click on the crossed out logo.


MicahZoltu:
![image](https://user-images.githubusercontent.com/886059/206357982-37029d97-5b49-4265-ad95-509ae5d0af64.png)

Interceptor has a red X through it, but I don't know why.  I should be able to click on it or hover over it to either figure out what the problem is or potentially resolve the problem.

PR #75, opened 2022-12-21
Explain interceptor icon on mouse over

KillariDev:
fixes: https://github.com/DarkFlorist/TheInterceptor/issues/31

image
image

### DIFF
--- a/extension/app/ts/components/subcomponents/CopyToClipboard.tsx
+++ b/extension/app/ts/components/subcomponents/CopyToClipboard.tsx
@@ -1,7 +1,8 @@
+import { ComponentChildren } from 'preact'
 import { clipboardCopy } from './clipboardcopy.js'
 
 interface CopyToClipboardParams {
-	children: preact.VNode
+	children: ComponentChildren
 	content: string
 	contentDisplayOverride?: string
 	copyMessage?: string
@@ -23,7 +24,7 @@ export function CopyToClipboard(props: CopyToClipboardParams) {
 }
 
 interface ToolTipParams {
-	children: preact.VNode
+	children: ComponentChildren
 	content: string
 }
 
--- a/extension/app/ts/components/subcomponents/CopyToClipboard.tsx
+++ b/extension/app/ts/components/subcomponents/CopyToClipboard.tsx
@@ -27,7 +27,7 @@ interface ToolTipParams {
 	content: string
 }
 
-export function ToolTip(props: ToolTipParams): preact.VNode {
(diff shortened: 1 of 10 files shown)


extension/app/ts/components/subcomponents/CopyToClipboard.tsx
```diff
@@ -21,3 +21,21 @@ export function CopyToClipboard(props: Props): preact.VNode {
 		</div>
 	)
 }
+
+interface ToolTipParams {
+	children: preact.VNode
+	content: string
+}
+
+export function ToolTip(props: ToolTipParams): preact.VNode {
```
MicahZoltu, line 30: The inferred type will likely be better than an explicit type here.
```suggestion
export function ToolTip(props: ToolTipParams) {
```

extension/app/ts/components/subcomponents/CopyToClipboard.tsx
```diff
@@ -1,13 +1,13 @@
...
+interface CopyToClipboardParams {
 	children: preact.VNode
 	content: string
 	contentDisplayOverride?: string
-	copyMessage: string | undefined
+	copyMessage?: string
 }
 
-export function CopyToClipboard(props: Props): preact.VNode {
+export function CopyToClipboard(props: CopyToClipboardParams): preact.VNode {
```
MicahZoltu, line 10: The inferred type will likely be better than an explicit type here.
```suggestion
export function CopyToClipboard(props: CopyToClipboardParams) {
```

Generally speaking, it is better to use inferred types over explicit types unless inference isn't doing the right thing for some reason.

extension/app/ts/components/subcomponents/CopyToClipboard.tsx
```diff
@@ -21,3 +21,21 @@ export function CopyToClipboard(props: Props): preact.VNode {
 		</div>
 	)
 }
+
+interface ToolTipParams {
+	children: preact.VNode
```
MicahZoltu, line 26: Should this be an array?  Also, I think you want `ComponentChildren` here?
https://preactjs.com/guide/v10/typescript#function-components

MicahZoltu (approved):

extension/app/ts/components/subcomponents/CopyToClipboard.tsx
KillariDev: yeah, seems like that works too. I replaced it in other places too

PR #78, opened 2022-12-22
When not used with a wallet, emulate the experimental Metamask APIs

DecentralisedTech:
When there is no wallet, include mock Metamask experimental methods. Tested with and without Metamask on both Brave and Firefox and works as expected - the Browser/Metamask connection actually connects.

### DIFF
--- a/extension/app/inpage/ts/inpage.ts
+++ b/extension/app/inpage/ts/inpage.ts
@@ -120,6 +120,9 @@ type InjectFunctions = {
 	removeListener: (kind: OnMessage, callback: AnyCallBack) => Promise<void>,
 	isConnected: () => boolean,
 	enable: () => void,
+	_metamask?: {
+		isUnlocked: () => boolean
+	}
 }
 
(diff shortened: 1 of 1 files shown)


KillariDev:
We probably should return output of `isConnected()` as its closer to the metamasks `isUnlocked`. You should commit the changed javascript files outputed by `npm run inpage` as well

KillariDev:
You also need to add the added methods to the typescript type definition (`WindowEthereum`), otherwise it should not even compile

MicahZoltu:
I'm a bit 👎 on this, but could be convinced otherwise.  I would like to avoid falling into the trap of wallet spoofing like we saw happen with browsers in the past, where everyone would just spoof being the dominant wallet.  If we do this, it discourages dapp developers from actually fixing their apps.

DecentralisedTech:
@MicahZoltu That makes sense, as long as the provider is EIP1193 compliant then its up to the dApp.
I'll close this for now

PR #79, opened 2022-12-23
Do not show eth donator losing funds anymore. Fix the look of make you rich tx (it seemed to have broken by previous changes)

KillariDev:
image

### DIFF
--- a/extension/app/ts/components/simulationExplaining/SimulationSummary.tsx
+++ b/extension/app/ts/components/simulationExplaining/SimulationSummary.tsx
@@ -6,6 +6,7 @@ import { ERC721Token, Ether, Token, TokenAmount, TokenPrice, TokenSymbol } from 
 import { Transactions } from './Transactions.js'
 import { CopyToClipboard } from '../subcomponents/CopyToClipboard.js'
 import { SomeTimeAgo } from '../subcomponents/SomeTimeAgo.js'
+import { CHAINS } from '../../utils/constants.js'
 
 type EtherChangeParams = {
 	textColor: string,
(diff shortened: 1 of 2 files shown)


extension/app/ts/components/simulationExplaining/SimulationSummary.tsx
```diff
@@ -420,9 +421,9 @@ export function SimulationSummary(param: SimulationSummaryParams) {
 	const ownAddresses = Array.from(summary.entries()).filter( ([address, _balanceSummary]) =>
 		param.simulationAndVisualisationResults.addressMetadata.get(address)?.metadataSource === 'addressBook' || BigInt(address) === param.simulationAndVisualisationResults.activeAddress
 	)
-	const notOwnAddresses = Array.from(summary.entries()).filter( ([address, _balanceSummary]) =>
-		param.simulationAndVisualisationResults.addressMetadata.get(address)?.metadataSource !== 'addressBook' && BigInt(address) !== param.simulationAndVisualisationResults.activeAddress
-	)
+	const notOwnAddresses = Array.from(summary.entries())
+		.filter( ([address, _balanceSummary]) => param.simulationAndVisualisationResults.addressMetadata.get(address)?.metadataSource !== 'addressBook' && BigInt(address) !== param.simulationAndVisualisationResults.activeAddress)
+		.filter( ([address, _balanceSummary]) => CHAINS[param.simulationAndVisualisationResults.chain].eth_donator !== BigInt(address) ) // remove eth donator
```
MicahZoltu, line 426: Will this cause any problems if someone wants to actually simulate as the donor address for some reason?

MicahZoltu (approved):

extension/app/ts/components/simulationExplaining/SimulationSummary.tsx
KillariDev: no, it will show correctly if they are simulating AS the donor (as we are just removing them from `notOwnAddress`). However, if you send funds to the donor, or otherwise interact with them, we are not showing changes to donor address.

MicahZoltu: Is it hard to fix that (so sending to donor works correctly)?  If not, maybe just an edge case we shouldn't worry about.

KillariDev: would add quite a bit complexity. I guess you mean that we would show tokens and eth send's but just hide the "-200k eth"?

MicahZoltu: Yeah, ideally we would *only* hide the make-me-rich transaction and nothing else.

PR #83, opened 2022-12-28
Add edit address button to the popup window

KillariDev:
Add edit address button to the main popup window (you cannot yet edit addresses on other popups, they require some more work)
The edit button is hidden and shows up on mouse over

image

image

### DIFF
--- a/extension/app/ts/components/subcomponents/address.tsx
+++ b/extension/app/ts/components/subcomponents/address.tsx
@@ -21,11 +21,11 @@ export function findAddressInfo(addressToFind: bigint, addressInfos: readonly Ad
 }
 
 export type BigAddressParams = {
-	address: bigint
-	title?: string
-	subtitle?: string
-	noCopying?: boolean
-	renameAddressCallBack: RenameAddressCallBack | undefined
+	readonly address: bigint
+	readonly title?: string
+	readonly subtitle?: string
+	readonly noCopying?: boolean
+	readonly renameAddressCallBack: RenameAddressCallBack | undefined
 }
 
 export function BigAddress(params: BigAddressParams) {
@@ -82,12 +82,12 @@ export function BigAddress(params: BigAddressParams) {
 }
 
 export type ActiveAddressParams = {
-	address: bigint
-	title?: string
-	subtitle?: string
-	simulationMode: boolean
-	changeActiveAddress: () => void
-	renameAddressCallBack: RenameAddressCallBack
+	readonly address: bigint
+	readonly title?: string
+	readonly subtitle?: string
+	readonly simulationMode: boolean
+	readonly changeActiveAddress: () => void
+	readonly renameAddressCallBack: RenameAddressCallBack
 }
 
 export function ActiveAddress(params: ActiveAddressParams) {
@@ -130,10 +130,10 @@ export function ActiveAddress(params: ActiveAddressParams) {
 }
 
 export type SmallAddressParams = {
-	address: bigint
-	addressMetaData: AddressMetadata | undefined
-	textColor?: string
-	renameAddressCallBack: RenameAddressCallBack | undefined
+	readonly address: bigint
+	readonly addressMetaData: AddressMetadata | undefined
+	readonly textColor?: string
+	readonly renameAddressCallBack: RenameAddressCallBack | undefined
 }
 
 export function getAddressName(address: bigint, metadata: AddressMetadata | undefined) {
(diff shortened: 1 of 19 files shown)


KillariDev:
I need to add the modify and switch button to be aware of what is the current address, so if its already the current address, there's no need to prompt user for that

extension/app/ts/components/subcomponents/address.tsx
```diff
@@ -20,48 +20,19 @@ export function findAddressInfo(addressToFind: bigint, addressInfos: readonly Ad
...
-		</div>
-	</div>
-}
-
 export type BigAddressParams = {
 	address: bigint
 	title?: string
 	subtitle?: string
 	noCopying?: boolean
+	renameAddressCallBack: RenameAddressCallBack | undefined
```
MicahZoltu, line 28:
```suggestion
	renameAddressCallBack?: RenameAddressCallBack
```
Same elsewhere.

MicahZoltu, line 28:
```suggestion
	readonly address: bigint
	readonly title?: string
	readonly subtitle?: string
	readonly noCopying?: boolean
	readonly renameAddressCallBack: RenameAddressCallBack | undefined
```
Generally speaking, object properties should be marked as readonly.  This makes auditing the code *way* easier, and often also encourages good design patterns as well (like no shared objects between different parts of an application).

MicahZoltu: (with both pieces of feedback included)
```suggestion
	readonly address: bigint
	readonly title?: string
	readonly subtitle?: string
	readonly noCopying?: boolean
	readonly renameAddressCallBack?: RenameAddressCallBack
```

MicahZoltu: Same comment for all of the other places with this pattern (both in this PR and throughout the codebase).

extension/app/ts/components/subcomponents/address.tsx
```diff
@@ -111,6 +87,7 @@ export type ActiveAddressParams = {
 	subtitle?: string
 	simulationMode: boolean
 	changeActiveAddress: () => void
+	renameAddressCallBack: RenameAddressCallBack
```
MicahZoltu, line 90: This one isn't optional like many others.  What is the reason for that?  Also, I don't understand why this needs to be optional at all, what is the situation where you wouldn't have this set?

extension/app/ts/components/subcomponents/address.tsx
```diff
@@ -128,11 +105,15 @@ export function ActiveAddress(params: ActiveAddressParams) {
...
-				</p>
-			</CopyToClipboard>
+			<div style = 'display: flex; position: relative;'>
+				<RenameAddressButton renameAddress = { () => params.renameAddressCallBack(title, addressString(params.address)) }>
+					<CopyToClipboard content = { ethers.utils.getAddress(addressString(params.address)) } copyMessage = 'Address copied!'>
+						<p class = 'title is-5 noselect nopointer' style = 'text-overflow: ellipsis; white-space: nowrap;'>
+							{ title }
+						</p>
+					</CopyToClipboard>
+				</RenameAddressButton>
```
MicahZoltu, line 115: Why does the rename address button wrap the clipboard?  That feels very bizarre to me.  If you click the rename address button you wouldn't want it to copy the address I don't think.

MicahZoltu (approved):

extension/app/ts/components/subcomponents/address.tsx
KillariDev: It's not set in other than the extension popup, as I don't have system to show the edit address dialog elsewhere at the moment. I could have added it to this PR, but then the PR would be way bigger. Once its implemented everywhere, that doesn't need to be optional anymore

extension/app/ts/components/subcomponents/address.tsx
KillariDev: it doesn't copy the address in that case as `RenameAddressButton` drawn over the text hijacks the click event and doesn't forward it to the `CopyToClipboard `

MicahZoltu: Why is `CopyToClipboard` a child of `RenameAddressButton` then?  It feels like the hierarchy should be something like:
```html
<Address>
	<CopyToClipboard>
		<AddressString/>
	</CopyToClipboard>
	<RenameAddressButton/>
</Address>
```

KillariDev: `CopyToClipboard` a child of `RenameAddressButton` because its trigger is underneath, we first check mouseovers on the address string, then make `RenameAddressButton` visible and at the same time show the tooltip of `CopyToClipboard`. On Click of the element, we'll first check if `RenameAddressButton` is pressed, and if so, open the dialog, if not, we copy the address with `CopyToClipboard`.

I believe we could maybe use the hierarchy you suggested, but I don't really know how. Maybe using `ref` to get the hover effect trigger to `RenameAddressButton`?

extension/app/ts/components/subcomponents/address.tsx
KillariDev: I think undefined is better here as it makes it more explicit that it is missing and it should be fixed

PR #86, opened 2023-01-02
Revert metadata changes

KillariDev:


MicahZoltu (approved):
I'm assuming this broke things which is why we are reverting?

KillariDev:
yeah, as discussed on discord

PR #87, opened 2023-01-02
interceptor_getSimulatedUnsignedTransactions_v1

KillariDev:
Adds method `interceptor_getSimulatedUnsignedTransactions_v1` that can be used to retrieve simulation stack. This can be called as follows:

```ts
console.log(await window.ethereum.request({ 'method': 'interceptor_getSimulatedUnsignedTransactions_v1' }))
```

this returns list of unsigned transactions
```ts
[
	{
		accessList:  []
		chainId:  "0x1"
		from: "0xda9dfa130df4de4673b89022ee50ff26f6ea73cf"
		gas: "0x5208"
		input: "0x"
		maxFeePerGas: "0x0"
		maxPriorityFeePerGas: "0x0"
		nonce: "0x18"
		to: "0xd8da6bf26964af9d7eed9e03e53415d37aa96045"
		type: "0x2"
		value: "0x2a5a058fc295ed000000"
	}
]
```

If no simulation mode is disabled or the stack is empty, it just returns empty array.

### DIFF
--- a/extension/app/ts/background/background.ts
+++ b/extension/app/ts/background/background.ts
@@ -3,7 +3,7 @@ import 'webextension-polyfill'
 import { Simulator } from '../simulation/simulator.js'
 import { EIP2612Message, EthereumAddress, EthereumQuantity, EthereumUnsignedTransaction } from '../utils/wire-types.js'
 import { getSettings, saveActiveChain, saveActiveSigningAddress, saveActiveSimulationAddress, Settings } from './settings.js'
-import { blockNumber, call, chainId, estimateGas, gasPrice, getAccounts, getBalance, getBlockByNumber, getCode, getPermissions, getTransactionByHash, getTransactionCount, getTransactionReceipt, personalSign, requestPermissions, sendRawTransaction, sendTransaction, signTypedDataV4, subscribe, switchEthereumChain, unsubscribe } from './simulationModeHanders.js'
+import { blockNumber, call, chainId, estimateGas, gasPrice, getAccounts, getBalance, getBlockByNumber, getCode, getPermissions, getSimulatedUnsignedTransactions, getTransactionByHash, getTransactionCount, getTransactionReceipt, personalSign, requestPermissions, sendRawTransaction, sendTransaction, signTypedDataV4, subscribe, switchEthereumChain, unsubscribe } from './simulationModeHanders.js'
 import { changeActiveAddress, changeAddressInfos, changeMakeMeRich, changePage, resetSimulation, confirmDialog, RefreshSimulation, removeTransaction, requestAccountsFromSigner, refreshPopupConfirmTransactionSimulation, confirmPersonalSign, confirmRequestAccess, changeInterceptorAccess, changeChainDialog, popupChangeActiveChain, enableSimulationMode, reviewNotification, rejectNotification, addOrModifyAddressInfo } from './popupMessageHandlers.js'
 import { AddressMetadata, SimResults, SimulationState, TokenPriceEstimate } from '../utils/visualizer-types.js'
 import { WebsiteApproval, SignerState, TabConnection } from '../utils/user-interface-types.js'
@@ -291,6 +291,7 @@ export const handlers = new Map<string, SimulationHandler >([
 	['eth_sendRawTransaction', sendRawTransaction],
 	['eth_gasPrice', gasPrice],
 	['eth_getTransactionCount', getTransactionCount],
+	['interceptor_getSimulatedUnsignedTransactions_v1', getSimulatedUnsignedTransactions],
 
 	/*
 	Missing methods:
--- a/extension/app/ts/background/simulationModeHanders.ts
+++ b/extension/app/ts/background/simulationModeHanders.ts
@@ -1,7 +1,7 @@
 import { Simulator } from '../simulation/simulator.js'
 import { bytes32String } from '../utils/bigint.js'
 import { InterceptedRequest } from '../utils/interceptor-messages.js'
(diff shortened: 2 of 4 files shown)


MicahZoltu (approved):

KillariDev:
Change format
```ts
{
    version: '1.0.0',
    transactions: [
        { ... },
        { ... },
    ],
}
```

KillariDev:
add multicall results

KillariDev:
now the call is `interceptor_getSimulationResults` and returns as follows:
```ts
{
	"version": "1.0.0",
	"result": [
		{
			"type": "0x2",
			"from": "0xda9dfa130df4de4673b89022ee50ff26f6ea73cf",
			"nonce": "0x18",
			"maxFeePerGas": "0x0",
			"maxPriorityFeePerGas": "0x0",
			"gas": "0x5208",
			"to": "0xd8da6bf26964af9d7eed9e03e53415d37aa96045",
			"value": "0x2a5a058fc295ed000000",
			"input": "0x",
			"chainId": "0x1",
			"accessList": [],
			"statusCode": 1,
			"gasSpent": "0x5208",
			"returnValue": "0x",
			"events": [],
			"balanceChanges": [
				{
					"address": "0xd8da6bf26964af9d7eed9e03e53415d37aa96045",
					"before": "0x1d7ffb6f77c4f07b3",
					"after": "0x2a5bdd8f798d694f07b3"
				},
				{
					"address": "0xda9dfa130df4de4673b89022ee50ff26f6ea73cf",
					"before": "0x1791b875991eed3f0f19a",
					"after": "0x14ec181c9cf58e6f0f19a"
				}
			],
			"realizedGasPrice": "0x0"
		}
	]
}
```

extension/app/ts/background/background.ts
```diff
@@ -291,7 +291,7 @@ export const handlers = new Map<string, SimulationHandler >([
 	['eth_sendRawTransaction', sendRawTransaction],
 	['eth_gasPrice', gasPrice],
 	['eth_getTransactionCount', getTransactionCount],
-	['interceptor_getSimulatedUnsignedTransactions_v1', getSimulatedUnsignedTransactions],
+	['interceptor_getSimulationResults', getSimulationResults],
```
MicahZoltu, line 294: Consider: `eth_getSimulationStack` or `eth_getTransactionStack`?  This query gives both the simulation stack **and** the results of that simulation and callers may not be specifically looking for results which could make this name confusing to them.

extension/app/ts/background/simulationModeHanders.ts
```diff
@@ -269,11 +269,14 @@ export async function getTransactionCount(simulator: Simulator, port: browser.ru
...
-export async function getSimulatedUnsignedTransactions(simulator: Simulator, port: browser.runtime.Port, request: InterceptedRequest) {
+export async function getSimulationResults(simulator: Simulator, port: browser.runtime.Port, request: InterceptedRequest) {
 	return postMessageIfStillConnected(port, {
 		interceptorApproved: true,
 		requestId: request.requestId,
 		options: request.options,
-		result: EthereumUnsignedTransactionArray.serialize(simulator.simulationModeNode.getSimulatedUnsignedTransactions())
+		result: {
+			version: '1.0.0',
+    		result: GetSimulationResults.serialize(simulator.simulationModeNode.getSimulationResults())
```
MicahZoltu, line 279:
```suggestion
    			result: GetSimulationResults.serialize(simulator.simulationModeNode.getSimulationResults())
```

MicahZoltu: Consider: `payload` rather than `result`, as this includes not only results of simulation but also transactions in the simulation.

extension/app/ts/simulation/services/SimulationModeEthereumClientService.ts
```diff
@@ -52,6 +52,13 @@ export class SimulationModeEthereumClientService {
 		return newSimulation
 	}
 
+	public getSimulationResults = () => {
+		if (this.simulationState === undefined) return []
+		return this.simulationState.simulatedTransactions.map( (x) => {
```
MicahZoltu, line 57: Style consistency:
```suggestion
		return this.simulationState.simulatedTransactions.map((x) => {
```

extension/app/ts/simulation/services/SimulationModeEthereumClientService.ts
```diff
@@ -52,6 +52,13 @@ export class SimulationModeEthereumClientService {
...
+			return { ...x.unsignedTransaction, ...x.multicallResponse, realizedGasPrice: x.realizedGasPrice }
+		})
```
MicahZoltu, line 59: Style: For lambdas that just return an object and do nothing else, consider dropping the return statement to simplify it a bit.  Note: When the expression is just an object, you must surround it with `()` to ensure the `{}` aren't interpreted as a code block.
```suggestion
		return this.simulationState.simulatedTransactions.map((x) => ({ ...x.unsignedTransaction, ...x.multicallResponse, realizedGasPrice: x.realizedGasPrice }))
```

MicahZoltu (approved):
We may want to consider having the returned object contain both a `gas` and a `gasLimit` field.  While `gas` is the "standard", there has been a push in libraries and clients (like ethers) to switch to `gasLimit` which is less confusing/ambiguous for developers.  We can support this endeavor by including both (the caller will simply read the one they want).

extension/app/ts/background/background.ts
KillariDev: I don't really want to call it `eth_` because it's not an eth standard. How about `interceptor_getSimulationStack`?

MicahZoltu: Sounds good to me.  It was my mistake to switch to `eth_`, my intent was to comment on the `SimulationResults` bit and my brain just switched to `eth_` automatically.  😖

PR #88, opened 2023-01-04
Simple readonly top 100 addresses prototype address book and new way to move data between backround page and popups

KillariDev:
- Adds address book that opens into **a new tab**
- The address book is view only at the moment, and only shows top 100 addresses for each category
- Address book communicates with backgroundpage by sending messages, instead of fetching `window.interceptor`, the plan is to change all popups to communicate like this, so we can get rid of the global `window.interceptor` eventually

image


extension/app/html/addressBook.html
```diff
@@ -0,0 +1,36 @@
...
+			"imports": {
+				"ethers": "../vendor/ethers/ethers.esm.js",
+				"webextension-polyfill": "../vendor/webextension-polyfill/browser-polyfill.js",
+				"preact": "../vendor/preact/preact.module.js",
+				"preact/jsx-runtime": "../vendor/preact/jsx-runtime/jsxRuntime.module.js",
+				"preact/hooks": "../vendor/preact/hooks/hooks.module.js",
+				"funtypes": "../vendor/funtypes/index.mjs",
+				"node-fetch": "../vendor/node-fetch/index.mjs",
+				"@zoltu/ethereum-abi-encoder": "../vendor/@zoltu/ethereum-abi-encoder/index.js",
+				"@zoltu/ethereum-crypto": "../vendor/@zoltu/ethereum-crypto/index.js",
```
MicahZoltu, line 24: We should consider moving away from `@zoltu/ethereum-crypto`.  There may also be better alternatives to rlp-encoder and abi-encoder as well that could be worth investigating.

extension/app/html/addressBook.html
```diff
@@ -0,0 +1,36 @@
...
+				"@zoltu/rlp-encoder": "../vendor/@zoltu/rlp-encoder/index.js",
+				"@darkflorist/address-metadata": "../vendor/@darkflorist/address-metadata/index.js"
+			}
+		}
+		</script>
+
+		<main>Loading...</main>
+
+		<!-- a workaround for bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1803984 -->
+		<script type = 'module'>import '../js/addressBookRender.js'</script>
```
MicahZoltu, line 34:
```suggestion
		<script async type = 'module'>import '../js/addressBookRender.js'</script>
```

extension/app/ts/AddressBook.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+export function FilterLink(param: { name: ActiveFilter, currentFilter: ActiveFilter, setActiveFilter: (activeFilter: ActiveFilter) => void}) {
+	return <a
+		class = { param.currentFilter === param.name ? `is-active` : '' }
+		onClick = { () => param.setActiveFilter(param.name) }>
+			{ param.name }
+	</a>
+}
+
+export function AddressList(param: { addressBookEntries: AddressBookEntries | undefined }) {
```
MicahZoltu, line 17: Consider destructuring here so you don't have to do `param.` everywhere.
```suggestion
export function AddressList({addressBookEntries}: { addressBookEntries: AddressBookEntries | undefined }) {
```

extension/app/ts/AddressBook.tsx
```diff
@@ -0,0 +1,121 @@
...
+		) ) }
+	</ul>
+}
+
+export function AddressBook() {
+	const [activeFilter, setActiveFilter] = useState<ActiveFilter>('My Active Addresses')
+	const [addressBookEntries, setAddressBookEntries] = useState<AddressBookEntries | undefined>(undefined)
+	const activeFilterRef = useRef<ActiveFilter>(activeFilter)
+
+	useEffect( () => {
```
MicahZoltu, line 71: Style consistency I think?  You have been doing this a lot lately, maybe it is intentional and the goal is to change the style to include a space here?  I'm not a big fan because it goes against the general style of "no space after parenthesis", but I'm open to having my mind changed.
```suggestion
	useEffect(() => {
```

MicahZoltu (approved):

extension/app/html/addressBook.html
KillariDev: created an issue to look into that

extension/app/ts/AddressBook.tsx
KillariDev: is not intentiol :D

PR #94, opened 2023-01-09
Add search and paging to addressbook

KillariDev:
image

image

### DIFF
--- a/extension/app/ts/AddressBook.tsx
+++ b/extension/app/ts/AddressBook.tsx
@@ -3,14 +3,15 @@ import { addressString } from './utils/bigint.js'
 import { AddressBookEntries } from './utils/user-interface-types.js'
 import Blockie from './components/subcomponents/PreactBlocky.js'
 import { GetAddressBookDataReply, MessageToPopup } from './utils/interceptor-messages.js'
+import { arrayToChunks } from './utils/typed-arrays.js'
 
 type ActiveFilter = 'My Active Addresses' | 'My Contacts' | 'Tokens' | 'Non Fungible Tokens' | 'Other Contracts'
 const PAGE_SIZE = 20
 const ELEMENT_SIZE_PX = 68 + 10
 const PAGE_SIZE_PX = (ELEMENT_SIZE_PX * PAGE_SIZE)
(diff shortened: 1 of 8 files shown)


extension/app/ts/AddressBook.tsx
```diff
@@ -5,6 +5,7 @@ import Blockie from './components/subcomponents/PreactBlocky.js'
 import { GetAddressBookDataReply, MessageToPopup } from './utils/interceptor-messages.js'
 
 type ActiveFilter = 'My Active Addresses' | 'My Contacts' | 'Tokens' | 'Non Fungible Tokens' | 'Other Contracts'
+const PAGE_SIZE = 100
```
MicahZoltu, line 8: What is the purpose of paging the data?  Since we aren't fetching from a remote source, we aren't saving resources or response time by paging the data.  It feels like a single long scrollable list is preferable to paging here.

extension/app/ts/background/metadataUtils.ts
```diff
@@ -9,38 +9,66 @@ import { GetAddressBookDataFilter } from '../utils/interceptor-messages.js'
...
-			data: addressInfos ? addressInfos.slice(filter.startIndex, filter.maxIndex).map( (x) => ({
-				type: 'addressInfo' as const,
-				...x
-			})) : [],
-			length: addressInfos ? addressInfos.length : 0
+		case 'My Active Addresses': {
+			const searchFiltered = searchString === undefined ? addressInfos : addressInfos?.filter( (x) =>
+				x.address.toString().includes(searchString)
+				|| x.name.toLowerCase().includes(searchString)
+			)
```
MicahZoltu, line 19: Consider fuzzy matching instead.  Fuzzy matching provides a better filter experience for users and is relatively easy to implement.

https://blog.amjith.com/fuzzyfinder-in-10-lines-of-python

MicahZoltu (changes requested):
Filtering looks fine, though I would like to see fuzzy matching instead of exact matching.  I'm fairly strongly against pagination given that we don't have a remote datasource.  There are techniques for dealing with large datasets in a UI such that the DOM doesn't explode and lag, and I think we should utilize those rather than pagination as they provide a far superior UX (e.g., scrolling contact list on mobile devices).

MicahZoltu (approved):

PR #99, opened 2023-01-18
addressbook: create/delete addresses, logoURI issue, addressbookStyling

KillariDev:
Allows creating new active addresses and removing them in address book. Also styles address book elements and fixes logoUri issue that was caused by update to the ne addressmetadata version.

image

image

### DIFF
--- a/extension/app/css/interceptor.css
+++ b/extension/app/css/interceptor.css
@@ -149,7 +149,7 @@ html, body {
 	background-color: var(--highlighted-primary-color)
 }
 
-.interceptorInput {
+.input {
 	background: rgba(0, 0, 0, 0);
 	color: var(--text-color);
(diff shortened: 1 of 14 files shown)


MicahZoltu (approved):
I just glanced over it to make sure that there were no backdoors added.  I wonder if we should start getting @0xjimmy and @jubalm to review your PRs as well?  I think there is value in everyone looking at everyone else's work, it helps keep people synchronized.

KillariDev:
> I just glanced over it to make sure that there were no backdoors added. I wonder if we should start getting @0xjimmy and @jubalm to review your PRs as well? I think there is value in everyone looking at everyone else's work, it helps keep people synchronized.

yeah, we could start doing that

PR #100, opened 2023-01-18
Add version number as mandatory field to interceptor_getSimulationStack

KillariDev:
now `interceptor_getSimulationStack` need to be called with version number param, like follows:

`
console.log(await window.ethereum.request({ 'method': 'interceptor_getSimulationStack', 'params': ['1.0.0'] }))
`

fixes https://github.com/DarkFlorist/TheInterceptor/issues/97

### DIFF
--- a/extension/app/ts/utils/wire-types.ts
+++ b/extension/app/ts/utils/wire-types.ts
@@ -791,7 +791,13 @@ export const GetTransactionCount = t.Object({
 }).asReadonly()
 
 export type GetSimulationStack = t.Static<typeof GetSimulationStack>
-export const GetSimulationStack = t.ReadonlyArray(t.Intersect(
+export const GetSimulationStack = t.Object({
+	method: t.Literal('interceptor_getSimulationStack'),
+	params: t.Tuple(t.String),
+}).asReadonly()
+
+export type GetSimulationStackReply = t.Static<typeof GetSimulationStackReply>
+export const GetSimulationStackReply = t.ReadonlyArray(t.Intersect(
 	EthereumUnsignedTransaction,
 	SingleMulticallResponse,
 	t.Object({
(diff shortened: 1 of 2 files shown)


extension/app/ts/utils/wire-types.ts
```diff
@@ -791,7 +791,13 @@ export const GetTransactionCount = t.Object({
 }).asReadonly()
 
 export type GetSimulationStack = t.Static<typeof GetSimulationStack>
-export const GetSimulationStack = t.ReadonlyArray(t.Intersect(
+export const GetSimulationStack = t.Object({
+	method: t.Literal('interceptor_getSimulationStack'),
+	params: t.Tuple(t.String),
```
MicahZoltu, line 796: I feel like this should be `t.Tuple(t.Union(t.Literal('1.0.0')))`.  There is a very constrained/finite set of strings that are actually supported, and anything else should be an error and I think it helps add clarity to deal with that error during deserialization.

KillariDev: I am hadling that error here:
```ts
	if (params.params[0] !== '1.0.0') return postMessageIfStillConnected(port, {
	interceptorApproved: true,
	requestId: request.requestId,
	options: request.options,
	error: {
		message: `Unsupported version number: ${ params.params[0] }`,
		code: 400,
	}
})
```

To give API user better user experience when they are supplying the API with wrong version

MicahZoltu: If a user fails input validation, what kind of error do they get back?  If it isn't informative, then I feel like we should fix that rather than trying to manually deal with every type of malformed data they can provide.

IMO, this shouldn't be treated any different than any other kind of malformed input, and we should give useful errors when malformed input is detected.

KillariDev: they get `Uncaught EthereumJsonRpcError: Unknown error` currently

KillariDev: I am not sure how we can add a general error that would nicely explain what went wrong with the call? Maybe tell what we expected in human readable way, which sounds quite challenging 🤔

KillariDev:
depending on https://github.com/DarkFlorist/TheInterceptor/pull/110

PR #110, opened 2023-01-20
Separates parsing and processing of RPC methods

KillariDev:
Allows us to throw funtypes error to RPC API user and then unknown error for our own errors happening in processing step

### DIFF
--- a/extension/app/ts/background/windows/confirmTransaction.ts
+++ b/extension/app/ts/background/windows/confirmTransaction.ts
@@ -1,10 +1,9 @@
 import { bytes32String } from '../../utils/bigint.js'
 import { METAMASK_ERROR_NOT_CONNECTED_TO_CHAIN, METAMASK_ERROR_USER_REJECTED_REQUEST } from '../../utils/constants.js'
 import { Future } from '../../utils/future.js'
-import { InterceptedRequest } from '../../utils/interceptor-messages.js'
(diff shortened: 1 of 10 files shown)


extension/app/ts/background/windows/confirmTransaction.ts
```diff
@@ -30,34 +29,44 @@ const onCloseWindow = () => { // check if user has closed the window on their ow
...
-	if (window.interceptor.settings === undefined) return
-	if (port.sender?.url === undefined) return
+	if (window.interceptor.settings === undefined) {
+		return {
+			error: {
+				code: 1,
+				message: 'Interceptor not ready'
+			}
+		}
+	}
```
MicahZoltu, line 44: I see this error at least twice.  Should it be extracted into a library of "common errors" or something both to ensure we don't repeat ourselves and to ensure we get consistently shaped errors delivered?

MicahZoltu (approved):
If there are any backdoors in this code, they are extremely underhanded and kudos to you for getting them in so subtly!

Would be good for @jubalm and @0xjimmy to review this even if you merge, just to get some more thorough eyes on it.

extension/app/ts/background/windows/confirmTransaction.ts
KillariDev: I added couple of these to constants. We could look to see how to improve these more in to future thought

PR #112, opened 2023-01-20
Addressbook improvements

KillariDev:
- Makes modal dialogs take at max 90% of the space, instead of always taking that much
- Refactor `getFullLogoUri` calculation behind one function instead of repeating it everywhere where address metadata was used
- Disable most of the buttons that do not work yet (like adding new tokens, removing tokens and such)
- Add edit button. Before the plan was that the user would edit the list directly, but I think it more clear when user clicks the edit button to do edits

image
image
image

### DIFF
--- a/extension/app/ts/AddressBook.tsx
+++ b/extension/app/ts/AddressBook.tsx
@@ -1,19 +1,17 @@
 import { useEffect, useRef, useState } from 'preact/hooks'
 import { addressString } from './utils/bigint.js'
 import { AddressBookEntries, AddressBookEntry, RenameAddressCallBack } from './utils/user-interface-types.js'
-import Blockie from './components/subcomponents/PreactBlocky.js'
 import { GetAddressBookDataReply, MessageToPopup } from './utils/interceptor-messages.js'
 import { arrayToChunks } from './utils/typed-arrays.js'
-import { LOGO_URI_PREFIX } from './background/metadataUtils.js'
-import { Notice } from './components/subcomponents/Error.js'
 import { AddNewAddress } from './components/pages/AddNewAddress.js'
 import { BigAddress } from './components/subcomponents/address.js'
 import { ethers } from 'ethers'
+import Hint from './components/subcomponents/Hint.js'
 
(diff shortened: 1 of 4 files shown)


extension/app/ts/AddressBook.tsx
```diff
@@ -382,77 +363,74 @@ export function AddressBook() {
 
 	return (
 		<main>
```
0xjimmy, line 365:
```suggestion
		<main class = 'max-width: 1920px; margin: auto;' >
```
May be a good idea to limit the width of the content area so content so lengths of wide elements don't start looking bad on wider displays.

Could set the styles in the extension pages ``/extension/app/html/addressBook.html`` instead of the component instead I just don't know how to suggest there.

image
image

0xjimmy:
Looks good. I agree having an edit button makes more sense for the UX, it makes it harder to accidentally edit contacts.

0xjimmy (approved):

extension/app/ts/AddressBook.tsx
MicahZoltu: > May be a good idea to limit the width of the content area so content so lengths of wide elements don't start looking bad on wider displays.

I pretty strongly disagree with this.  The user should be the one ultimately in control of how wide of a window they want, as they can shrink their window if they desire.  By hard coding a limit, you are constraining sizing options available to the user and helping the subset of users who have particularly wide monitors and do not know how to resize a window.  I am of the belief that this set of users is incredibly small and should not be optimized for, especially at the cost of flexibility/responsiveness to other users who *want* to have a very wide view of things.

KillariDev: I guess we could limit the amount on how much the account entry grows to the size where there is no benefit for it to grow anymore (everything fits already

MicahZoltu: Yeah I would be OK with "fit to content" and the content boxes are scaled to fit their content etc.  I am just against a fixed pixel or % width is all.

Issue #117, opened 2023-01-23
Add privatekey 0x1 as an default address and allow signing with it


KillariDev:
What would be a good name for such address?

MicahZoltu:
Name suggestion: `DO NOT SEND ANY ETH TO THIS ADDRESS`

KillariDev:
> Name suggestion: `DO NOT SEND ANYTHING TO THIS ADDRESS`

The problem with this name is, that it doesn't explain where you can use it for

PR #119, opened 2023-01-23
Add 0x1 privatekeys address as second default address and allow signing with it

KillariDev:
fixes https://github.com/DarkFlorist/TheInterceptor/issues/117


MicahZoltu (approved):
This alone probably isn't enough to be useful, and I'm not sure we should merge without doing the rest of the work.

"The rest" of the work is adding the ability to sign messages with the key, so things that require SIWE can be used with interceptor without needing to actually give a real account.

KillariDev:
This adds ability to sign with that address for personal sign. I made it as a special address that can do it

PR #122, opened 2023-01-23
Add version number as mandatory field to interceptor_getSimulationStack

KillariDev:
new version of https://github.com/DarkFlorist/TheInterceptor/pull/100 as the internals of RPC method parsing was changed significantly

now interceptor_getSimulationStack need to be called with version number param, like follows:

`console.log(await window.ethereum.request({ 'method': 'interceptor_getSimulationStack', 'params': ['1.0.0'] }))`

fixes https://github.com/DarkFlorist/TheInterceptor/issues/97

### DIFF
--- a/extension/app/ts/utils/wire-types.ts
+++ b/extension/app/ts/utils/wire-types.ts
@@ -791,8 +791,8 @@ export const GetTransactionCount = t.Object({
 	params: t.Tuple(EthereumAddress, EthereumBlockTag)
 }).asReadonly()
 
-export type GetSimulationStack = t.Static<typeof GetSimulationStack>
-export const GetSimulationStack = t.ReadonlyArray(t.Intersect(
+export type GetSimulationStackReply = t.Static<typeof GetSimulationStackReply>
+export const GetSimulationStackReply = t.ReadonlyArray(t.Intersect(
 	EthereumUnsignedTransaction,
 	SingleMulticallResponse,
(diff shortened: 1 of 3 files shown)


extension/app/ts/utils/wire-types.ts
```diff
@@ -801,6 +801,12 @@ export const GetSimulationStack = t.ReadonlyArray(t.Intersect(
 	}).asReadonly(),
 ))
 
+export type GetSimulationStack = t.Static<typeof GetSimulationStack>
+export const GetSimulationStack = t.Object({
+	method: t.Literal('interceptor_getSimulationStack'),
+	params: t.Tuple(t.Union(t.Literal('1.0.0'))),
```
MicahZoltu, line 807: You may have done this intentionally since we will *eventually* add other versions, in which case feel free to ignore this comment.
```suggestion
	params: t.Tuple(t.Literal('1.0.0')),
```

MicahZoltu (approved):

extension/app/ts/utils/wire-types.ts
KillariDev: Haha, it was directly your suggestion from the previous version :). But yeah, we can remove it was its bit premature

MicahZoltu: I can totally believe that.  🤣

PR #125, opened 2023-01-25
Big metadata refactoring

KillariDev:
- gets rid of `AddressMetadata` variable type and replaces it with `AddressBookEntry` which is adressbook compatible
- In the UI code do not use `bigint` / `string`s as addresses anymore and then retrieve the name of the address from `addressMetadata` mapping, but instead move `AddressBookEntry`:s around. This enables that the UI will know where the address comes from and in the future, can rename and refer to it properly
- Instead of making big nested objects, make more flat objects with only the needed variables in them

Backgroundpage still require more refactoring to get rid of the `window.interceptor` global, but I kept that outside this PR

### DIFF
--- a/extension/app/ts/AddressBook.tsx
+++ b/extension/app/ts/AddressBook.tsx
@@ -133,7 +133,7 @@ export function ListElement(entry: ListElementParam) {
 						<button class = 'card-header-icon' style = 'padding: 0px; margin-left: auto;' aria-label = 'delete'>
 							<span class = 'icon' style = 'color: var(--text-color);' onClick = { entry.type != 'empty' && entry.category === 'My Active Addresses' ? () => entry.removeEntry(entry) : () => {} }> X </span>
 						</button>
-						<button class = 'button is-primary is-small' onClick = { entry.type != 'empty' ? () => entry.renameAddressCallBack(entry) : () => {} }>Edit</button>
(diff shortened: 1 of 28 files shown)


extension/app/ts/AddressBook.tsx
```diff
@@ -135,7 +133,7 @@ export function ListElement(entry: ListElementParam) {
 						<button class = 'card-header-icon' style = 'padding: 0px; margin-left: auto;' aria-label = 'delete'>
 							<span class = 'icon' style = 'color: var(--text-color);' onClick = { entry.type != 'empty' && entry.category === 'My Active Addresses' ? () => entry.removeEntry(entry) : () => {} }> X </span>
 						</button>
-						<button class = 'button is-primary is-small' onClick = { () => {} } disabled = { true }>Edit</button>
+						<button class = 'button is-primary is-small' onClick = { entry.type != 'empty' ? () => entry.renameAddressCallBack(entry) : () => {} }>Edit</button>
```
0xjimmy, line 136: No more disabled attribute being set, I'm not sure how often ``entry.type`` is ``'empty'``  but if its not almost never or not only for a few moments before data is loaded then you may want to have it disabled for help with styling.

0xjimmy (approved):

extension/app/ts/AddressBook.tsx
KillariDev: its always empty when its fetching data from background page, which should take short while and making them disabled would just cause blinking

PR #131, opened 2023-01-26
Allow users to copy addresses in address book

KillariDev:
fixes https://github.com/DarkFlorist/TheInterceptor/issues/123


MicahZoltu (approved):
Why does this option even exist?

KillariDev:
image

When changing active address, you want to select the address by clicking and you don't want to copy the address you are clicking

PR #132, opened 2023-01-26
Activate old address book if it's already open, add title to all html pages

KillariDev:
No when you attempt to open address book and you have it open already, Interceptor will activate that tab instead of opening a new one

### DIFF
--- a/extension/app/ts/background/background.ts
+++ b/extension/app/ts/background/background.ts
@@ -4,7 +4,7 @@ import { Simulator } from '../simulation/simulator.js'
 import { EIP2612Message, EthereumQuantity, EthereumUnsignedTransaction, PersonalSignParams, SendTransactionParams, SupportedETHRPCCall, SwitchEthereumChainParams } from '../utils/wire-types.js'
 import { getSettings, saveActiveChain, saveActiveSigningAddress, saveActiveSimulationAddress, Settings } from './settings.js'
 import { blockNumber, call, chainId, estimateGas, gasPrice, getAccounts, getBalance, getBlockByNumber, getCode, getPermissions, getSimulationStack, getTransactionByHash, getTransactionCount, getTransactionReceipt, personalSign, requestPermissions, sendTransaction, signTypedDataV4, subscribe, switchEthereumChain, unsubscribe } from './simulationModeHanders.js'
-import { changeActiveAddress, changeAddressInfos, changeMakeMeRich, changePage, resetSimulation, confirmDialog, RefreshSimulation, removeTransaction, requestAccountsFromSigner, refreshPopupConfirmTransactionSimulation, confirmPersonalSign, confirmRequestAccess, changeInterceptorAccess, changeChainDialog, popupChangeActiveChain, enableSimulationMode, reviewNotification, rejectNotification, addOrModifyAddressInfo, getAddressBookData, removeAddressBookEntry } from './popupMessageHandlers.js'
+import { changeActiveAddress, changeAddressInfos, changeMakeMeRich, changePage, resetSimulation, confirmDialog, RefreshSimulation, removeTransaction, requestAccountsFromSigner, refreshPopupConfirmTransactionSimulation, confirmPersonalSign, confirmRequestAccess, changeInterceptorAccess, changeChainDialog, popupChangeActiveChain, enableSimulationMode, reviewNotification, rejectNotification, addOrModifyAddressInfo, getAddressBookData, removeAddressBookEntry, openAddressBook } from './popupMessageHandlers.js'
 import { SimResults, SimulationState, TokenPriceEstimate } from '../utils/visualizer-types.js'
 import { WebsiteApproval, SignerState, TabConnection, AddressBookEntry, AddressInfoEntry } from '../utils/user-interface-types.js'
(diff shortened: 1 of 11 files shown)


extension/app/ts/background/popupMessageHandlers.ts
```diff
@@ -211,3 +211,20 @@ export async function getAddressBookData(_simulator: Simulator, payload: PopupMe
 		}
 	})
 }
+
+export async function openAddressBook(_simulator: Simulator, _payload: PopupMessage) {
+	const html = '/html/addressBook.html'
+
+	const openInNewTab = async () => {
+		const tab = await browser.tabs.create({ url: `..${ html }` })
```
0xjimmy, line 219:
```suggestion
	const openInNewTab = async () => {
		const tab = await browser.tabs.create({ url: '../html/addressBook.htm' })
```
Whats the reason for setting the URI from ``'...' + html``?

extension/app/ts/background/background.ts
```diff
@@ -539,7 +539,7 @@ async function onContentScriptConnected(port: browser.runtime.Port) {
 	})
 }
 
-type PopupMessageHandler = (simulator: Simulator, request: PopupMessage) => Promise<void>
+type PopupMessageHandler = (simulator: Simulator, request: PopupMessage) => Promise<unknown>
```
MicahZoltu, line 542: I'm not a huge fan of having some functions return a value while others don't.  I also don't see the return value being used anywhere in this PR, but maybe I'm just missing it?  If it isn't used, then this should stay as `Promise<void>` I think.

KillariDev: Its not used anywhere. That's why I don't see the point in requiring it to be void. The return value doesn't matter

MicahZoltu: You can assign a function that returns a value to a function that returns void, so I think `Promise<void>` is correct here?

MicahZoltu: Oh, this is a variance issue with generics.  I remember this now... 😖

KillariDev: That's other way to do it, but it requires more boilerplate

MicahZoltu: No, your way is right I think.  One could create a new return type like `type Awaitable = Promise<unknown>` and then use that as the acceptable return type, but I'm not sure that is any better.

extension/app/ts/background/popupMessageHandlers.ts
KillariDev: yeah, that's interesting. I doesn't do anything, I removed it

extension/app/ts/background/popupMessageHandlers.ts
```diff
@@ -211,3 +211,20 @@ export async function getAddressBookData(_simulator: Simulator, payload: PopupMe
 		}
 	})
 }
+
+export async function openAddressBook(_simulator: Simulator, _payload: PopupMessage) {
+	const html = '/html/addressBook.html'
+
+	const openInNewTab = async () => {
+		const tab = await browser.tabs.create({ url: html })
```
MicahZoltu, line 219: I think this is what @0xjimmy was getting at previously.  The local variable doesn't provide any value here since this is only used once.  Might as well just inline it to avoid the reader needing to do a lookup.
```suggestion
	const openInNewTab = async () => {
		const tab = await browser.tabs.create({ url: '/html/addressBook.html' })
```

extension/app/ts/background/settings.ts
```diff
@@ -107,3 +108,12 @@ export function saveSimulationMode(simulationMode: boolean) {
 export function savePendingAccessRequests(pendingAccessRequests: readonly PendingAccessRequest[]) {
 	browser.storage.local.set({ pendingAccessRequests: pendingAccessRequests })
 }
+export function saveOpenedAddressBookTabId(addressbookTabId: number) {
+	browser.storage.local.set({ addressbookTabId: addressbookTabId })
+}
+
+export async function getOpenedAddressBookTabId() {
+	const tabIdData = await browser.storage.local.get(['addressbookTabId'])
```
MicahZoltu, line 116: Seems odd to store a tab ID in local storage.  When you restart your browser, do tab IDs get reused?  Could we accidentally overwrite someone's tab by doing this?  Why not just store it in memory?

MicahZoltu (approved):

extension/app/ts/background/settings.ts
KillariDev: if you store it in memory and restart the browser, the memory gets lost, so the tab gets lost too and the addressbook would not be attached to anything.

ID's in chrome and firefox seemed to be big incremental numbers, so they get reused only after long time I believe, I don't know if this is some standard or not.

KillariDev: Initially I tried to just scan through all tabs and see if adressbook is open, but that requires extra access rights for the extension, and I thought I can implement it like this as well

MicahZoltu: > if you store it in memory and restart the browser, the memory gets lost, so the tab gets lost too and the addressbook would not be attached to anything.

Tab IDs are stable across browser restarts?  What happens if you have your browser setup to *not* remember tabs across restarts?

Issue #124, opened 2023-01-24
fix header


KillariDev:
![image](https://user-images.githubusercontent.com/13102010/214224038-97b12b5c-7fbe-4c05-8fcc-b66b3a46fb6b.png)

PR #134, opened 2023-01-26
Removes old address list, let's use the new and shiny address book from this point on

KillariDev:
fixes https://github.com/DarkFlorist/TheInterceptor/issues/124 (one less element)


MicahZoltu (approved):
I love pure deletion PRs!

0xjimmy (approved):

PR #135, opened 2023-01-27
Show signed data for permit2 / EIP712 instead of swallowing them

KillariDev:
We now support all ERC methods eth_signTypedData, V1, V2, V3, V4 and display them beatifully:

image

Known public key signing is only available 'personal_sign' currently.

I plan to refactor this personal sign dialog (first to get rid of the global state and then visual improvements), but this PR gets this already working, so I think its worth it.


extension/app/ts/background/windows/personalSign.ts
```diff
@@ -30,37 +30,63 @@ function getAddressMetadataForEIP2612Message(message: EIP2612Message, addressInf
 	]
 }
 
+function getAddressMetadataForPermitMessage(message: Permit2, addressInfos: readonly AddressInfo[] | undefined) : [string, AddressBookEntry][] {
+	return [
+		[addressString(message.message.details.token), getAddressMetaData(message.message.details.token, addressInfos)],
+		[addressString(message.message.spender), getAddressMetaData(message.message.spender, addressInfos)],
+		[addressString(message.domain.verifyingContract), getAddressMetaData(message.domain.verifyingContract, addressInfos)],
+	]
```
MicahZoltu, line 38: Anytime you return an object or a tuple like this, it usually is a good idea to add `as const` to ensure that you are returning the narrowest type possible, rather than a broad type which is what TS will default to.
```suggestion
	return [
		[addressString(message.message.details.token), getAddressMetaData(message.message.details.token, addressInfos)],
		[addressString(message.message.spender), getAddressMetaData(message.message.spender, addressInfos)],
		[addressString(message.domain.verifyingContract), getAddressMetaData(message.domain.verifyingContract, addressInfos)],
	] as const
```

MicahZoltu: Though, since you have an explicit return type this doesn't actually matter much.  I would remove the explicit return type personally but I think we already discussed that and you disagree.

extension/app/ts/background/windows/personalSign.ts
```diff
@@ -30,37 +30,63 @@ function getAddressMetadataForEIP2612Message(message: EIP2612Message, addressInf
 	]
 }
 
+function getAddressMetadataForPermitMessage(message: Permit2, addressInfos: readonly AddressInfo[] | undefined) : [string, AddressBookEntry][] {
```
MicahZoltu, line 33:
```suggestion
function getAddressMetadataForPermitMessage(message: Permit2, addressInfos: readonly AddressInfo[] | undefined): [string, AddressBookEntry][] {
```

extension/app/ts/background/windows/personalSign.ts
```diff
@@ -30,37 +30,63 @@ function getAddressMetadataForEIP2612Message(message: EIP2612Message, addressInf
...
+			const parsed = EIP2612Message.parse(params.params[1])
+			window.interceptor.personalSignDialog =  {
+				simulationMode: simulationMode,
+				requestId: requestId,
+				message: stringifyJSONWithBigInts(parsed.message),
+				account: addressString(params.params[0]),
+				method: params.method,
+				addressBookEntries: getAddressMetadataForEIP2612Message(parsed, window.interceptor.settings?.addressInfos),
+				eip2612Message: parsed,
+			}
```
MicahZoltu, line 68: Looking forward to the removal of global state usage in a future PR.

MicahZoltu (approved):

extension/app/ts/background/windows/personalSign.ts
KillariDev: well here we don't really want to return a constant sized array, but a dynamic array as its stored into one

Issue #126, opened 2023-01-25
Can't switch away from BSC after switching to it


epheph:
I was playing around with pancakeSwap and using BSC (in bypass mode with interceptor). Then when i came back to Uniswap, i couldn't get it back on mainnet.

https://user-images.githubusercontent.com/361654/214674434-e5dc4d43-7696-4802-9579-403c743822da.mov

epheph:
I also noticed that there was an alert for account approval (couldn't see because scroll required). i dismissed it, and then it connected but it still said it was on BSC.

PR #143, opened 2023-01-31
Remove window.interceptor.changeChainDialog global. Fix chain changing bugs and add handleSigningMode

KillariDev:
- Removes `window.interceptor.changeChainDialog` global
- Fixes https://github.com/DarkFlorist/TheInterceptor/issues/126
- Add `handleSigningMode` that is similar to `handleSimulationMode` to make more explicit which methods are forward and which are not. Handling these things vaguely was also one cause of #126 bug

### DIFF
--- a/extension/app/ts/components/pages/ChangeChain.tsx
+++ b/extension/app/ts/components/pages/ChangeChain.tsx
@@ -1,6 +1,7 @@
 import { useState, useEffect } from 'preact/hooks'
 import { getChainName, isSupportedChain } from '../../utils/constants.js'
-import { Error, ErrorCheckBox } from '../subcomponents/Error.js'
+import { Error as ErrorContainer, ErrorCheckBox } from '../subcomponents/Error.js'
+import { ChangeChainRequest, MessageToPopup } from '../../utils/interceptor-messages.js'
 
 interface InterceptorChainChangeRequest {
 	isInterceptorSupport: boolean,
@@ -8,6 +9,7 @@ interface InterceptorChainChangeRequest {
 	origin: string,
 	icon: string | undefined,
 	simulationMode: boolean,
+	requestId: number,
 }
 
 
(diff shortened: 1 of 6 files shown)


extension/app/ts/components/pages/ChangeChain.tsx
```diff
@@ -16,39 +18,37 @@ export function ChangeChain() {
 	const [connectAnyway, setConnectAnyway] = useState<boolean>(false)
 
 	useEffect( () => {
-		function popupMessageListener(msg: unknown) {
-			console.log('popup message')
-			console.log(msg)
-			fetch()
+		async function popupMessageListener(msg: unknown) {
```
jubalm, line 21: this can be declared outside `useEffect`?

extension/app/ts/components/pages/ChangeChain.tsx
```diff
@@ -16,39 +18,37 @@ export function ChangeChain() {
...
+		browser.runtime.sendMessage( { method: 'popup_changeChainDialog', options: { accept: true, requestId: chainChangeData.requestId } } )
+		window.close()
 	}
 
 	function reject() {
-		browser.runtime.sendMessage( { method: 'popup_changeChainDialog', options: { accept: false } } )
+		if ( chainChangeData?.requestId === undefined) throw new Error('Request id is missing')
+		browser.runtime.sendMessage( { method: 'popup_changeChainDialog', options: { accept: false, requestId: chainChangeData.requestId } } )
+		window.close()
 	}
```
jubalm, line 53: can use guard clause here?

jubalm (approved):

extension/app/ts/components/pages/ChangeChain.tsx
KillariDev: it can be declared outside, but its not used anywhere else. Do you feel its cleaner if its declared outside?

MicahZoltu: I personally have a preference for declaring things as locally as possible.  I find it helps me keep track of where stuff are used and ensures they get deleted when they aren't used anymore (though TS mostly handles this for up to file local declarations).

PR #148, opened 2023-02-01
Add sendPopupMessageToBackgroundPage to benefit from type safety of messages

KillariDev:
wraps `browser.runtime.sendMessage` with `sendPopupMessageToBackgroundPage` to make sure the type is correct for messages

### DIFF
--- a/extension/app/ts/background/backgroundUtils.ts
+++ b/extension/app/ts/background/backgroundUtils.ts
@@ -1,4 +1,4 @@
-import { MessageToPopup } from '../utils/interceptor-messages.js'
+import { MessageToPopup, PopupMessage } from '../utils/interceptor-messages.js'
 
 export function getActiveAddress() {
 	if (window.interceptor.settings === undefined) return undefined
@@ -18,4 +18,8 @@ export async function sendPopupMessageToOpenWindows(message: MessageToPopup) {
 			if (error) return console.error(`Popup message error: ${ error.message }`);
 		}
 	}
+}
+
+export async function sendPopupMessageToBackgroundPage(message: PopupMessage) {
+	await browser.runtime.sendMessage(PopupMessage.serialize(message))
 }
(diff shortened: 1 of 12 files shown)


extension/app/ts/background/backgroundUtils.ts
```diff
@@ -19,3 +19,7 @@ export async function sendPopupMessageToOpenWindows(message: MessageToPopup) {
 		}
 	}
 }
+
+export async function sendPopupMessageToBackgroundPage(message: PopupMessage) {
+	await browser.runtime.sendMessage(PopupMessage.serialize(message))
+}
```
MicahZoltu, line 25: At a glance, it appears this could be made easier to use by changing the parameters:
```suggestion
export async function sendPopupMessageToBackgroundPage(method: PopupMessage['method'], options: PopupMessage['options']) {
	await browser.runtime.sendMessage(PopupMessage.serialize({ method, options }))
}
```
This would make it so the caller didn't have to repetitiously type so much and could just do `sendPopupMesageToBackgroundPage('popup_removeAddressBookEntry', { address, addressBookCategory })`.

MicahZoltu (approved):

extension/app/ts/background/backgroundUtils.ts
KillariDev: not all Popup messages have options field, so I am not sure how we could get this working with that. On option would be to add empty options on all such messages

KillariDev: hmm, then the challenge is that you always need to match the method and options types to each other, I am not sure typescript can do that. Maybe with templates?

MicahZoltu: Ah, I didn't realize `PopupMessage` was a union.  Do this instead then:
```suggestion
export async function sendPopupMessageToBackgroundPage<T extends PopupMessage>(method: T['method'], options: T['options']) {
	await browser.runtime.sendMessage({ method, options })
}
```
I *think* that will handle optional `options`, though I'm not 100% certain.

PR #151, opened 2023-02-02
Remove `window.interceptor.confirmTransactionDialog` and refactoring

KillariDev:
- In popup messages, use 'method' as the identifier instead of mix between method/message
- In simulation' remove `unsignedTransaction` as the same data is already covered in `signedTransaction` field


KillariDev:
the test failure is fixed in the other pr: <https://github.com/DarkFlorist/TheInterceptor/pull/153>

0xjimmy (approved):

PR #153, opened 2023-02-03
Fix transaction conversion

KillariDev:


extension/app/ts/simulation/services/SimulationModeEthereumClientService.ts
```diff
@@ -98,9 +98,14 @@ export class SimulationModeEthereumClientService {
...
+	public static mockSignTransaction = async (transaction: EthereumUnsignedTransaction) : Promise<EthereumSignedTransaction> => {
 		const unsignedTransaction = EthereumUnsignedTransactionToUnsignedTransaction(transaction)
+		if (transaction.type === 'legacy') {
+			const signatureParams = { r: 0n, s: 0n, v: 0n }
+			const hash = await keccak256.hash(serializeTransactionToBytes({ ...unsignedTransaction, ...signatureParams }))
+			return { ...transaction, ...signatureParams, hash }
+		}
+		const signatureParams = { r: 0n, s: 0n, yParity: 'even' as const }
 		const hash = await keccak256.hash(serializeTransactionToBytes({ ...unsignedTransaction, ...signatureParams }))
 		return { ...transaction, ...signatureParams, hash }
```
MicahZoltu, line 110: Style: Consider an else here.  I like using if/else when both code paths are equally interesting/important.  In this case, `if (transaction.type === 'legacy'` isn't a guard clause but rather a primary code path, as is the other case.  This helps make it more clear that we are branch here, not guarding.
```suggestion
		if (transaction.type === 'legacy') {
			const signatureParams = { r: 0n, s: 0n, v: 0n }
			const hash = await keccak256.hash(serializeTransactionToBytes({ ...unsignedTransaction, ...signatureParams }))
			return { ...transaction, ...signatureParams, hash }
		} else {
			const signatureParams = { r: 0n, s: 0n, yParity: 'even' as const }
			const hash = await keccak256.hash(serializeTransactionToBytes({ ...unsignedTransaction, ...signatureParams }))
			return { ...transaction, ...signatureParams, hash }
		}
```

extension/app/ts/utils/ethereum.ts
```diff
@@ -68,12 +78,6 @@ function isSignedTransaction(maybeSigned: unknown): maybeSigned is ISignedTransa
 		&& 'yParity' in maybeSigned
 }
 
-export function getV(transaction: Pick<ISignedTransaction, 'chainId' | 'yParity'>) {
-	return 'chainId' in transaction && transaction.chainId !== undefined
-	? (transaction.yParity === 'even' ? 0n : 1n) + 35n + 2n * transaction.chainId
-	: transaction.yParity === 'even' ? 27n : 28n
-}
-
```
MicahZoltu, line 76: I thought after our discussion we determined that this is necessary?

extension/app/ts/utils/ethereum.ts
```diff
@@ -50,15 +50,25 @@ export interface IUnsignedTransaction1559 {
 	}[]
 }
 
-export interface ITransactionSignature {
+export interface ITransactionSignatureLegacy {
+	readonly r: bigint
+	readonly s: bigint
+	readonly v: bigint
+	readonly hash: bigint
+}
```
MicahZoltu, line 58: Internally, I personally prefer storing all transaction signatures as `r,s,yParity` and only converting to `v` at (de)serialization time.  This helps keep things consistent even though the wire protocol is insane.

extension/app/ts/utils/ethereum.ts
```diff
@@ -214,6 +218,59 @@ export function EthereumUnsignedTransactionToUnsignedTransaction(transaction: Et
...
+			to: transaction.to,
+			value: transaction.value,
+			input: transaction.input,
+			r: transaction.r,
+			s: transaction.s,
+			v: transaction.v,
+			hash: transaction.hash,
+		}
+	}
+}
```
MicahZoltu, line 272:
```suggestion
export function EthereumSignedTransactionToSignedTransaction(transaction: EthereumSignedTransaction): ISignedTransaction {
	if (transaction.accessList !== undefined) return transaction
	if (transaction.type !== 'legacy') return transaction
	return { ...transaction, accessList: [] }
}
```

MicahZoltu (approved):
This PR seems to not align with what I *thought* the conclusion of our discussion in Discord was.  I thought we agreed that things were not in fact broken other than legacy transaction signing wasn't supported?

extension/app/ts/utils/ethereum.ts
KillariDev: its not as ethers calculates it from yParity so that we don't have to

KillariDev:
> This PR seems to not align with what I _thought_ the conclusion of our discussion in Discord was. I thought we agreed that things were not in fact broken other than legacy transaction signing wasn't supported?

are you referring to storing yparity/v separion with this or something else?

MicahZoltu:
The PR appears to transaction signing, but I thought after our discussing that we determined that transaction signing is actually working correctly so doesn't need to be changed?  Perhaps there are other reasons to make these changes?

KillariDev:
I modified `mockSignTransaction` on the previous PR and it broke the tests, and then I figured the test for mocking was broken anyway. I added other test that is positive test for transaction signing

extension/app/ts/utils/ethereum.ts
```diff
@@ -50,15 +50,25 @@ export interface IUnsignedTransaction1559 {
 	}[]
 }
 
-export interface ITransactionSignature {
+export interface ITransactionSignatureLegacy {
+	readonly r: bigint
+	readonly s: bigint
+	readonly v: bigint
+	readonly hash: bigint
```
MicahZoltu, line 57: Is this the hash of the unsigned transaction or the signed transaction?

KillariDev: it should be signed transactions hash

KillariDev: maybe worth specifying, but I don't unsigned hashes are used anywhere except in the signing process?

PR #157, opened 2023-02-07
Make RPC output more aligned with nethermind

KillariDev:
- Adds `EthereumJSONRpcRequestHandler` to allow us to mock JSON rpc requests
- Add testing for `eth_getBlockByNumber` and `eth_getTransactionByHash`
- Align output of these methods to Nethermind


extension/app/ts/simulation/services/EthereumJSONRpcRequestHandler.ts
```diff
@@ -0,0 +1,27 @@
...
+export type IEthereumJSONRpcRequestHandler = Pick<EthereumJSONRpcRequestHandler, keyof EthereumJSONRpcRequestHandler>
+export class EthereumJSONRpcRequestHandler {
+	private nextRequestId: number = 1
+	private endpoint: string
+
+	constructor(endpoint: string) {
+		this.endpoint = endpoint
+    }
+
+	public readonly jsonRpcRequest = async (method: string, params: readonly unknown[]) => {
```
MicahZoltu, line 12: Consider making these into a single object parameter and having the type be a union of all known Ethereum JSON-RPC types like:
```suggestion
	type EthereumJsonRpcRequest = EthereumGetBlockByNumberRequest | EthereumGetBlockByHashRequest | EthereumGetTransactionReceipt | ...
	...
	public readonly jsonRpcRequest = async (request: EthereumJsonRpcRequest) => {
```
This way, you can do serialization inside this function instead of having to do it prior to calling this function everywhere.

If you *really* want to go crazy you can do some complicated type stuff to make it so you can even have this function parse the result based on the expected result type by having a mapping of request to response types somewhere.  This requires quite a bit of effort to setup and at least one sacrifice to the TypeScript Gods, but it can really improve the ergonomics of the library going forward.

extension/app/ts/utils/typescript.ts
```diff
@@ -36,3 +36,8 @@ export function assertsOneOf<T, U extends T>(item: T, array: readonly U[], messa
 export type DistributedPick<T, K extends string> = T extends unknown ? { [P in K & keyof T]: T[P] } : never
 export type DistributedOmit<T, K extends keyof T> = T extends unknown ? Omit<T, K> : never
 export type DistributiveOmit<T extends object, K extends keyof T> = T extends unknown ? Omit<T, K> : never
+
+export function asObject(object: unknown) : Object {
+	if (typeof object !== 'object' || object === null) throw new Error('Not object')
+	return object
+}
```
MicahZoltu, line 43: 6 suggestions in one:
1. `object` is a keyword, so I recommend not using it as a variable name as it can confuse some tools sometimes.
2. No space after parameters before return type `:`.
3. Change this into a type asserting function instead of something that returns itself.
4. Check to see if it is an array as well.
5. Include the object type in the error.
6. Make both a type checking and a type asserting function, as you may need both eventually.
```suggestion
export function isObject(maybe: unknown): maybe is Object {
	return typeof maybe === 'object' && maybe !== null && !Array.isArray(maybe)
}
export function assertIsObject(maybe: unknown): asserts maybe is Object {
	if (!isObject(maybe)) throw new Error('Expected object but got ${typeof maybe}')
}
```
Usage example:
```ts
declare const apple: unknown

apple // type: unknown
assertIsObject(apple)
apple // type: Object

declare const banana: unknown
if (isObject(banana)) {
	banana // type: Object
} else {
	banana // type: unknown
}
```

MicahZoltu (approved):

extension/app/ts/simulation/services/EthereumJSONRpcRequestHandler.ts
KillariDev: yeah, I added input value checking as that was easy to do and has quite significant improvements. The return values is a good idea, but Its bit too much work for now, I created issue about it

PR #158, opened 2023-02-07
Fix forgotten handler for interceptorAccessReadyAndListening

KillariDev:
This causes interceptor to throw when approving access request that pops initially


MicahZoltu (approved):
Can we do anything with type checking to ensure we don't forget items in the future?  Do we have a union of all required keys somewhere (or can we create one)?

0xjimmy (approved):

KillariDev:
> Can we do anything with type checking to ensure we don't forget items in the future? Do we have a union of all required keys somewhere (or can we create one)?

yeah. I was planning to do this refactor at some point, but I guess this was a good time to do it as I already run into a bug that was caused by this :). I refactored popup message handling to be similar with rpc method handling

PR #166, opened 2023-02-09
replace `import * as t from 'funtypes'` with `import * as funtypes from 'funtypes'`

KillariDev:


MicahZoltu (approved):
I'm fine with the change, but this is one of the very few situations where I'm OK with a shorthand.  This entire file uses `funtypes` on almost every line, so while normally I would argue that verbosity is valuable, I think in this case the verbosity is a net loss and jargon is favorable due to the sheer volume of references.

KillariDev:
> I'm fine with the change, but this is one of the very few situations where I'm OK with a shorthand. This entire file uses `funtypes` on almost every line, so while normally I would argue that verbosity is valuable, I think in this case the verbosity is a net loss and jargon is favorable due to the sheer volume of references.

I mostly want to do this change as other files use some amount of funtypes, and I don't like the import to be different here. But I also think calling the lib "funtypes" is also more clear than "t"

PR #168, opened 2023-02-09
Adds Sepolia

KillariDev:
Fixes https://github.com/DarkFlorist/TheInterceptor/issues/82

### DIFF
--- a/extension/app/ts/utils/user-interface-types.ts
+++ b/extension/app/ts/utils/user-interface-types.ts
@@ -8,7 +8,7 @@ import { SignerName } from './interceptor-messages.js'
 import { WebsiteAccessArray } from '../background/settings.js'
 
 export type CHAIN = keyof typeof CHAINS
-export const CHAIN = funtypes.Union(funtypes.Literal('1'), funtypes.Literal('5'))
+export const CHAIN = funtypes.Union(funtypes.Literal('1'), funtypes.Literal('5'), funtypes.Literal('11155111'))
 
 export type AddressInfo = funtypes.Static<typeof AddressInfo>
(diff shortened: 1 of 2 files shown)


MicahZoltu (approved):

extension/app/ts/utils/user-interface-types.ts
```diff
@@ -8,7 +8,7 @@ import { SignerName } from './interceptor-messages.js'
 import { WebsiteAccessArray } from '../background/settings.js'
 
 export type CHAIN = keyof typeof CHAINS
-export const CHAIN = funtypes.Union(funtypes.Literal('1'), funtypes.Literal('5'))
+export const CHAIN = funtypes.Union(funtypes.Literal('1'), funtypes.Literal('5'), funtypes.Literal('11155111'))
```
MicahZoltu, line 11: Consider using a literal parser to make it so these are serialized by chain ID, but the value passed around is a string name of the chain instead.  So you could have a transaction with `{ chainId: 'sepolia' }` which would serialize to `{ chainId: '11155111' }`.

KillariDev: hmm. I don't think that's a good idea. I wouldn't want user to be able to send transaction with `chainId: 'sepolia'` as thats not a standard

MicahZoltu: Sorry, I meant *internally* we would use strings and we would only convert to a number at serialization time.  The idea here is that debugging and troubleshooting is easier if our internal datastructures are easy to understand, rather than using the chain IDs internally.

PR #169, opened 2023-02-10
Upload source code containing version and git hash for reproducible builds

epheph:


.github/workflows/release-extension.yml
```diff
@@ -22,6 +22,8 @@ jobs:
           echo export const gitCommitSha = \"$GIT_COMMIT_SHA\" > $VERSION_FILE
           echo export const version = \"$VERSION\" >> $VERSION_FILE
           sed -i '/"version"/c\\t"version": "'$VERSION'",' $MANIFEST_FILE
+          zip /tmp/source.zip -r .
+          mv /tmp/source.zip .
```
MicahZoltu, line 26: Why not zip directly to final location?

MicahZoltu (approved):

.github/workflows/release-extension.yml
epheph: Changed it to -x exclude since i needed to exclude .git anyway.

Issue #128, opened 2023-01-26
edit address button expads the whole view and causes shaking


KillariDev:
![image](https://user-images.githubusercontent.com/13102010/214780727-fc45adee-7ce7-4157-9ea9-8d58f447cc84.png)

MicahZoltu:
If it is the shaking I'm thinking of, it is probably because you are adding a border on hover and using browser default [box sizing](https://developer.mozilla.org/en-US/docs/Web/CSS/box-sizing).  Easy fix (to make it apply everywhere):
```css
* {
	box-sizing: border-box;
}
```
The reason browsers default to `content-box` is for legacy compatibility.  I think at this point everyone agrees that `border-box` is the right default for any *new* website.

PR #171, opened 2023-02-13
Rename button fix and make tooltips more reliable

KillariDev:
Fixes https://github.com/DarkFlorist/TheInterceptor/issues/128

Also makes tooltips appear more reliably, and next to the curser instead of dependent on the object


MicahZoltu (approved):
It would be cool (but almost certainly not worth the time) if we had some kind of mock site that showcased a bunch of different features without needing to actually run/use the app.  Could serve it as a static site on each PR so the reviewer can easily poke around and look at the changes without downloading/installing the extension.

KillariDev:
> It would be cool (but almost certainly not worth the time) if we had some kind of mock site that showcased a bunch of different features without needing to actually run/use the app. Could serve it as a static site on each PR so the reviewer can easily poke around and look at the changes without downloading/installing the extension.

yeah, we have that now for the absurd use and lunaria, but not for interceptor :/

MicahZoltu:
The other two projects I think are notably easier because they are natively static hosted pages.  In the case of the extension, it would be cool if we can have a statically hosted version of some of the screens with some placeholder data mocked so PRs can easily be linked.  This is likely a lot of work.

KillariDev:
> The other two projects I think are notably easier because they are natively static hosted pages. In the case of the extension, it would be cool if we can have a statically hosted version of some of the screens with some placeholder data mocked so PRs can easily be linked. This is likely a lot of work.

I added issue about this.

PR #174, opened 2023-02-14
Fixes curves issues

KillariDev:
- Fixes `eth_call` to replace missing values correctly
- If we get `'sender has deployed code'` error, retry the call with different caller (fixes issues like curve)
- If we already know common contract callers (like curve), we can just swap the address without making that additional call
- If call fails, return errors correctly and don't just return the error return value

### DIFF
--- a/extension/app/ts/background/simulationModeHanders.ts
+++ b/extension/app/ts/background/simulationModeHanders.ts
@@ -1,6 +1,6 @@
 import { Simulator } from '../simulation/simulator.js'
 import { bytes32String } from '../utils/bigint.js'
-import { ERROR_INTERCEPTOR_UNKNOWN_ORIGIN } from '../utils/constants.js'
+import { ERROR_INTERCEPTOR_UNKNOWN_ORIGIN, KNOWN_CONTRACT_CALLER_ADDRESSES } from '../utils/constants.js'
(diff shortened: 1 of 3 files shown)


extension/app/ts/background/simulationModeHanders.ts
```diff
@@ -69,27 +69,47 @@ export async function sendTransaction(getActiveAddressForDomain: (websiteAccess:
...
 		accessList: [],
 	}
-	const result = request.params.length > 1 ? await simulator.simulationModeNode.call(transaction, request.params[1]) : await simulator.simulationModeNode.call(transaction)
-	return { result: result }
+	return await simulator.simulationModeNode.call(transaction, blockTag)
 }
+
+export async function call(simulator: Simulator, request: EthCallParams) {
+	const callParams = request.params[0]
+	const from = callParams.from !== undefined && !KNOWN_CONTRACT_CALLER_ADDRESSES.includes(callParams.from) ? callParams.from : defaultCallAddress
```
epheph, line 102: so we can add addresses here to optimize against lots of failed calls. It is unfortunate that if a site does this over and over again (as it is likely to do) we will need to discover this over and over via initial failures, but it works for now.

KillariDev: yeah, I was thinking of adding caching here, but I didn't want to make it overly complex. So that if it makes a call and identifies its contract, it would change the address always for those for that client

epheph (approved):
Works great

extension/app/ts/utils/constants.ts
```diff
@@ -204,3 +204,7 @@ export const CHAIN_NAMES = new Map<string, string>( [
 export function getChainName(chainId: bigint) { return CHAIN_NAMES.get(chainId.toString()) || `Chain: ${chainId.toString()}` }
 
 export const MOCK_PRIVATE_KEYS_ADDRESS = 0x7E5F4552091A69125d5DfCb7b8C2659029395Bdfn // an address represeting 0x1 privatekey
+
+export const KNOWN_CONTRACT_CALLER_ADDRESSES = [
+	0xca11bde05977b3631167028862be2a173976ca11n // curve multicaller
+]
```
MicahZoltu, line 210: If we think this is going to grow to a significant size we should use an object or a Set.  While array lookups can be faster for very small datasets, eventually a set or object will be significantly faster to test set inclusion.

MicahZoltu:
Someone should really just fix eth_multicall...

extension/app/ts/utils/constants.ts
KillariDev: sure, but I felt it was premature optimization, like with caching

PR #176, opened 2023-02-15
Standardize manifest json to mozilla-style

epheph:


extension/app/manifest.json
```diff
@@ -1,41 +1,46 @@
...
-		"400": "img/LOGOA_400x400.png"
-	},
-	"permissions": [
-		"activeTab",
-		"<all_urls>",
-		"storage",
-		"webNavigation"
-	],
-	"web_accessible_resources": ["vendor/*", "js/*", "inpage/*"],
-	// all hashes except the first one are from .html files `<script type = 'module'>import '../js/popup.js'</script>` part
```
KillariDev, line 26: I feel this is valuable commit. But if Firefox will approve interceptor after its, I think its worth getting rid of these. Is it sure that this is the solution, or do we try, and then bring these back if that was not the reason?

MicahZoltu (commented):
I'm somewhat against this change for two reasons:
1. Mozilla's documentation *explicitly* says that comments are allowed in `manifest.json` (https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json).
2. Tabs are allowed in JSON files and changing this means we have one file in our repository that doesn't follow the repository wide conventions.  At the *least* we should add a special case to `.editorconfig`.

I would *much* rather pressure Mozilla to fix the bug on their end than just hack around it.  This hack will live with us forever and I'm sure there are other extension developers that will run into this same problem, so pressing Mozilla to fix it would be a generally good thing to do.

What is the status of getting approvals?  Is there somewhere I can add commentary/feedback to argue that they are the ones that are broken and/or follow along with the case?

MicahZoltu:
This also feels like one of those problems that if we hack around it today it is just going to come back and bite us later.  For example, if their approval process is broken in some weird obscure way and we hack around it, then it is quite possible that their broken approval process will change subtly and we'll run into another problem.  I think applying pressure to them to have a sane approval process (fix their bug) is the way to move forward, so we don't have to keep mutating things infinitely to get extensions approved.

epheph:
I'm beginning to think the issue is the comparison of `app` directory, because manifest.json was the only difference, i assumed that was the problem, but i think they were just looking at the entirely wrong directory and we were overfitting here. Let's not merge this now. it is a part of v0.0.21 , so if that gets approved, let's not merge and see if v0.0.22, without, will get approved too

epheph:
Got accepted v0.0.25 without, closing

Issue #139, opened 2023-01-30
For all popup windows, add title and icon of the website that is doing the request


KillariDev:

KillariDev:
Added

PR #178, opened 2023-02-15
Bring website origin and icon to confirm tx dialog

KillariDev:
This allows to show user which website is requesting the transaction

Work towards https://github.com/DarkFlorist/TheInterceptor/issues/139


MicahZoltu (approved):
Website title would also be nice.  I use IPFS for many sites, and seeing the origin alone doesn't really help me a whole lot.

PR #187, opened 2023-02-22
add website type

KillariDev:
- Add website title with icon and origin
- rename `origin` to `websiteOrigin` where its possible to avoid accidentally using `origin` global variable
- replace placeholder title with real title
- add website information to transactions on where the transaction originated from


extension/app/ts/simulation/services/SimulationModeEthereumClientService.ts
```diff
@@ -19,12 +20,18 @@ type Subscription = {
 	rpcSocket: WebSocket
 }
 
+export type EthereumUnsignedTransactionWithWebsite = EthereumUnsignedTransaction & { website: Website }
```
MicahZoltu, line 23: I am not a huge fan of this.  I would rather see an envelope type that contains the website and the transaction as separate parts:
```ts
export type EnvelopeNameTbd = { transaction: EthereumUnsignedTransaction, website: Website }
```

MicahZoltu (approved):

PR #192, opened 2023-02-23
`on` and `removeListener` should return `window.ethereum`

KillariDev:


extension/app/inpage/ts/inpage.ts
```diff
@@ -194,52 +194,54 @@ class InterceptorMessageListener {
...
+				break
 			case 'disconnect':
 				this.onDisconnectCallBacks.add(callback as (error: ProviderRpcError) => void)
-				return
+				break
 			case 'chainChanged':
 				this.onChainChangedCallBacks.add(callback as (chainId: string) => void)
-				return
+				break
 			default:
```
MicahZoltu, line 217: Since you are switching from `return` to `break` you should do an exhaustiveness check in `default`.  Alternatively, just `return window.ethereum` from each case.
```suggestion
			default: assertNever(kind)
```

extension/app/inpage/ts/inpage.ts
```diff
@@ -194,52 +194,54 @@ class InterceptorMessageListener {
...
+				break
 			case 'disconnect':
 				this.onDisconnectCallBacks.delete(callback as (error: ProviderRpcError) => void)
-				return
+				break
 			case 'chainChanged':
 				this.onChainChangedCallBacks.delete(callback as (chainId: string) => void)
-				return
+				break
 			default:
```
MicahZoltu, line 242:
```suggestion
			default: assertNever(kind)
```

MicahZoltu (approved):

extension/app/inpage/ts/inpage.ts
KillariDev: I tried to look from the docs what is the correct response, but metamask does not error on any strings

PR #197, opened 2023-02-24
Unify transaction identification and naming

KillariDev:


extension/app/ts/components/simulationExplaining/identifyTransaction.tsx
```diff
@@ -3,55 +3,82 @@ import { CHAINS, FourByteExplanations, isSupportedChain, MAKE_YOU_RICH_TRANSACTI
...
-	if (identifyTransaction(transaction, activeAddress) === 'MakeYouRichTransaction') {
-		return 'Rich'
+	if (isSupportedChain(chainString)
+		&& CHAINS[chainString].eth_donator === transaction.from.address
+		&& transaction.to?.address === activeAddress
+		&& transaction.type === MAKE_YOU_RICH_TRANSACTION.type
+		&& transaction.maxFeePerGas === MAKE_YOU_RICH_TRANSACTION.maxFeePerGas
+		&& transaction.maxPriorityFeePerGas === MAKE_YOU_RICH_TRANSACTION.maxPriorityFeePerGas
+		&& transaction.input.toString() === MAKE_YOU_RICH_TRANSACTION.input.toString()
+		&& transaction.value === MAKE_YOU_RICH_TRANSACTION.value
```
MicahZoltu, line 26: Similar to comment on previous PR, I would much rather see an envelope that contained the transaction plus some metadata about it, rather than trying to infer the transaction type based on the transaction details.

That being said, we will have to do this inference at some point for many transaction types (e.g., Swaps), so doing it for all of them isn't terrible.  Even in that case though, I feel like we should have an envelope that stores that information so we don't have to infer multiple times.

```ts
const envelope = { transaction, website, type }
```

MicahZoltu (approved):

extension/app/ts/components/simulationExplaining/identifyTransaction.tsx
```diff
@@ -3,55 +3,82 @@ import { CHAINS, FourByteExplanations, isSupportedChain, MAKE_YOU_RICH_TRANSACTI
...
-	if (transaction.input.length == 0) return 'Ether Transfer'
 
-	//TODO: add simple token transfer
+	if (transaction.input.length == 0) return {
+		type: 'EtherTransfer',
+		title: 'Ether Transfer',
+		signingAction: 'Transfer Ether',
+		simulationAction: 'Simulate Ether Transfer',
+		rejectAction: 'Reject Ether Transfer',
+	}
```
MicahZoltu, line 43: This isn't correct I don't think.  There are lots of contracts that have a receive function that takes some action when ETH is received even in the abscence of calldata.

You could *also* check that gasLimit is 21,000, but that is also wrong as you can attach extra gas, and sending ETH to a contract wallet would have a higher gas limit but would also be just a transfer.

At the least, we should track a GitHub issue to revisit this and figure out the best way to identify ETH transfers.  If we simulate first, we could check to see how much gas was *used*, which would notably help identify "complex transaction" vs "simple send".  There is probably some threshold of gas used where "something interesting happened on receipt".

extension/app/ts/components/simulationExplaining/identifyTransaction.tsx
KillariDev: We are doing that inference for swaps, ether sends and token sends. That envelope would only be used to separate MAKE_YOU_RICH_TRANSACTION ? Its the only transaction that the Interceptor creates.

extension/app/ts/components/simulationExplaining/identifyTransaction.tsx
KillariDev: ah you are right. We could also check for produced logs

MicahZoltu: A contract wallet may produce logs on ETH received.

extension/app/ts/components/simulationExplaining/identifyTransaction.tsx
MicahZoltu: My thinking was to classify transactions once as soon as the Interceptor receives them, and then store that classification in the transaction envelope's metadata.  This way everywhere we want to know the type, we already have it.  Essentially, classification happens as part of the initial deserialization/creation of the transaction.

KillariDev: yeah, I do classification many times with the same function as its not that complex operation

extension/app/ts/components/simulationExplaining/identifyTransaction.tsx
KillariDev: I mean token logs, if it produces token logs, I won't categorize it as eth send

extension/app/ts/components/simulationExplaining/identifyTransaction.tsx
KillariDev: hmm, doing classification "when we get the transaction" won't do it, as it can change. We need to do it every time we run simulation again. Unless we make the categorization not dependent on the simulation, which would be safer, but more difficult

MicahZoltu: The classification of a transaction changing between simulations *feels* like an edge case that we shouldn't worry about/deal with.  Do you have some thoughts on a reasonable way that would happen?

If not, then we can do classification once when we first see the transaction and then store the classification in the envelope and never do it again.

KillariDev: the only reasonable way is when the transaction fails. You don't get simulation results so you cannot classify it anymore. This happens quite often with uniswap swaps timeoutting. But then the initial classification would be correct. Its a swap that now fails

PR #200, opened 2023-02-24
Send tokens and refactor custom UI handling

KillariDev:
- Fix: https://github.com/DarkFlorist/TheInterceptor/issues/189
- Refactor custom UI code to be more modular and with less duplication
- We currently cannot show token balances in this UI (opened issue: https://github.com/DarkFlorist/TheInterceptor/issues/199 )
image


extension/app/ts/components/simulationExplaining/customExplainers/SimpleSendVisualisations.tsx
```diff
@@ -0,0 +1,139 @@
...
+}
+
+function getBeforeAndAfterBalanceForAddress(ethBalances: readonly EthBalanceChangesWithMetadata[], address: bigint) {
+	const filtered = ethBalances.filter((x) => x.address.address === address)
+	if (filtered.length === 0) {
+		return {
+			before: 0n,
+			after: 0n,
+		}
+	}
```
MicahZoltu, line 89: What is the situation where this will be length 0?

Also, I wouldn't mind seeing this slimmed down, 6 lines seems excessive for a guard clause:
```suggestion
	if (filtered.length === 0) return { before: 0n, after: 0n }
```

extension/app/ts/components/simulationExplaining/identifyTransaction.tsx
```diff
@@ -15,7 +15,7 @@ type IdenttifiedTransaction = {
 
 export function identifyTransaction(transaction: SimulatedAndVisualizedTransaction, activeAddress: bigint): IdenttifiedTransaction {
 	const chainString = transaction.chainId.toString()
-
+	console.log(transaction)
```
MicahZoltu, line 18:
```suggestion
```

MicahZoltu (approved):

extension/app/ts/components/simulationExplaining/customExplainers/SimpleSendVisualisations.tsx
KillariDev: It should never happen, I could make it throw an error too

extension/app/ts/components/simulationExplaining/identifyTransaction.tsx
KillariDev: noo, my debug print!

PR #204, opened 2023-02-27
Simple token approval UI

KillariDev:
fixes https://github.com/DarkFlorist/TheInterceptor/issues/190

image

### DIFF
--- a/extension/app/ts/components/simulationExplaining/customExplainers/SimpleTokenApprovalVisualisation.tsx
+++ b/extension/app/ts/components/simulationExplaining/customExplainers/SimpleTokenApprovalVisualisation.tsx
@@ -1,0 +1,63 @@
+import { RenameAddressCallBack } from '../../../utils/user-interface-types.js'
+import { SimulatedAndVisualizedTransaction } from '../../../utils/visualizer-types.js'
+import { BigAddress } from '../../subcomponents/address.js'
+import { Token721AmountField, TokenAmount, TokenSymbol } from '../../subcomponents/coins.js'
+import { GasFee } from '../SimulationSummary.js'
+
+export function SimpleTokenApprovalVisualisation({ transaction, renameAddressCallBack }: { transaction: SimulatedAndVisualizedTransaction, renameAddressCallBack: RenameAddressCallBack }) {
+	if (transaction.to === undefined) throw new Error('Contract creation trasaction')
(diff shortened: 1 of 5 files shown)


jubalm (approved):

extension/app/ts/components/simulationExplaining/customExplainers/SimpleTokenApprovalVisualisation.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { BigAddress } from '../../subcomponents/address.js'
+import { Token721AmountField, TokenAmount, TokenSymbol } from '../../subcomponents/coins.js'
+import { GasFee } from '../SimulationSummary.js'
+
+export function SimpleTokenApprovalVisualisation({ transaction, renameAddressCallBack }: { transaction: SimulatedAndVisualizedTransaction, renameAddressCallBack: RenameAddressCallBack }) {
+	if (transaction.to === undefined) throw new Error('Contract creation trasaction')
+	if (transaction.tokenResults.length != 1) throw new Error('Multiple token events transfer')
+
+	const approval = transaction.tokenResults[0]
+	if (!approval.isApproval) throw new Error('Is Not Approval')
```
MicahZoltu, line 12: Where is this error caught?  Is it presented?

MicahZoltu (approved):
Would like to better understand error handling.

extension/app/ts/components/simulationExplaining/customExplainers/SimpleTokenApprovalVisualisation.tsx
KillariDev: this should not never occur as its tested beforehand, so its not caught anywhere. If its thrown, something has wrong elsewhere. I guess one option would be to add it to the typing so we wouldn't need to check it

MicahZoltu: There are several errors being thrown, are these all critical programmer errors and not something we should reasonably be able to deal with?

KillariDev: which can be fixed with typings, but I was lazy :D

MicahZoltu: Recommending fixing with type checking then.  😀

PR #206, opened 2023-02-27
Refactor second to last getBackground call away

KillariDev:
- Also replace Page enum with strings
- Also put `port.postMessage(message)` to try catch in case we are about to send to message to a popup that gets closed
- Fixes a bug caused by previous PR 😬


extension/app/ts/background/background.ts
```diff
@@ -444,7 +444,16 @@ export function postMessageIfStillConnected(port: browser.runtime.Port, message:
...
+	try {
+		port.postMessage(message)
+	} catch (error) {
+		if (error instanceof Error) {
+			if (error.message?.includes('Attempting to use a disconnected port object')) {
+				return
+			}
+		}
+		throw error
+	}
```
MicahZoltu, line 456: I'm pretty strongly against using errors for control flow.  Is there any other way to accomplish this without relying on catching a specific error?  Is there some way we can check the port before posting to it?

For example, can we do something like this:
```ts
port.onDisconnect(() => port = undefined)
```
and then when we use the port do `port?.postMessage(message)`?

extension/app/ts/simulation/simulator.ts
```diff
@@ -54,7 +54,7 @@ export class Simulator {
 		for (let i = 0; i < transactions.length; i++) {
 			resultPromises.push(this.visualizeTransaction(transactions[i], blockNumber, multicallResults[i]))
 		}
-		return Promise.all(resultPromises)
+		return await Promise.all(resultPromises)
```
MicahZoltu, line 57: Technically, I believe these are the same thing.  If you return a Promise from an async function, it "does the right thing" and adds the returned promise as a continuation.
```ts
async function apple() { return new Promise(resolve => setTimeout(() => resolve(1), 1000)) } 
await apple()
// <1 second later>
// <- 1
```

extension/app/ts/utils/interceptor-messages.ts
```diff
@@ -434,6 +446,83 @@ export const ConfirmTransactionSimulationStateChanged = funtypes.Object({
...
+	simulationMode: funtypes.Boolean,
+	pendingAccessRequests: PendingAccessRequestArray,
+	userAddressBook: UserAddressBook,
+})
+
+export type UpdateHomePage = funtypes.Static<typeof UpdateHomePage>
+export const UpdateHomePage = funtypes.Object({
+	method: funtypes.Literal('popup_UpdateHomePage'),
+	data: funtypes.Object({
+		simulation: funtypes.Object({
```
MicahZoltu, line 508: Should these be readonly?  Same for other objects above.  Unless we specifically need a mutable object, I think we should be using ReadonlyObject and ReadonlyArray everywhere (looks like Array is already Readonly in most places).
```suggestion
export const UpdateHomePage = funtypes.ReadonlyObject({
	method: funtypes.Literal('popup_UpdateHomePage'),
	data: funtypes.ReadonlyObject({
		simulation: funtypes.ReadonlyObject({
```

MicahZoltu (approved):

extension/app/ts/background/background.ts
KillariDev: this error happens in race conditions only so I don't think there's a way

extension/app/ts/simulation/simulator.ts
KillariDev: yeah, I think they are the exactly the same. It somehow feels right to add Await there 🤔

extension/app/ts/utils/interceptor-messages.ts
KillariDev: yeah I guess we should have ReadonlyObject everywhere

extension/app/ts/background/background.ts
MicahZoltu: Hmm, race conditions shouldn't be possible in JavaScript without async code.  It is *possible* that the `onDisconnect` callback isn't called until the next frame after the disconnect, but that would surprise me a bit.  Usually in JS libraries the callbacks are called inline.

extension/app/ts/simulation/simulator.ts
MicahZoltu: I agree, the `await` is good as you should generally always await promises.  Just commenting that it isn't strictly necessary.

PR #216, opened 2023-02-28
Fix refresh bug in accesslist

KillariDev:
Fix bug when access rights update in middle of users editing process causing user to lose the changes on access rights

### DIFF
--- a/extension/app/ts/components/pages/InterceptorAccessList.tsx
+++ b/extension/app/ts/components/pages/InterceptorAccessList.tsx
@@ -238,12 +238,12 @@ export function InterceptorAccessList(param: InterceptorAccessListParams) {
 										{ access.addressAccess.length === 0 ? <p className = 'paragraph'> No individual address accesses given </p> : <>
 											{ access.addressAccessModified.map( (websiteAccessAddress, addressIndex) => (
 												<li style = { `margin: 0px; margin-bottom: ${ addressIndex < access.addressAccessModified.length - 1  ? '10px;' : '0px' }` }>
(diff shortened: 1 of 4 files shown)


extension/app/ts/components/pages/InterceptorAccessList.tsx
```diff
@@ -195,12 +238,12 @@ export function InterceptorAccessList(param: InterceptorAccessListParams) {
 										{ access.addressAccess.length === 0 ? <p className = 'paragraph'> No individual address accesses given </p> : <>
 											{ access.addressAccessModified.map( (websiteAccessAddress, addressIndex) => (
 												<li style = { `margin: 0px; margin-bottom: ${ addressIndex < access.addressAccessModified.length - 1  ? '10px;' : '0px' }` }>
-													{ websiteAccessAddress.removed ? <p style = 'color: var(--negative-color)' > { `Forgot ${ websiteAccessAddress.address }`} </p> :
+													{ websiteAccessAddress.removed ? <p style = 'color: var(--negative-color)' > { `Forgot ${ ethers.utils.getAddress(addressString(websiteAccessAddress.address)) }`} </p> :
```
MicahZoltu, line 241: Why are we doing bigint => string => string?  Is it because we ultimately need a checksummed string and there is no method for direct from bigint to checksummed string?  If so, then perhaps there is value in creating such a helper as it seems like a common need?

MicahZoltu (approved):

extension/app/ts/components/pages/InterceptorAccessList.tsx
KillariDev: yeah thats the reason, yeah, we could create helper for it

PR #220, opened 2023-03-07
Confirm transaction saving

KillariDev:
- Reconnects after service worker disconnects
- Instead of using sockets to identify connection, use connection name (generated randomly) and tab id
- Saves confirm transaction dialog information to disk so on a reconnect, so if service worker is lost when the dialog is open, we can recover

still todos: other dialogs, saving other state stuff (like simulation state)

### DIFF
--- a/extension/app/inpage/ts/document_start.ts
+++ b/extension/app/inpage/ts/document_start.ts
@@ -1,18 +1,27 @@
-function listenInContentScript() {
+function listenInContentScript(conectionName: string | undefined) {
 	/**
 	 * this script executed within the context of the active tab when the user clicks the extension bar button
 	 * this script serves as a _very thin_ proxy between the page scripts (dapp) and the extension, simply forwarding messages between the two
 	*/
 	// the content script is a very thin proxy between the background script and the page script
-	const extensionPort = browser.runtime.connect()
-	let connected = true
+
+	const dec2hex = (dec: number) => dec.toString(16).padStart(2, '0')
+
+	function generateId (len: number) {
+		const arr = new Uint8Array((len || 40) / 2)
+		globalThis.crypto.getRandomValues(arr)
+		return Array.from(arr, dec2hex).join('')
+	}
+	const connectionNameNotUndefined = conectionName === undefined ? generateId(40) : conectionName
+
+	const extensionPort = browser.runtime.connect({ name: connectionNameNotUndefined })
 
 	// forward all message events to the background script, which will then filter and process them
-	globalThis.addEventListener('message', messageEvent => {
+	const listener = (messageEvent: MessageEvent<any>) => {
 		try {
 			// we only want the data element, if it exists, and postMessage will fail if it can't clone the object fully (and it cannot clone a MessageEvent)
 			if (!('data' in messageEvent)) return
-			if (connected) extensionPort.postMessage({ data: messageEvent.data })
+			extensionPort.postMessage({ data: messageEvent.data })
 		} catch (error) {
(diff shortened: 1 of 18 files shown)


extension/app/ts/listenContentScript.ts
```diff
@@ -1,18 +1,27 @@
...
-	globalThis.addEventListener('message', messageEvent => {
+	const listener = (messageEvent: MessageEvent<any>) => {
 		try {
 			// we only want the data element, if it exists, and postMessage will fail if it can't clone the object fully (and it cannot clone a MessageEvent)
 			if (!('data' in messageEvent)) return
-			if (connected) extensionPort.postMessage({ data: messageEvent.data })
+			extensionPort.postMessage({ data: messageEvent.data })
 		} catch (error) {
 			// CONSIDER: should we catch data clone error and then do `extensionPort.postMessage({data:JSON.parse(JSON.stringify(messageEvent.data))})`?
 			if (error instanceof Error) {
```
0xjimmy, line 27: Same here

extension/app/ts/background/simulationModeHanders.ts
```diff
@@ -55,20 +55,27 @@ export async function sendTransaction(getActiveAddressForDomain: (websiteAccess:
...
 			nonce: await transactionCount,
-			maxFeePerGas: request.params[0].maxFeePerGas ? request.params[0].maxFeePerGas : maxFeePerGas,
-			maxPriorityFeePerGas: request.params[0].maxPriorityFeePerGas ? request.params[0].maxPriorityFeePerGas : 1n,
-			gas: request.params[0].gas ? request.params[0].gas : 90000n,
-			to: request.params[0].to ? request.params[0].to : 0n,
-			value: request.params[0].value ? request.params[0].value : 0n,
-			input: 'data' in request.params[0] && request.params[0].data !== undefined ? request.params[0].data : new Uint8Array(),
+			maxFeePerGas: sendTransactionParams.params[0].maxFeePerGas ? sendTransactionParams.params[0].maxFeePerGas : maxFeePerGas,
+			maxPriorityFeePerGas: sendTransactionParams.params[0].maxPriorityFeePerGas ? sendTransactionParams.params[0].maxPriorityFeePerGas : 1n,
+			gas: sendTransactionParams.params[0].gas ? sendTransactionParams.params[0].gas : 90000n,
```
0xjimmy, line 60: Not related pr but why is default gas amount 9000?

0xjimmy (approved):

extension/app/inpage/ts/document_start.ts
```diff
@@ -1,18 +1,27 @@
...
-	globalThis.addEventListener('message', messageEvent => {
+	const listener = (messageEvent: MessageEvent<any>) => {
 		try {
 			// we only want the data element, if it exists, and postMessage will fail if it can't clone the object fully (and it cannot clone a MessageEvent)
 			if (!('data' in messageEvent)) return
-			if (connected) extensionPort.postMessage({ data: messageEvent.data })
+			extensionPort.postMessage({ data: messageEvent.data })
 		} catch (error) {
 			// CONSIDER: should we catch data clone error and then do `extensionPort.postMessage({data:JSON.parse(JSON.stringify(messageEvent.data))})`?
 			if (error instanceof Error) {
```
0xjimmy, line 27: Can combine if branches to one

extension/app/ts/background/simulationModeHanders.ts
KillariDev: hmm, I don't remember. Good catch, I'll create issue about it

PR #223, opened 2023-03-09
Other popups working

KillariDev:
- Make other popups to work with service worker too
- Make interceptor to reply every message from content script to be sure in content script that the message was received (this also makes it so that requestID is now mandatory field)
-  Rework of the permission system. Now we just have permissions stored into only one place: `websiteTabConnections`
-  Rework content script message handling, now permission dialog gatekeeping is separated to its own path
- use `WebsiteSocket` type for differentiating between connections. This makes it easier to reconnect to webpages when webworker goes down


0xjimmy (approved):

extension/app/ts/background/popupMessageHandlers.ts
```diff
@@ -266,7 +263,13 @@ export async function homeOpened() {
...
 		}
 	})
 }
+
+export async function interceptorAccessChangeAddress(params: InterceptorAccessChangeAddress) {
+	requestAddressChange(params)
+}
+export async function interceptorAccessRefresh(params: InterceptorAccessRefresh) {
+	requestAddressChange(params)
+}
```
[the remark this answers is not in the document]
KillariDev: they used to be different but I refactored :). I merged them together

extension/app/ts/background/windows/interceptorAccess.ts
```diff
@@ -137,114 +133,138 @@ export async function requestAccessFromUser(port: browser.runtime.Port | undefin
...
+	}
+
+	browser.runtime.onMessage.addListener(windowReadyAndListening)
+
+	openedInterceptorAccessWindow = await browser.windows.create(
+		{
+			url: getHtmlFile('interceptorAccess'),
+			type: 'popup',
+			height: 600,
+			width: 600,
```
MicahZoltu, line 179: This seems unnecessarily small, and I don't think will ever fit everything.  Can we make the default bigger?

extension/package-lock.json
```diff
@@ -19,6 +18,7 @@
 				"webextension-polyfill": "0.10.0"
 			},
 			"devDependencies": {
+				"@types/chrome": "^0.0.218",
```
MicahZoltu, line 21: Why did this change without `package.json` changing?

extension/package-lock.json
```diff
@@ -7,7 +7,6 @@
 			"name": "interceptor-extension",
 			"dependencies": {
 				"@darkflorist/address-metadata": "0.2.0",
-				"@types/chrome": "^0.0.218",
 				"@types/node-fetch": "2.6.1",
```
MicahZoltu, line 10: All `@types/*` should be a devDependency.

PR #233, opened 2023-03-21
Fixes access rights getting out of sync if you modify access rights on a popup at the same time with having them open at the same time

KillariDev:
Fixes https://github.com/DarkFlorist/TheInterceptor/issues/228


MicahZoltu (approved):
I don't fully grok why this fixes the problem, but it just deletes code which I'm always a fan of!

KillariDev:
> I don't fully grok why this fixes the problem, but it just deletes code which I'm always a fan of!

https://github.com/DarkFlorist/TheInterceptor/pull/233/files#diff-bf8795aeb0e6a7c3c92c4859d3608f9dac850f6cb2925b307034669f3ce6d97dR71

Its this line, I forgot to update the variable so old variable remains, causing those two arrays to be out of sync from each other

Issue #226, opened 2023-03-13
Implement eth_getLogs in simulation mode.


MicahZoltu:
This is a pretty commonly used function by dapps, so we should support it in simulation mode.  For getting logs for blocks older than `latest + 1`, we can just fallthrough to the node.  For logs in block `latest + 1` (the simulated block) we will need to keep track of the logs returned by each call in the multicall and manually apply the getLogs filters to them.

PR #235, opened 2023-03-21
`getLogs`, topics for logs, `getTransactionCount` bug

KillariDev:
Fixes https://github.com/DarkFlorist/TheInterceptor/issues/226

### DIFF
--- a/extension/app/ts/simulation/services/SimulationModeEthereumClientService.ts
+++ b/extension/app/ts/simulation/services/SimulationModeEthereumClientService.ts
@@ -459,8 +459,22 @@ export class SimulationModeEthereumClientService {
 				}
 			}
 		}
-		//TODO: handle other filter options
-		return EthGetLogsResponse.parse(events).filter((x) => logFilter.address === undefined || x.address === logFilter.address)
+
+		const includeLogByTopic = (logsTopics: readonly bigint[], filtersTopics: readonly (bigint | readonly bigint[] | null)[] | undefined) => {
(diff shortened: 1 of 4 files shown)


extension/app/ts/simulation/services/SimulationModeEthereumClientService.ts
```diff
@@ -471,7 +485,7 @@ export class SimulationModeEthereumClientService {
 		}
 		if('fromBlock' in logFilter) {
 			const logs = await this.ethereumClientService.getLogs(logFilter)
-			if(logFilter.toBlock === 'latest' || logFilter.toBlock === await this.getBlockNumber() ) {
+			if (logFilter.toBlock === 'latest' || logFilter.toBlock > await this.ethereumClientService.getBlockNumber() ) {
 				return [...logs, ...await this.getSimulatedLogs(logFilter)]
 			}
 			return logs
```
MicahZoltu, line 491: This feels racey.  Can we instead do something like:
```suggestion
			const logs = await this.ethereumClientService.getLogs(logFilter)
			const [simulatedBlockNumber, simulatedLogs] = await this.getSimulatedLogs(logFilter)
			if (logFilter.toBlock === 'latest' || logFilter.toBlock >= simulatedBlockNumber) {
				return [...logs, ...await simulatedLogs]
			} else {
				return logs
			}
```
This would ensure that we aren't getting logs for a different "block number" than the block number we test against to see if we *should* get logs.

MicahZoltu (approved):
Tentatively approved, though I really do feel like the potential race condition should be addressed.

PR #239, opened 2023-03-21
Identify and show contract deployment

KillariDev:


MicahZoltu (approved):
Screenshot or it didn't happen.

 Also, I feel like we should list all contracts created by the transaction, not just the root contract deployment.  As a simple example, anyone using Deterministic Deployment Proxy wouldn't see a contract created.

My guess is that almost no one creates a transaction directly (via `to: null`) these days, everyone uses some proxy (e.g., Gnosis Safe, Maker Vaults, etc.)

KillariDev:
> Screenshot or it didn't happen.
> 
> Also, I feel like we should list all contracts created by the transaction, not just the root contract deployment. As a simple example, anyone using Deterministic Deployment Proxy wouldn't see a contract created.
> 
> My guess is that almost no one creates a transaction directly (via `to: null`) these days, everyone uses some proxy (e.g., Gnosis Safe, Maker Vaults, etc.)

yeah, I am not sure how we should do this

KillariDev:
> Screenshot or it didn't happen.
> 
> Also, I feel like we should list all contracts created by the transaction, not just the root contract deployment. As a simple example, anyone using Deterministic Deployment Proxy wouldn't see a contract created.
> 
> My guess is that almost no one creates a transaction directly (via `to: null`) these days, everyone uses some proxy (e.g., Gnosis Safe, Maker Vaults, etc.)

image

MicahZoltu:
We should include the contract address (copyable) and blockie in this UI I think.

PR #246, opened 2023-03-22
Don't throw errors for transaction rejection (just forward it to dapp) and refactoring

KillariDev:


extension/app/inpage/output/injected_document_start.js
MicahZoltu, line 1: Is this file purely a generated file?  If so, can we remove it from `.git` so it is just generated on build instead?

MicahZoltu (approved):

extension/app/inpage/output/injected_document_start.js
KillariDev: yeah its generated

PR #249, opened 2023-03-22
fix ethers `parseTransaction` parsing and token decimal fetching simulation

KillariDev:


extension/app/ts/components/formVisualizerResults.ts
```diff
@@ -42,8 +41,8 @@ export function formSimulatedAndVisualizedTransaction(simState: SimulationState,
 					token: tokenEntry
 				}
 			}
-			throw new Error('wrong token type')
-		})
+			return undefined // a token that is not NFT, but does not have decimals either, let's just not visualize them
+		}).filter((x): x is TokenVisualizerResultWithMetadata => x !== undefined)
```
MicahZoltu, line 45: Syntax may be a bit off here, but this is a more accurate signature and will help protect against bugs by avoiding blind typecasting.
```suggestion
		}).filter(<T>(x: T | undefined): x is T => x !== undefined)
```

extension/app/ts/utils/bigint.ts
```diff
@@ -140,7 +140,7 @@ export function squareRoot(value: bigint) {
 
 export function stringifyJSONWithBigInts(value: any): string {
 	return JSON.stringify(value, (_key, value) => {
-		return typeof value === "bigint" ? `0x${value.toString(16)}n` : value
+		return typeof value === "bigint" ? `0x${ value.toString(16) }` : value
```
MicahZoltu, line 143: The `n` here makes it so you can automatically identify which hex strings are bigints and which are binary data.  It allows us to put something like this into a JSON.parse decoder:
```ts
if (typeof x === 'string') {
	if (/^0x[a-fA-F0-9]+n$/.test(x)) return BigInt(x)
	else if (/^0x[a-fA-F0-9]+$/.test(x)) return hexStringToUint8Array(x)
} else {
	decodeAsNormalJson(x)
}
```

extension/app/ts/simulation/services/SimulationModeEthereumClientService.ts
```diff
@@ -529,6 +529,26 @@ export class SimulationModeEthereumClientService {
...
+			to: token,
+			input: balanceOfCallData,
+			maxFeePerGas: 0n,
+			maxPriorityFeePerGas: 0n,
+			gas: 21000n,
+			value: 200000000000000000000000n,
+			accessList: [],
+			nonce: 0n,
+			chainId: await this.getChainId()
+		}
```
MicahZoltu, line 547: Several of these should be optional and don't really make sense for `call`s.  We should also use `as const` to help the type checker along (just as a general best practice when defining constant objects).
```suggestion
		const callTransaction = {
			type: '1559',
			from: MOCK_ADDRESS,
			to: token,
			input: balanceOfCallData,
			value: 200000000000000000000000n,
		} as const
```

MicahZoltu (approved):

extension/app/ts/simulation/services/SimulationModeEthereumClientService.ts
KillariDev: We are not actually making a call here, but a multicall, in which case we need to convert the call to a transaction. I have done that in a way that call requires a transaction directly to have the same API

extension/app/ts/utils/bigint.ts
KillariDev: the problem with that identification is that if you write a string that is not intended to be bigint, it just becomes bigint. Like for example (not actually done) you have a token name "0xn" would become come bigint 0, while its supposed to be a name of a token.

extension/app/ts/simulation/services/SimulationModeEthereumClientService.ts
KillariDev: hmm actually I can just Pick here

extension/app/ts/utils/bigint.ts
MicahZoltu: While I agree it is possible to have a string like `0x0n` that was meant to just be that string, it seems *incredibly* unlikely and not worth the cost of choosing to have byte arrays and numbers encoded the same way (thus cannot be auto-decoded).

Issue #248, opened 2023-03-22
swapping tiny amounts overflows


KillariDev:
![image](https://user-images.githubusercontent.com/13102010/226948939-8ae3addd-de6d-41de-a911-bd2644577b0a.png)

MicahZoltu:
Can you just make the page wider?  No desktop user (where our extension is used) has a screen that small...

PR #254, opened 2023-03-23
Fix importance box overflow

KillariDev:
image

fixes https://github.com/DarkFlorist/TheInterceptor/issues/248

### DIFF
--- a/extension/app/css/interceptor.css
+++ b/extension/app/css/interceptor.css
@@ -618,6 +618,7 @@ svg.spinner > circle {
 	padding: 10px;
 	margin: auto;
 	margin-bottom: 10px;
-	max-width: 400px;
+	min-width: 400px;
+	max-width: fit-content;
 	color: var(--text-color)
 }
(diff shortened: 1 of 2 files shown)


MicahZoltu (approved):

MicahZoltu:
Why min-width?  It feels like we could allow the user to shrink as much as they want?

KillariDev:
> Why min-width? It feels like we could allow the user to shrink as much as they want?

because I don't feel this looks nice
image

KillariDev:
I guess ideally it should prefer that 400px size, but if it cannot make it, it would flex in different directions

KillariDev:
I am not sure how to achieve that tho

MicahZoltu:
If you want it to fill the container, then you should have it fill the container.  I agree with you that in your screenshot it shouldn't look like that, but I think min-width is the wrong solution.  Flexboxes/grids and whatnot are (though they are tricky to grok).

PR #255, opened 2023-03-23
use clearly relative path

KillariDev:


MicahZoltu (approved):
Note: This technically might be a behavior change, though I don't know if it matters in this case.

PR #256, opened 2023-03-23
store simulation results to protect against worker shutdown

KillariDev:
Fixes https://github.com/DarkFlorist/TheInterceptor/issues/251

This also gets rid of more global variables

### DIFF
--- a/extension/app/ts/background/settings.ts
+++ b/extension/app/ts/background/settings.ts
@@ -1,6 +1,7 @@
 import { MOCK_PRIVATE_KEYS_ADDRESS } from '../utils/constants.js'
 import { AddressBookTabIdSetting, LegacyWebsiteAccessArray, Page, PendingAccessRequestArray, PendingChainChangeConfirmationPromise, PendingInterceptorAccessRequestPromise, PendingPersonalSignPromise, PendingUserRequestPromise, Settings, WebsiteAccessArray, WebsiteAccessArrayWithLegacy, pages } from '../utils/interceptor-messages.js'
 import { Semaphore } from '../utils/semaphore.js'
+import { browserStorageLocalGet } from '../utils/typescript.js'
 import { AddressInfo, ContactEntries } from '../utils/user-interface-types.js'
 import { SimulationResults } from '../utils/visualizer-types.js'
 import { EthereumAddress, EthereumQuantity } from '../utils/wire-types.js'
@@ -163,7 +164,7 @@ export async function savePendingInterceptorAccessRequestPromise(promise: Pendin
 
 const SIMULATION_RESULTS_STORAGE_KEY = 'simulationResults'
 export async function getSimulationResults() {
-	const results = await browser.storage.local.get([SIMULATION_RESULTS_STORAGE_KEY]) as unknown
+	const results = await browserStorageLocalGet(['simulationResults'])
 	if (results.simulationResults === undefined) {
 		return {
 			simulationId: 0,
@@ -178,11 +179,11 @@ export async function getSimulationResults() {
 }
 
 const simulationResultsSemaphore = new Semaphore(1)
-export async function updateSimulationResults(simulationResults: SimulationResults) {
+export async function updateSimulationResults(newResults: SimulationResults) {
 	simulationResultsSemaphore.execute(async () => {
-		const results = await browser.storage.local.get([SIMULATION_RESULTS_STORAGE_KEY]) as unknown
+		const oldResults = await getSimulationResults()
 		// TODO: use funtypes to parse/validate `results` here.  `browser.storage.local.get` returns an `any` I think (which is terrible)
-		if (results.simulationResults !== undefined && simulationResults.simulationId < results.simulationResults.simulationId) return // do not update state with older state
-		return await browser.storage.local.set({ [SIMULATION_RESULTS_STORAGE_KEY]: SimulationResults.serialize(simulationResults) })
+		if (newResults.simulationId < oldResults.simulationId) return // do not update state with older state
+		return await browser.storage.local.set({ [SIMULATION_RESULTS_STORAGE_KEY]: SimulationResults.serialize(newResults) })
 	})
 }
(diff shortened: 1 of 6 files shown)


extension/app/ts/utils/semaphore.ts
```diff
@@ -0,0 +1,159 @@
+// https://github.com/Zoltu/ethereum-ledger/blob/master/source/vendor/semaphore-async-await/semaphore.ts
```
MicahZoltu, line 1:
```suggestion
// https://github.com/jsoendermann/semaphore-async-await/blob/master/src/Semaphore.ts
```

extension/app/ts/background/settings.ts
```diff
@@ -159,3 +161,26 @@ export async function savePendingInterceptorAccessRequestPromise(promise: Pendin
...
+}
+
+const simulationResultsSemaphore = new Semaphore(1)
+export async function updateSimulationResults(simulationResults: SimulationResults) {
+	simulationResultsSemaphore.execute(async () => {
+		const results = await browser.storage.local.get(['simulationResults'])
+		if (results.simulationResults !== undefined && simulationResults.simulationId < results.simulationResults.simulationId) return // do not update state with older state
+		return await browser.storage.local.set({ simulationResults: SimulationResults.serialize(simulationResults) })
+	})
+}
```
MicahZoltu, line 186: Recommend using a constant for this type of thing so you don't have to worry about typos.

Also don't use `any`, and when a library uses it immediately cast to `unknown` to protect yourself from the terrible library.
```suggestion
const SIMULATION_RESULTS_STORAGE_KEY = 'simulationResults'
export async function getSimulationResults() {
	const results = await browser.storage.local.get([SIMULATION_RESULTS_STORAGE_KEY]) as unknown
	if (results.simulationResults === undefined) {
		return {
			simulationId: 0,
			simulationState: undefined,
			visualizerResults: undefined,
			addressBookEntries: [],
			tokenPrices: [],
			activeAddress: undefined
		}
	}
	return SimulationResults.parse(results.simulationResults)
}

const simulationResultsSemaphore = new Semaphore(1)
export async function updateSimulationResults(simulationResults: SimulationResults) {
	simulationResultsSemaphore.execute(async () => {
		const results = await browser.storage.local.get([SIMULATION_RESULTS_STORAGE_KEY]) as unknown
		// TODO: use funtypes to parse/validate `results` here.  `browser.storage.local.get` returns an `any` I think (which is terrible)
		if (results.simulationResults !== undefined && simulationResults.simulationId < results.simulationResults.simulationId) return // do not update state with older state
		return await browser.storage.local.set({ [SIMULATION_RESULTS_STORAGE_KEY]: SimulationResults.serialize(simulationResults) })
	})
}
```

MicahZoltu (changes requested):
Would like to see the `any` issue addressed before this is merged.  The constant is optional, but recommended.

extension/app/ts/utils/typescript.ts
```diff
@@ -49,3 +49,7 @@ export function isObject(maybe: unknown): maybe is Object {
 export function assertIsObject(maybe: unknown): asserts maybe is Object {
 	if (!isObject(maybe)) throw new Error(`Expected object but got ${ typeof maybe }`)
 }
+
+export async function browserStorageLocalGet(keys: string | string[]) {
+	return (await browser.storage.local.get(keys)) as { [key: string]: unknown }
```
MicahZoltu, line 54: Two minor style suggestions:
1. `Record<string, unknown>` instead of `{ [key: string]: unknown }`.  I believe they are synonymous, but I think `Record` looks a bit nicer and is a bit more clear.
2. You can get rid of the parenthesis by just changing the return type to `Promise<T>`.  I like this a bit better personally, but I can appreciate that it is longer and not everyone will agree.  I generally dislike a lot of parenthesis.
```suggestion
	return await browser.storage.local.get(keys) as Promise<Record<string, unknown>>
```

MicahZoltu: It is worth noting that the values that can correctly be stored are limited to `boolean, string, number, [], and {}`.  Essentially, only stuff that can be JSON encoded which means no `bigint, Map, Set, etc.`.  I wonder if it is worth creating a type for this that is more specific, and possibly a similar helper function for `browserStorageLocalSet`, that is type constrained?
```ts
type JsonPrimitive = string | number | boolean
type JsonType = JsonPrimitive | Array<JsonPrimitive> | Record<string, JsonPrimitive>
```

Alternatively, we could perhaps override the types for `browser.storage.local.get` and `browser.storage.local.set` so we don't need the helper functions?  You'll have to lookup how to do type overrides (and if it is possible in this context), I always forget how to do it.

MicahZoltu (approved):
Style comments are entirely optional.  The comment about further type constraining I think may be worth considering, but not a blocker for this PR.  It is just a way to help avoid shooting yourself in the foot later by getting the compiler to check that you aren't trying to store something that cannot be stored.

extension/app/ts/background/settings.ts
```diff
@@ -159,3 +162,28 @@ export async function savePendingInterceptorAccessRequestPromise(promise: Pendin
...
+		}
+	}
+	return SimulationResults.parse(results.simulationResults)
+}
+
+const simulationResultsSemaphore = new Semaphore(1)
+export async function updateSimulationResults(newResults: SimulationResults) {
+	simulationResultsSemaphore.execute(async () => {
+		const oldResults = await getSimulationResults()
+		// TODO: use funtypes to parse/validate `results` here.  `browser.storage.local.get` returns an `any` I think (which is terrible)
```
MicahZoltu, line 185: Looks like this was forgotten, see other comment about the actual types allowed though.
```suggestion
```

extension/app/ts/utils/typescript.ts
KillariDev: I created issue about this. I would rather use helper functions than override types & function calls as it can make it hard to understand why it behaves differently on this repo than in some other

PR #258, opened 2023-03-23
remove `globalThis.interceptor.websiteAccessAddressMetadata`

KillariDev:


extension/app/ts/background/popupMessageHandlers.ts
```diff
@@ -249,7 +249,7 @@ export async function homeOpened(simulator: Simulator) {
 		method: 'popup_UpdateHomePage',
 		data: {
 			simulation: await getSimulationResults(),
-			websiteAccessAddressMetadata: globalThis.interceptor.websiteAccessAddressMetadata,
+			websiteAccessAddressMetadata: getAddressMetadataForAccess(globalThis.interceptor.settings.websiteAccess),
```
MicahZoltu, line 252: Does it really count as "getting rid of" when you just move it into a sub-object?  😬

MicahZoltu (approved):

extension/app/ts/background/popupMessageHandlers.ts
KillariDev: I don't move it to sub object, I get rid of the metadata variable, and calculate it as needed from the actual accesses, which will be removed later (as it can be retrieved directly from storage)

PR #260, opened 2023-03-23
remove `globalThis.interceptor.pendingAccessMetadata`

KillariDev:


extension/app/ts/background/popupMessageHandlers.ts
```diff
@@ -236,12 +238,16 @@ export async function homeOpened(simulator: Simulator) {
 	const signerAccounts = signerState === undefined ? undefined : signerState.signerAccounts
 	const tabIconDetails = tabId === undefined ? undefined : globalThis.interceptor.websiteTabConnections.get(tabId)?.tabIconDetails
 
+	const pendingAccessRequestsAddresses = new Set(globalThis.interceptor.settings.pendingAccessRequests.map((x) => x.requestAccessToAddress === undefined ? [] : x.requestAccessToAddress).flat())
+	const addressInfos = globalThis.interceptor.settings.userAddressBook.addressInfos
```
MicahZoltu, line 242: As commented in the other PR, I'm not sure if it counts as getting rid of a global if you just move it into a subobject within an existing global.  That being said, I think this change is overall good and a step in the right direction.

MicahZoltu (approved):

extension/app/ts/background/popupMessageHandlers.ts
KillariDev: as I mentioned in the other PR, these are different variables

Issue #262, opened 2023-03-24
Remove Chrome bug 1316588 workaround code from backgroundServiceWorker.ts (when fixed)


KillariDev:
https://bugs.chromium.org/p/chromium/issues/detail?id=1316588

KillariDev:
related:
https://bugs.chromium.org/p/chromium/issues/detail?id=1271154
https://bugs.chromium.org/p/chromium/issues/detail?id=1426461

PR #263, opened 2023-03-24
Workaround for Chromium Issue 131658

KillariDev:
added todo as issue: https://github.com/DarkFlorist/TheInterceptor/issues/262

### DIFF
--- a/extension/app/ts/backgroundServiceWorker.ts
+++ b/extension/app/ts/backgroundServiceWorker.ts
@@ -28,3 +28,36 @@ const injectContentScript = async () => {
 }
 
 injectContentScript()
+
+/// Workaround for ChromiumIssue1316588 https://bugs.chromium.org/p/chromium/issues/detail?id=1316588
+// TODO, remove when chrome fixes the bug
+const storageArea = chrome.storage.local as chrome.storage.LocalStorageArea & { onChanged: chrome.storage.StorageChangedEvent }
+
+const TEST_INTERVAL_MS = 10000
+const STORAGE_WAIT_TIME_MS = 100
+
+const hasChromiumIssue1316588 = () => {
+	return new Promise((resolve) => {
+		let dispatched = false
+		const testEventDispatching = () => {
+			storageArea.onChanged.removeListener(testEventDispatching)
+			dispatched = true
+		}
+		storageArea.onChanged.addListener(testEventDispatching)
+		storageArea.set({ testEventDispatching: Math.random() })
+		setTimeout(() => resolve(!dispatched), STORAGE_WAIT_TIME_MS)
+	})
+}
+
+const fixChromiumIssue1316588 = () => {
+	hasChromiumIssue1316588().then((hasIssue) => {
+		if (hasIssue) {
+			chrome.runtime.reload()
+		} else {
+			setTimeout(fixChromiumIssue1316588, TEST_INTERVAL_MS)
+		}
+	})
+}
+
+fixChromiumIssue1316588()
+// End of workaround for ChromiumIssue1316588


extension/app/ts/backgroundServiceWorker.ts
```diff
@@ -28,3 +28,36 @@ const injectContentScript = async () => {
...
+		let dispatched = false
+		const testEventDispatching = () => {
+			storageArea.onChanged.removeListener(testEventDispatching)
+			dispatched = true
+		}
+		storageArea.onChanged.addListener(testEventDispatching)
+		storageArea.set({ testEventDispatching: Math.random() })
+		setTimeout(() => resolve(!dispatched), STORAGE_WAIT_TIME_MS)
+	})
+}
```
MicahZoltu, line 50: We have an `async sleep(milliseconds: number)` helper function that makes it so we can use async functions rather than having to do this sort of manual promise management.
```suggestion
const hasChromiumIssue1316588 = async () => {
	let dispatched = false
	const testEventDispatching = () => {
		storageArea.onChanged.removeListener(testEventDispatching)
		dispatched = true
	}
	storageArea.onChanged.addListener(testEventDispatching)
	storageArea.set({ testEventDispatching: Math.random() })
	await sleep(STORAGE_WAIT_TIME_MS)
	return !dispatched
}
```

extension/app/ts/backgroundServiceWorker.ts
```diff
@@ -28,3 +28,36 @@ const injectContentScript = async () => {
...
+
+const fixChromiumIssue1316588 = () => {
+	hasChromiumIssue1316588().then((hasIssue) => {
+		if (hasIssue) {
+			chrome.runtime.reload()
+		} else {
+			setTimeout(fixChromiumIssue1316588, TEST_INTERVAL_MS)
+		}
+	})
+}
```
MicahZoltu, line 60: Three suggestions here:
1. Use an async function instead of manual Promise stuff.
2. Use setInterval rather than queuing up a bunch of set timeouts in a row, if the intent is to do something every 10 seconds.
3. Don't silently swallow errors.
```suggestion
const intervalId = setInterval(fixChromiumIssue1316588, TEST_INTERVAL_MS)
const fixChromiumIssue1316588 = async () => {
	try {
		const hasIssue = await hasChromiumIssue1316588()
		if (!hasIssue) return
		chrome.runtime.reload()
		clearInterval(intervalId)
	} catch (error) {
		console.error(error)
	}
}
```

MicahZoltu (changes requested):
Left several notes of feedback.  It appears this was just copied from the linked issue, and it follows some patterns/practices that I don't thin are great.  In particular, it looks like it was written to run on ancient browsers that don't support `async/await` for some reason as well as being highly Chrome specific (possibly crashing on other browsers) and swallowing errors silently.

extension/app/ts/backgroundServiceWorker.ts
```diff
@@ -28,3 +28,36 @@ const injectContentScript = async () => {
 }
 
 injectContentScript()
+
+/// Workaround for ChromiumIssue1316588 https://bugs.chromium.org/p/chromium/issues/detail?id=1316588
+// TODO, remove when chrome fixes the bug
+const storageArea = chrome.storage.local as chrome.storage.LocalStorageArea & { onChanged: chrome.storage.StorageChangedEvent }
```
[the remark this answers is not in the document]
KillariDev: this is only run in manifest v3 where I already use chrome namespace as not all features are supported in `browser` namespace

MicahZoltu (approved):

PR #265, opened 2023-03-24
remove `globalThis.interceptor.signerName`

KillariDev:


extension/app/ts/background/settings.ts
```diff
@@ -187,3 +187,15 @@ export async function updateSimulationResults(simulationResults: SimulationResul
...
+
+export async function saveSignerName(signerName: SignerName) {
+	return await browser.storage.local.set({ signerName: signerName})
+}
+
+export async function getSignerName() {
+	const results = await browser.storage.local.get(['signerName'])
+	if (results) {
+		return SignerName.parse(results.signerName)
+	}
```
MicahZoltu, line 199: This reinforces my belief (see comment in other PR) that we should either fix the type signature of `browser.storage.local.get` or create a helper (like you did for `set`) that corrects the type signature.
```suggestion
	const results = await browser.storage.local.get(['signerName']) as unknown
	if (typeof results === 'object' && results && 'signerName' in results) {
		return SignerName.parse(results.signerName)
	}
```

extension/app/ts/background/settings.ts
```diff
@@ -187,3 +187,15 @@ export async function updateSimulationResults(simulationResults: SimulationResul
 		return await browser.storage.local.set({ simulationResults: SimulationResults.serialize(simulationResults) })
 	})
 }
+
+export async function saveSignerName(signerName: SignerName) {
+	return await browser.storage.local.set({ signerName: signerName})
+}
```
MicahZoltu, line 193: This should be updated to use the new `browserStorageLocalSet` function you added in the other PR.

MicahZoltu (changes requested):
I encourage addressing the root problem of `browser.storage.local.get` returning `any` which allows for this class of mistake.

extension/app/ts/utils/typescript.ts
```diff
@@ -53,3 +53,9 @@ export function assertIsObject(maybe: unknown): asserts maybe is Object {
 export async function browserStorageLocalGet(keys: string | string[]) {
 	return await browser.storage.local.get(keys) as Promise<Record<string, unknown>>
```
MicahZoltu, line 54:
```suggestion
	return await browser.storage.local.get(keys) as Promise<Record<string, JSONEncodeable>
```

MicahZoltu (approved):

PR #266, opened 2023-03-24
do not prompt access dialog for `eth_accounts`

KillariDev:


MicahZoltu (approved):

MicahZoltu:
The code will do what it says it will, but why would we want to return `[]` instead of `[account}` when the dapp asks for accounts?

KillariDev:
> The code will do what it says it will, but why would we want to return `[]` instead of `[account}` when the dapp asks for accounts?

Because here the user has not given access to the dapp. Returning the accounts would dox the users account without users permission (the `verifyAccess` has returned `askAccess` means that there's currently no access but it should be asked from the user). The standard is to return [] if there's no access given

PR #268, opened 2023-03-26
Fix creating tabs like Address Book on *nix systems

0xjimmy:
- Use absolute path for V2 urls


extension/app/ts/background/backgroundUtils.ts
```diff
@@ -20,7 +20,7 @@ export async function sendPopupMessageToOpenWindows(message: MessageToPopup) {
 			if (error?.message?.includes('A listener indicated an asynchronous response by returning true, but the message channel closed before a response was received')) {
 				return
 			}
-			if (error) return console.error(`Popup message error: ${ error.message }`);
+			if (error) return console.error(`Popup message error: ${error.message}`);
```
KillariDev, line 23: on this repo we always have spaces inside curly brackets

MicahZoltu: Much to Micah's annoyance.

PR #272, opened 2023-03-27
getCurrent can be undefined

KillariDev:


MicahZoltu:
Would be nice to get a PR into Definitely Typed to fix the type signature as well.  I think that is an ideal long-term fix, and it also helps everyone else in the world rather than just us.

PR #282, opened 2023-03-30
Fix release for two artifacts

epheph:


.github/workflows/release-extension.yml
```diff
@@ -14,21 +14,23 @@ jobs:
         id: derive_tag
         run: |
           VERSION_FILE=extension/app/ts/version.ts
-          MANIFEST_FILE=extension/app/manifest.json
+          MANIFEST_V2_FILE=extension/app/manifestV2.json
+          MANIFEST_V3_FILE=extension/app/manifestV3.json
```
KillariDev, line 18: on chrome the manifest file needs to be called "manifest.json", and calling it V2 or V3 doesn't work. In my pipline I rename v2 or v3 to manifest.json depending on which is being built

epheph: Right, this relies on setup to rename the file, and it modifies the file that gets copied in.

KillariDev (approved):

Issue #294, opened 2023-03-31
If we wanted, we could instead adjust the typing of `getStorageLocal` and `setStorageLocal` functions so these are the only valid keys.  That, along with fixing the type errors that stem from that initial change, would result in the type checker ensuring only these specific strings are used everywhere but without the need to create constants like this.  It would also give us strong guarantees (compiler validated) that these are the *only* keys we write to/read from storage and someone can't accidentally write some other key that isn't recorded in a central location.


KillariDev:
If we wanted, we could instead adjust the typing of `getStorageLocal` and `setStorageLocal` functions so these are the only valid keys.  That, along with fixing the type errors that stem from that initial change, would result in the type checker ensuring only these specific strings are used everywhere but without the need to create constants like this.  It would also give us strong guarantees (compiler validated) that these are the *only* keys we write to/read from storage and someone can't accidentally write some other key that isn't recorded in a central location.

_Originally posted by @MicahZoltu in https://github.com/DarkFlorist/TheInterceptor/pull/280#discussion_r1153327481_

PR #296, opened 2023-03-31
Storage keys typing

KillariDev:
Fix https://github.com/DarkFlorist/TheInterceptor/issues/294

### DIFF
--- a/extension/app/ts/utils/typescript.ts
+++ b/extension/app/ts/utils/typescript.ts
@@ -50,18 +50,39 @@ export function assertIsObject(maybe: unknown): asserts maybe is Object {
 	if (!isObject(maybe)) throw new Error(`Expected object but got ${ typeof maybe }`)
 }
 
-export async function browserStorageLocalGet(keys: string | string[]) {
-	return await browser.storage.local.get(keys) as Promise<Record<string, JSONEncodeable>>
+type StorageKey = 'activeSigningAddress'
+	| 'activeSimulationAddress'
+	| 'addressInfos'
+	| 'page'
+	| 'useSignersAddressAsActiveAddress'
+	| 'websiteAccess'
+	| 'activeChain'
+	| 'simulationMode'
+	| 'pendingAccessRequests'
+	| 'contacts'
+	| 'makeMeRich'
+	| 'addressbookTabId'
+	| 'ConfirmationWindowPromise'
+	| 'ChainChangeConfirmationPromise'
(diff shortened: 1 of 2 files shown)


extension/app/ts/utils/typescript.ts
```diff
@@ -50,18 +50,39 @@ export function assertIsObject(maybe: unknown): asserts maybe is Object {
...
+	| 'makeMeRich'
+	| 'addressbookTabId'
+	| 'ConfirmationWindowPromise'
+	| 'ChainChangeConfirmationPromise'
+	| 'PersonalSignPromise'
+	| 'InterceptorAccessRequestPromise'
+	| 'simulationResults'
+	| 'signerName'
+	| 'currentTabId'
+	| `tabState_${ number }`
```
MicahZoltu, line 72: What is this storing?  Feels like something where we could end up with unbounded keys stored that we cannot easily clear out?  Perhaps it would be better to store an object at `tabState` with `{ [key: number]: ... }` instead so we can easily iterate, delete the whole thing, etc.

MicahZoltu (approved):

extension/app/ts/utils/typescript.ts
KillariDev: It storing information about metamasks state and then what is the state of the extension icon.
```ts
export const TabState = funtypes.Object({
	signerName: SignerName,
	signerAccounts: funtypes.ReadonlyArray(EthereumAddress),
	signerChain: funtypes.Union(EthereumQuantity, funtypes.Undefined),
	tabIconDetails: TabIconDetails,
})
```

I was thinking of storing it to `tabState ` parameter, but then when retrieving for single tabstate, you would always need to retrieve them all, which feels wasteful.

I have function to clear them like this:
```ts
export async function clearTabStates() {
	const allStorage = Object.keys(await browser.storage.local.get())
	const keysToRemove = allStorage.filter((entry) => entry.match(/^tabState_[0-9]+/))
	await browser.storage.local.remove(keysToRemove)
}
```

MicahZoltu: Oh, `browser.storage.local.get` returns all keys?  If so then my original concern is mitigated.

KillariDev: yeah. I feel your suggestion would be better if you could retrieve individual keys from a hierarchy. I was thinking that, but there's no way to do that.

Issue #302, opened 2023-04-04
eth_subscriptions break if webworker goes down


KillariDev:

KillariDev:
instead of using `eth_subscribe`, use the block we are already querying anyway

PR #303, opened 2023-04-04
Refactor only: separate subscriptions from ethereumclient service

KillariDev:


extension/app/ts/simulation/services/ETHSubscriptionService.ts
```diff
@@ -0,0 +1,92 @@
...
+import { ErrorWithData } from '../../utils/errors.js'
+import { Future } from '../../utils/future.js'
+
+type Subscription = {
+	callback: (subscriptionId: string, reply: JsonRpcNewHeadsNotification) => void
+	params: EthSubscribeParams,
+	rpcSocket: WebSocket
+}
+
+export class ETHSubscriptionService {
```
MicahZoltu, line 11: To match naming convention elsewhere:
```suggestion
export class EthereumSubscriptionService {
```
Alternatively (not a fan of this but better than `ETH`):
```suggestion
export class EthSubscriptionService {
```

extension/app/ts/simulation/services/ETHSubscriptionService.ts
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	public readonly createSubscription = async (params: EthSubscribeParams, callback: (subscriptionId: string, reply: JsonRpcNewHeadsNotification) => void) => {
+		switch(params.params[0]) {
+			case 'newHeads': {
+				const rpcSocket = new WebSocket(this.webSocketConnectionString)
+				const subscriptionId = new Future<string>()
+
+				rpcSocket.addEventListener('open', _event => {
+					const request = { jsonrpc: '2.0', id: 0, method: 'eth_subscribe', params: ['newHeads'] }
```
MicahZoltu, line 37: Out of scope of this PR, but I recommend against using `eth_subscribe` and instead just polling every block.

MicahZoltu (approved):

extension/app/ts/simulation/services/ETHSubscriptionService.ts
KillariDev: the point was to refer to `eth_subscription` with the naming

extension/app/ts/simulation/services/ETHSubscriptionService.ts
KillariDev: yeah, there's PR for it: https://github.com/DarkFlorist/TheInterceptor/issues/302

extension/app/ts/simulation/services/ETHSubscriptionService.ts
MicahZoltu: I recommend not actually implementing this service using `eth_subscription` under the hood, hence I recommend not porting that name over to the class.  😀

PR #319, opened 2023-04-12
Fix gas estimation if we run out of gas, calculate nonce right on simgas and hardcode getcode gas

KillariDev:


extension/app/ts/simulation/services/SimulationModeEthereumClientService.ts
```diff
@@ -66,7 +66,7 @@ export const simulateEstimateGas = async (ethereumClientService: EthereumClientS
 
 // calculates gas price for receipts
 export const calculateGasPrice = (transaction: EthereumUnsignedTransaction, gasUsed: bigint, gasLimit: bigint, baseFeePerGas: bigint) => {
-	if( 'gasPrice' in transaction) {
+	if('gasPrice' in transaction) {
```
0xjimmy, line 69:
```suggestion
	if ('gasPrice' in transaction) {
```

0xjimmy (approved):

MicahZoltu (approved):

extension/app/ts/simulation/services/SimulationModeEthereumClientService.ts
```diff
@@ -326,17 +326,11 @@ export const getSimulatedCode = async (ethereumClientService: EthereumClientServ
...
 		maxFeePerGas: 0n,
 		maxPriorityFeePerGas: 0n,
-		gas: await simulateEstimateGas(ethereumClientService, simulationState, {
-			from: MOCK_ADDRESS,
-			to: GET_CODE_CONTRACT,
-			data: input,
-			value: 0n,
-			gasPrice: 0n,
-		}),
+		gas: 94104n,
```
MicahZoltu, line 333: Gas costs can change.  Can we perhaps memoize this value instead of hard coding it so we don't have something randomly break in a future hardfork?

KillariDev: you mean its called once and then stored? When do we know when we should update it again?

MicahZoltu: Correct.  Just once on startup is probably fine, if it breaks the user can just "turn it off and on again".

KillariDev: hmm, I would rather not add a lot stuff at the startup. Maybe it could be done first time its called and stored

Issue #317, opened 2023-04-11
Swaps with multiple paths are not  detected as swaps


0xjimmy:
Multi-hop swaps are fine, like `ETH -> USDC -> DAI` detects a ETH to DAI swap.
But Uniswap can also create multiple swap paths to which are not detected, for example `ETH -> USDC -> DAI` + `ETH -> DAI`.

KillariDev:
Hmm, they seem to be? Here's an example, do you have example case?
![image](https://user-images.githubusercontent.com/13102010/231100863-0962aa2d-2008-4447-8ffc-dddeea5b3abe.png)
![image](https://user-images.githubusercontent.com/13102010/231101053-1ceafe50-b970-4745-a051-be340aadb2a7.png)

KillariDev:
ah right. The issue arises when your account receives multiple token batches, these should be grouped together

PR #320, opened 2023-04-12
refactor, instead of using `is721` flag, use `type` flag instead. Add `BeforeAfterBalance` to swap identification

KillariDev:
this should fix https://github.com/DarkFlorist/TheInterceptor/issues/317 as well. But I wasn't able to generate such transaction with uniswap now to test it.

### DIFF
--- a/extension/app/ts/background/background.ts
+++ b/extension/app/ts/background/background.ts
@@ -33,51 +33,47 @@ let simulator: Simulator | undefined = undefined
 
 export async function updateSimulationState(getUpdatedSimulationState: () => Promise<SimulationState | undefined>, activeAddress: bigint | undefined) {
 	if (simulator === undefined) return
-	try {
-		const simId = (await getSimulationResults()).simulationId + 1
-		const updatedSimulationState = await getUpdatedSimulationState()
+	const simId = (await getSimulationResults()).simulationId + 1
+	const updatedSimulationState = await getUpdatedSimulationState()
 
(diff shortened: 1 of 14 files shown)


extension/app/ts/background/background.ts
```diff
@@ -247,6 +245,8 @@ async function handleSigningMode(
 	try {
 		parsedRequest = EthereumJsonRpcRequest.parse(request.options)
 	} catch (error) {
+		console.log(request)
+		console.log(error)
```
MicahZoltu, line 249:
```suggestion
		console.error(request)
		console.error(error)
```

extension/app/ts/background/background.ts
```diff
@@ -162,6 +158,8 @@ async function handleSimulationMode(
...
```
MicahZoltu, line 162:
```suggestion
		console.error(request)
		console.error(error)
```

MicahZoltu (approved):

extension/app/ts/background/background.ts
MicahZoltu: What do you mean?  I don't think an app has insight into console errors.  The only thing this should do is change the color and how it gets filtered when using dev tooling.

KillariDev: Chrome shows them as errors for the extension

KillariDev: image
this button appears

PR #331, opened 2023-04-13
Fix Error: No tab with id

KillariDev:
When querying for icon for tab that user has already closed results in an error, this will get rid of that error by making icon and title undefined for such page

### DIFF
--- a/extension/app/ts/background/iconHandler.ts
+++ b/extension/app/ts/background/iconHandler.ts
@@ -71,13 +71,10 @@ export async function retrieveWebsiteDetails(port: browser.runtime.Port, website
 		try {
 			return browser.tabs.get(tabId)
 		} catch (error) {
-			if (error instanceof Error) {
-				if (error.message?.includes('No tab with id')) {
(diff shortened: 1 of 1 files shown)


extension/app/ts/background/iconHandler.ts
```diff
@@ -67,6 +67,20 @@ export async function updateExtensionBadge() {
...
+		try {
+			return browser.tabs.get(tabId)
+		} catch (error) {
+			if (error instanceof Error) {
+				if (error.message?.includes('No tab with id')) {
+					// if tab is not found (user might have closed it)
+					return undefined
+				}
+			}
+			throw error
```
MicahZoltu, line 80: Consider inverting the conditional with guard clauses.
```suggestion
			if (!(error instanceof Error)) throw error
			if (!error.message?.includes('No tab with id')) throw error
			// if tab is not found (user might have closed it)
			return undefined
```

MicahZoltu (approved):

PR #332, opened 2023-04-13
Use to address in transfer visualisation

0xjimmy:
``simTx.transaction.to`` should already exist based on identifyTransaction checks. Check anyway and throw if not.
Remove ``to`` from  SimulatedAndVisualizedEtherTransferTransaction as not used anywhere.


extension/app/ts/components/simulationExplaining/customExplainers/SimpleSendVisualisations.tsx
```diff
@@ -90,8 +90,9 @@ function getBeforeAndAfterBalanceForAddress(ethBalances: readonly EthBalanceChan
 }
 
 export function EtherTransferVisualisation({ simTx, renameAddressCallBack }: { simTx: SimulatedAndVisualizedEtherTransferTransaction, renameAddressCallBack: RenameAddressCallBack }) {
+	if (typeof simTx.transaction.to === undefined) throw new Error("Not a simple transfer, missing to address")
```
KillariDev, line 93: I feel this should be fixed in the type information and not with runtime checks

KillariDev (commented):
I don't really see how this fixes the undefined bug, wouldn't this just throw?

0xjimmy:
Moved type casting to the identifyTransaction function.
```ts
	if (isEtherTransfer(simTx)) return {
		type: 'EtherTransfer',
		title: 'Ether Transfer',
		signingAction: 'Transfer Ether',
		simulationAction: 'Simulate Ether Transfer',
		rejectAction: 'Reject Ether Transfer',
		identifiedTransaction: { ...simTx, to: simTx.transaction.to as AddressBookEntry }
	}
```
`simTx.transaction.to` is `AddressBookEntry | undefined`, which we now check in ``isEtherTransfer``.

The `isEtherTransfer` now suggests that simTx.to exists, even though we only add it on the identify transaction. May be better to have the type just show that ``simTx.transaction.to`` is now only `AddressBookEntry`

extension/app/ts/components/simulationExplaining/identifyTransaction.tsx
```diff
@@ -140,7 +141,7 @@ export function identifyTransaction(simTx: SimulatedAndVisualizedTransaction): I
 		signingAction: 'Transfer Ether',
 		simulationAction: 'Simulate Ether Transfer',
 		rejectAction: 'Reject Ether Transfer',
-		identifiedTransaction: simTx,
+		identifiedTransaction: { ...simTx, to: simTx.transaction.to as AddressBookEntry }
```
KillariDev, line 144: we should not be doing casting here

0xjimmy: Removed, found that other tx types expect ``simTx.to`` to exist when the `to` is not undefined. ``simTx.to`` was never set anywhere during `formSimulatedAndVisualizedTransaction` so we now set it.

0xjimmy:
`to` is type `AddressBookEntry | undefined` so it is safe to cast after ruling out undefined. Adding the to should fix issues with visualizing token approvals and token transfers too. Since they expected a simTx.to where were previously never set simTx.to

```ts
...(to !== undefined ? { to: to as AddressBookEntry } : {}),
```

extension/app/ts/components/simulationExplaining/customExplainers/SimpleSendVisualisations.tsx
```diff
@@ -90,6 +90,7 @@ function getBeforeAndAfterBalanceForAddress(ethBalances: readonly EthBalanceChan
...
```
0xjimmy, line 93:
```suggestion
```
No longer needed

KillariDev:
> `to` is type `AddressBookEntry | undefined` so it is safe to cast after ruling out undefined. Adding the to should fix issues with visualizing token approvals and token transfers too. Since they expected a simTx.to where were previously never set simTx.to
> 
> ```ts
> ...(to !== undefined ? { to: to as AddressBookEntry } : {}),
> ```

if you rule it out, typescript would not demand you to cast it all

KillariDev (approved):

extension/app/ts/components/simulationExplaining/identifyTransaction.tsx
```diff
@@ -94,6 +94,7 @@ export type SimulatedAndVisualizedEtherTransferTransaction = SimulatedAndVisuali
 export function isEtherTransfer(simTx: SimulatedAndVisualizedTransaction): simTx is SimulatedAndVisualizedEtherTransferTransaction {
 	if (simTx.transaction.input.length == 0
```
jubalm, line 95: can use `===` for these

Issue #150, opened 2023-02-01
can use guard clause here?


KillariDev:
can use guard clause here? 
*edit: just realized the code i'm referring to is a couple steps below this one and is not shown on this chage set*

_Originally posted by @jubalm in https://github.com/DarkFlorist/TheInterceptor/pull/143#discussion_r1093061683_

PR #344, opened 2023-04-17
guard clause and styling, also remove [] from useEffect

KillariDev:
fix https://github.com/DarkFlorist/TheInterceptor/issues/150

### DIFF
--- a/extension/app/ts/components/pages/ChangeChain.tsx
+++ b/extension/app/ts/components/pages/ChangeChain.tsx
@@ -18,16 +18,16 @@ export function ChangeChain() {
 	const [chainChangeData, setChainChangeData] = useState<InterceptorChainChangeRequest | undefined>(undefined)
 	const [connectAnyway, setConnectAnyway] = useState<boolean>(false)
 
-	useEffect( () => {
+	useEffect(() => {
 		async function updatePage(message: ChangeChainRequest) {
-			setChainChangeData( {
+			setChainChangeData({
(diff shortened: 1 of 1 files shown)


0xjimmy (approved):

extension/app/ts/components/pages/ChangeChain.tsx
```diff
@@ -36,25 +36,25 @@ export function ChangeChain() {
 			await updatePage(message)
 		}
 		browser.runtime.onMessage.addListener(popupMessageListener)
-		sendPopupMessageToBackgroundPage( { method: 'popup_changeChainReadyAndListening' } )
+		sendPopupMessageToBackgroundPage({ method: 'popup_changeChainReadyAndListening' })
 		return () => browser.runtime.onMessage.removeListener(popupMessageListener)
-	}, [])
```
MicahZoltu, line 41: Will this result in this effect never running?  With `useSignalEffect` it will automatically detect any read signals and re-run the effect anytime any of those signals change.  Does it work the same with `useState` variables, where if they are set it will run again?

MicahZoltu (approved):

extension/app/ts/components/pages/ChangeChain.tsx
KillariDev: ,[] means run once, no param means run always, so it will always have the newest state available when updated. If you just run it once, the state it has available is the default initial state, which is not general what we want.

I really should look into signals 😬

Issue #323, opened 2023-04-12
Add calldata to confirm transaction view


KillariDev:
- show as highlight for transactions we dont understand
- show them in all transactions but don't highlight

PR #352, opened 2023-04-19
some extra details for transactions (such as input)

KillariDev:
fix https://github.com/DarkFlorist/TheInterceptor/issues/323
image


extension/app/ts/components/simulationExplaining/SimulationSummary.tsx
```diff
@@ -646,3 +646,66 @@ export function SimulationSummary(param: SimulationSummaryParams) {
...
+		const resizeTextArea = () => {
+			if (textareaRef && textareaRef.current) {
+				textareaRef.current.style.height = '0px'
+				const scrollHeight = textareaRef.current.scrollHeight
+				textareaRef.current.style.height = scrollHeight + 'px'
+			}
+		}
+		resizeTextArea()
+		globalThis.addEventListener('resize', resizeTextArea)
+		return () => globalThis.removeEventListener('resize', resizeTextArea)
```
MicahZoltu, line 667: This should really be done with CSS rather than JavaScript.  Not a blocker for this PR, but an issue should be created to fix it if this is merged as is.

MicahZoltu (approved):

extension/app/ts/components/simulationExplaining/SimulationSummary.tsx
KillariDev: hmm, I feel it cannot be done with CSS? but I could be wrong. Maybe I have been looking at too old CSS examples 🤔

PR #355, opened 2023-04-20
get rid of `@types/node-fetch` and `node-fetch` dependencies

KillariDev:
currently fails to ethers:
```
node_modules/ethers/types/providers/provider-ipcsocket.d.ts:1:23 - error TS2688: Cannot find type definition file for 'node'.

1 /// <reference types="node" resolution-mode="require"/>
                        ~~~~

node_modules/ethers/types/providers/provider-ipcsocket.d.ts:3:29 - error TS2307: Cannot find module 'net' or its corresponding type declarations.

3 import type { Socket } from "net";
```


KillariDev:
poblematic file: https://github.com/ethers-io/ethers.js/blob/main/types/providers/provider-ipcsocket.d.ts#L3

extension/app/ts/utils/typescript.ts
MicahZoltu, line 1: Should probably rename this file, or extract stuff out of it.  Original intent was to put general purpose types in here and typescript helpers like `assertNever`.  This seems to be expanding to now include general purpose utilities which feel like they should live in a different file, or at least in a better named file.

MicahZoltu (approved):
Big fan overall.

Issue #335, opened 2023-04-14
Figure out what to do when we cannot connect to the RPC provider


KillariDev:
![image](https://user-images.githubusercontent.com/13102010/231979877-c0ad7937-6c7a-405b-aad7-3cd75f701224.png)

PR #358, opened 2023-04-20
confirm transaction and simulation state connection errors

KillariDev:
Work towards https://github.com/DarkFlorist/TheInterceptor/issues/335
- Now if you lose connection when forming transaction you get an error and forming transaction is retryed when new block arrives
- Show error on simulation stack too if we cannot update it

still todo:
1) If dapp queries the extension we don't manage the error nicely
2) we should change our icon as not connected if no connection

### DIFF
--- a/extension/app/ts/background/background.ts
+++ b/extension/app/ts/background/background.ts
@@ -121,12 +121,9 @@ export async function refreshConfirmTransactionSimulation(
 			data: { ...info, ...await visualizeSimulatorState(simulationStateWithNewTransaction, simulator) }
 		}
 	} catch(error) {
-		if (error instanceof Error) {
-			if (isFailedToFetchError(error)) {
-				return { method: 'popup_confirm_transaction_simulation_failed' as const, data: info }
-			}
-		}
-		throw error
+		if (!(error instanceof Error)) throw error
+		if (!isFailedToFetchError(error)) throw error
+		return { method: 'popup_confirm_transaction_simulation_failed' as const, data: info }
 	}
 }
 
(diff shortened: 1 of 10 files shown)


extension/app/ts/background/background.ts
```diff
@@ -94,20 +104,29 @@ export async function refreshConfirmTransactionSimulation(
...
+			method: 'popup_confirm_transaction_simulation_state_changed' as const,
+			data: { ...info, ...await visualizeSimulatorState(simulationStateWithNewTransaction, simulator) }
 		}
+	} catch(error) {
+		if (error instanceof Error) {
+			if (isFailedToFetchError(error)) {
+				return { method: 'popup_confirm_transaction_simulation_failed' as const, data: info }
+			}
+		}
+		throw error
```
MicahZoltu, line 129: Consider inverting the conditional here with guard clauses?
```suggestion
		if (!(error instanceof Error)) throw error
		if (!isFailedToFetchError(error)) throw error
		return { method: 'popup_confirm_transaction_simulation_failed' as const, data: info }
```

extension/app/ts/background/popupMessageHandlers.ts
```diff
@@ -282,6 +282,17 @@ export async function homeOpened(simulator: Simulator) {
...
 
+	let blockNumber = undefined
+	try {
+		blockNumber = await simulator.ethereum.getBlockNumber()
+	} catch (error) {
+		if (error instanceof Error) {
+			if (isFailedToFetchError(error)) await sendPopupMessageToOpenWindows({ method: 'popup_failed_to_get_block' })
+		} else {
+			throw error
+		}
```
MicahZoltu, line 293:
```suggestion
		if (!(error instanceof Error)) throw error
		if (!isFailedToFetchError(error)) throw error
		await sendPopupMessageToOpenWindows({ method: 'popup_failed_to_get_block' })
```

MicahZoltu (approved):

PR #360, opened 2023-04-21
Adds not connected icon badge

KillariDev:
adds not connected icon
image


0xjimmy (approved):
The changes themselves look fine, but I didn't manage to see the badge when using the extension. What conditions do I need to see it on the extension?

KillariDev:
> The changes themselves look fine, but I didn't manage to see the badge when using the extension. What conditions do I need to see it on the extension?

cut off internet connection and it should happen when it cannot reach the RPC anymore

extension/app/ts/background/iconHandler.ts
```diff
@@ -62,7 +62,12 @@ export async function updateExtensionIcon(websiteTabConnections: WebsiteTabConne
 }
 
 export async function updateExtensionBadge() {
+	if ((await getIsConnected())?.isConnected === false) {
+		await setExtensionBadgeBackgroundColor({ color: '#f6685e' })
+		return await setExtensionBadgeText({ text: '!' })
```
MicahZoltu, line 67: Consider a yellow warning icon instead of a red exclamation point.  I feel like we should reserve red exclamation points for "you are being attacked!"
```suggestion
		await setExtensionBadgeBackgroundColor({ color: 'yellow' })
		return await setExtensionBadgeText({ text: '⚠️' })
```

MicahZoltu (approved):

jubalm:
Tested in Chrome build fine 👍

jubalm (approved):

extension/app/ts/background/iconHandler.ts
KillariDev: image

that does not really work, but I agree with the yellow color:
image

I made the dialog box yellow too:
image

MicahZoltu: Can we do transparent color and set the text to `⚠️`?  I feel like that would look quite good.

Issue #354, opened 2023-04-20
do text box resizing with CSS instead with javascript


KillariDev:
This should really be done with CSS rather than JavaScript.  Not a blocker for this PR, but an issue should be created to fix it if this is merged as is.

_Originally posted by @MicahZoltu in https://github.com/DarkFlorist/TheInterceptor/pull/352#discussion_r1171332872_

PR #364, opened 2023-04-24
get rid of textarea

KillariDev:
fix https://github.com/DarkFlorist/TheInterceptor/issues/361
fix https://github.com/DarkFlorist/TheInterceptor/issues/354
image

### DIFF
--- a/extension/app/ts/components/pages/PersonalSign.tsx
+++ b/extension/app/ts/components/pages/PersonalSign.tsx
@@ -1,4 +1,4 @@
-import { useState, useEffect, useRef } from 'preact/hooks'
+import { useState, useEffect } from 'preact/hooks'
 import { bigintToRoundedPrettyDecimalString, stringToUint8Array } from '../../utils/bigint.js'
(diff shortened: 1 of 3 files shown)


0xjimmy (approved):

extension/app/ts/components/pages/PersonalSign.tsx
```diff
@@ -178,8 +169,8 @@ export function PersonalSign() {
 								</p>
 							</header>
 							<div class = 'card-content'>
-								<div class = 'control'>
-									<textarea class = 'textarea' readonly ref = { textareaRef } style = 'overflow: hidden; resize: none;'>{ signRequest.message }</textarea>
+								<div class = 'textbox'>
+									<p class = 'paragraph' style = 'color: var(--subtitle-text-color)'>{ signRequest.message }</p>
```
MicahZoltu, line 173: Do we render this at all, or is it just a raw byte array?  If it is rendered text, we shouldn't use `break-all`.

Issue #147, opened 2023-02-01
Interceptor requesting access to page and then to address right after each other


KillariDev:
hmm I figured out when interceptor sometimes first asks access to the interceptor and then to your account.

Uniswap sometimes asks first eth_chainId which pops up the access dialog, and then it requests accounts right after

Make it so that the old access request is overwritten (disable the ok button when this happen and then reload)

PR #366, opened 2023-04-24
Update pending access right request if new one comes

KillariDev:
- now if app first asks to access to interceptor and then to address, we update the access request dialog and disable the buttons for 3 secs
Fix https://github.com/DarkFlorist/TheInterceptor/issues/147

### DIFF
--- a/extension/app/ts/components/pages/InterceptorAccess.tsx
+++ b/extension/app/ts/components/pages/InterceptorAccess.tsx
@@ -116,17 +116,24 @@ interface InterceptorAccessRequest {
 	socket: WebsiteSocket
 }
 
+const DISABLED_DELAY_MS = 3000
+
 export function InterceptorAccess() {
 	const [accessRequest, setAccessRequest] = useState<InterceptorAccessRequest | undefined>(undefined)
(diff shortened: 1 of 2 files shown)


extension/app/ts/components/pages/InterceptorAccess.tsx
```diff
@@ -116,17 +116,24 @@ interface InterceptorAccessRequest {
 	socket: WebsiteSocket
 }
 
+const DISABLED_DELAY_MS = 3000
```
MicahZoltu, line 119: I started to write a comment about how we should do this, but I'm happy to see you already thought of it!

extension/app/ts/components/pages/InterceptorAccess.tsx
```diff
@@ -197,6 +204,13 @@ export function InterceptorAccess() {
 		} } )
 	}
 
+	const informationChangedRecently = () => new Date().getTime() < informationUpdatedTimestamp + DISABLED_DELAY_MS
+
+	useEffect(() => {
+		const id = setInterval(() => setTimeSinceInformationUpdate((old) => old + 1), 1000)
+		return () => clearInterval(id)
+	}, [])
```
MicahZoltu, line 212: If you used signals you could avoid the `useEffect` here.  😀

MicahZoltu (approved):

PR #367, opened 2023-04-24
Cannot catch error without await here it seems?

KillariDev:


0xjimmy (approved):
Yes, otherwise you are directly returning a promise that hasn't resolved/rejected yet.

PR #369, opened 2023-04-26
Node20

KillariDev:
make project work with node 20 too


MicahZoltu:
Build error:
```
> prepare-inpage
> tsc --project ./tsconfig.json && node --enable-source-maps ../app/inpage/js/prepare.mjs && node --input-type=module -e "import { promises as fs } from 'fs'; await fs.rm('../app/inpage/js/prepare.mjs')"

[Error: ENOENT: no such file or directory, open '/extension/app/inpage/js/inpage.js'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/extension/app/inpage/js/inpage.js'
}
The command '/bin/sh -c npm run setup-firefox' returned a non-zero code: 1
```

First thing to try to fix it is bump the NodeJS version in the Dockerfile to 18 (it is currently node:16.16.0-alpine3.16).

You can do this by changing this line
https://github.com/DarkFlorist/TheInterceptor/blob/6874ceb6cf38b293d2a23c5399364c5fa75f4cac/extension/Dockerfile#L2
to
```Dockerfile
FROM node:18.16.0-alpine3.17@sha256:44aaf1ccc80eaed6572a0f2ef7d6b5a2982d54481e4255480041ac92221e2f11
```
You can also delete the comment at the top of the file.

See if that is sufficient for the following to work locally, and then on the build server:
```
cd extension
docker image build --tag temp .
docker image rm temp
```

MicahZoltu:
Ping me again once build is passing.

MicahZoltu (approved):

extension/tsconfig-inpage-create-injection-script.json
```diff
@@ -6,8 +6,8 @@
 		"outDir": "app/inpage/js/",
 		"rootDir": "app/inpage/ts/",
 		"noEmit": false,
-		"sourceMap": true,
-		"inlineSources": true,
+		"sourceMap": false,
+		"inlineSources": false,
```
MicahZoltu, line 10: While we do delete this file after creating, having the source maps available makes troubleshooting significantly easier in the case where you need to debug a failure.  My guess is that the added cost of generating the source map is less than the cost of having a poor debugging experience in the few cases where you actually need to debug.

Of course, you could just turn source maps back on in those cases...

PR #370, opened 2023-04-27
Fix race condition, close popup listener is removed before event

### LINKED ISSUE #347: Switching Networks breaks after first popup to switch.
When trying to switch networks multiple times the rpc returns 'User has declined request' and we are not prompted with a popup to accept/decline network change.
After reloading the extension the popup works for the first request only.

0xjimmy:
Fixes #347 
When `onCloseWindow` is called, it checks the closed window in event is the same  window.id and then sets `openedWindow = null`

On the finally block during cleanup, the onCloseWindow listener is removed before the popup event comes in.


KillariDev (approved):
good catch!

PR #375, opened 2023-04-28
Improved message signing

KillariDev:
- make permit 1/2 appear as approval messages
- if permit messages are malformed (wrong account, or wrong chain), show error

image

### DIFF
--- a/extension/app/ts/background/windows/personalSign.ts
+++ b/extension/app/ts/background/windows/personalSign.ts
@@ -1,3 +1,5 @@
+import { QUARANTINE_CODE } from '../../simulation/protectors/quarantine-codes.js'
+import { Simulator } from '../../simulation/simulator.js'
 import { METAMASK_ERROR_USER_REJECTED_REQUEST } from '../../utils/constants.js'
 import { Future } from '../../utils/future.js'
 import { HandleSimulationModeReturnValue, InterceptedRequest, PersonalSign, ExternalPopupMessage, Settings, UserAddressBook, SignerName, PersonalSignRequest } from '../../utils/interceptor-messages.js'
@@ -5,8 +7,8 @@ import { Website, WebsiteSocket, WebsiteTabConnections } from '../../utils/user-
 import { EIP2612Message, Permit2, PersonalSignParams, SignTypedDataParams } from '../../utils/wire-types.js'
 import { personalSignWithSimulator, sendMessageToContentScript } from '../background.js'
 import { getHtmlFile, sendPopupMessageToOpenWindows } from '../backgroundUtils.js'
-import { getAddressMetaData } from '../metadataUtils.js'
-import { getPendingPersonalSignPromise, setPendingPersonalSignPromise } from '../settings.js'
+import { getAddressMetaData, getTokenMetadata } from '../metadataUtils.js'
+import { getPendingPersonalSignPromise, getSettings, getSignerName, setPendingPersonalSignPromise } from '../settings.js'
 
 let pendingPersonalSign: Future<PersonalSign> | undefined = undefined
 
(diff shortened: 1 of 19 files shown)


MicahZoltu:
"Signature expires in" and "Spender can spend for" fields seem like they are the same thing, but in your screenshot they have very different values.  I feel like if I can't figure out what the difference is then users definitely won't be able to and we should try to improve the wording.

What does each mean?

extension/app/ts/background/windows/personalSign.ts
```diff
@@ -43,85 +46,124 @@ function reject() {
...
+		} as const
+	}
+	const namedParams = { param: originalParams.params[1], account: originalParams.params[0] }
+	const account = getAddressMetaData(namedParams.account, userAddressBook)
+
+	if (namedParams.param.primaryType === 'Permit') {
+		const parsed = EIP2612Message.parse(namedParams.param)
+		const token = await getTokenMetadata(ethereumClientService, parsed.domain.verifyingContract)
+		const owner = getAddressMetaData(parsed.message.owner, userAddressBook)
+		if (token.type === 'NFT') throw 'Attempted to perform Permit to an NFT'
```
MicahZoltu, line 108: Does permit not work with NFTs?  Is there some other standard for permitting NFTs?

extension/app/ts/simulation/protectors/quarantine-codes.ts
```diff
@@ -32,6 +32,14 @@ export const QUARANTINE_CODES_DICT = {
 	'FAILED_CHECK': {
 		label: 'A check failed',
 		longExplanation: `A check failed to complete. Most likely because the transaction stack is too big.`
+	},
+	'SIGNATURE_CHAIN_ID_DOES_NOT_MATCH': {
+		label: 'Chain mismatch',
+		longExplanation: `The signature request is for different chain than what is the active address.`
+	},
```
MicahZoltu, line 39: Is this actually an error?  Seems warning worthy at least, but I'm not sure it is error worthy?  One could imagine an L2 driven entirely by signatures so you never actually need to switch chains.

MicahZoltu (approved):

extension/app/ts/background/windows/personalSign.ts
KillariDev: permit and permit2 are only for erc20: https://eips.ethereum.org/EIPS/eip-2612. I guess this is the same for NFT's but its a draft: https://eips.ethereum.org/EIPS/eip-4494

KillariDev: I added issue for eip4494, I didn't directly find a dapp that would use it tho

extension/app/ts/simulation/protectors/quarantine-codes.ts
KillariDev: yeah, one could imagine that, and maybe we should change it then. Maybe if see such use cases we could change

KillariDev:
> "Signature expires in" and "Spender can spend for" fields seem like they are the same thing, but in your screenshot they have very different values. I feel like if I can't figure out what the difference is then users definitely won't be able to and we should try to improve the wording.
> 
> What does each mean?

Signature expiration refers to time when you need to sign the message, if you sign it too late, it won't be valid. I feel this is done in signer level only, so that a signer refuses to sign a message that is too old. When the message is signed "Spender can spend for" tells how long the spender can spend the funds until the signature is invalid

MicahZoltu:
Any idea why there is a time limit on how long you have to sign the message?  The signer can ignore this, and a malicious actor can put whatever they want there so it is unclear to me what problem is being solved.  Maybe interceptor should just ignore that field entirely so it isn't as confusing to users?

PR #378, opened 2023-04-28
When log does not match what we expect, ethers can throw instead of returning null

KillariDev:
ethers throw:
`RangeError: data out-of-bounds (buffer=0x, length=0, offset=32, code=BUFFER_OVERRUN, version=6.1.0)`
when `data` is empty

### DIFF
--- a/extension/app/ts/simulation/services/SimulationModeEthereumClientService.ts
+++ b/extension/app/ts/simulation/services/SimulationModeEthereumClientService.ts
@@ -581,6 +581,14 @@ const getSimulatedTokenBalances = async (ethereumClientService: EthereumClientSe
 	}))
 }
 
+const parseLogIfPossible = (ethersInterface: ethers.Interface, log: { topics: string[], data: string}) => {
+	try {
+		return ethersInterface.parseLog(log)
+	} catch (error) {
+		return null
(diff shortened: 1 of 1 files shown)


extension/app/ts/simulation/services/SimulationModeEthereumClientService.ts
```diff
@@ -581,6 +581,14 @@ const getSimulatedTokenBalances = async (ethereumClientService: EthereumClientSe
 	}))
 }
 
+const parseLogIfPossible = (ethersInterface: ethers.Interface, log: { topics: string[], data: string}) => {
+	try {
+		return ethersInterface.parseLog(log)
```
MicahZoltu, line 586: Ethers doesn't have a `tryParse` log feature?  Might be worth filing a feature request issue on the repository or submitting a PR for it.  This seems like a generally useful thing.

MicahZoltu (approved):

extension/app/ts/simulation/services/SimulationModeEthereumClientService.ts
KillariDev: I asked ricmoo about it in discord

PR #382, opened 2023-05-02
Get rid of /extension/ folder

KillariDev:
fix https://github.com/DarkFlorist/TheInterceptor/issues/159


Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+RUN npm run test
+
+WORKDIR /app
+RUN zip ../interceptor-firefox.zip -r .
+
+WORKDIR /
+RUN npm run setup-chrome
+
+WORKDIR /app
+RUN zip ../interceptor-chrome.zip -r .
```
MicahZoltu, line 39: I'm assuming this was a file move plus adjusting all of the paths?  No code changes otherwise?

test/tests/signTypedData.ts
```diff
@@ -1,4 +1,5 @@
-import { EIP712Message, Permit2 } from '../../app/ts/utils/wire-types.js'
+import { EIP712Message } from '../../app/ts/utils/wire-types.js'
+import { Permit2 } from '../../app/ts/utils/personal-message-definitions.js'
```
MicahZoltu, line 2: How did this not fail CI previously?  Sounds like a bug in our CI process?

MicahZoltu (approved):

Dockerfile
KillariDev: yea, I removed `extension/` from multiple places

test/tests/signTypedData.ts
KillariDev: It did fail the CI. I forgot to wait for CI to complete before merging (as CI is not run to branches not merging to main). I wonder if we could get it to run for those too?

KillariDev: I tried to create release and then I noticed it was failing for this: <https://github.com/DarkFlorist/TheInterceptor/actions/runs/4858826351>

Issue #136, opened 2023-01-27
Add support for opensea signatures


KillariDev:
https://docs.opensea.io/v2.0/reference/seaport-overview

PR #386, opened 2023-05-03
Opensea orders

KillariDev:
fix https://github.com/DarkFlorist/TheInterceptor/issues/136
image


MicahZoltu (approved):
Did not do a thorough review, so it would be good to get more eyes on this as it is pretty big.  This PR doesn't have any obviously malicious code in it though.  😬

app/ts/components/pages/PersonalSign.tsx
```diff
@@ -155,17 +162,48 @@ function SignRequest({ personalSignRequestData, renameAddressCallBack }: SignReq
...
+	return <>
+		<CellElement text = 'Chain: '/>
+	<	CellElement text = { getChainName(BigInt(permit2.message.domain.chainId)) }/>
+		<CellElement text = 'Nonce: '/>
+		<CellElement text = { permit2.message.message.details.nonce.toString(10) }/>
+		<CellElement text = 'Signature expires  in:'/>
+		<CellElement text = { <SomeTimeAgo priorTimestamp = { new Date(Number(permit2.message.message.sigDeadline) * 1000) } countBackwards = { true }/> }/>
+		<CellElement text = 'Spender can spend for:'/>
+		<CellElement text = { <>
+			<SomeTimeAgo priorTimestamp = { new Date(Number(permit2.message.message.details.expiration) * 1000) } countBackwards = { true }/>
```
jubalm, line 179: new Date instances could be called before the return call so it's only instantiated once.

app/ts/components/pages/PersonalSign.tsx
```diff
@@ -155,17 +162,48 @@ function SignRequest({ personalSignRequestData, renameAddressCallBack }: SignReq
...
+			{` (until ${ new Date(Number(permit2.message.message.details.expiration) * 1000).toISOString().split('T')[0] })`}
```
jubalm, line 180: date localization maybe? `new Date().toLocaleDateString("en-US")`

app/ts/components/simulationExplaining/customExplainers/OpenSeaOrder.tsx
```diff
@@ -0,0 +1,126 @@
...
+			<p class = 'paragraph' style = 'color: var(--subtitle-text-color); text-overflow: ellipsis; overflow: hidden;'>{ param.text }</p>
+		</div>
+	}
+	return <>
+		<CellElement text = 'Conduit key: '/>
+		<CellElement text = { bytes32String(orderComponents.conduitKey) }/>
+		<CellElement text = 'Counter: '/>
+		<CellElement text = { orderComponents.counter }/>
+		<CellElement text = 'Start time: '/>
+		<CellElement text = { new Date(Number(orderComponents.startTime) * 1000).toISOString().split('T')[0] }/>
```
jubalm, line 110: can use a date format utility if used often

jubalm (approved):

app/ts/components/pages/PersonalSign.tsx
MicahZoltu: I have a pretty strong preference for ISO8601 dates rather than localized dates (because most date localizations are incredibly stupid and the world needs to move towards a global standard).  The US localization in particular is literally the worst of them all.  😆

app/ts/components/pages/PersonalSign.tsx
KillariDev: hmm what do you mean by this?

MicahZoltu: I believe the idea is to just do `const priorTimestamp = new Date(...)` up above somewhere, then your elements would be something like `<SomeTimeAgo priorTimestamp countBackwards = { true }/>`

Issue #389, opened 2023-05-04
malformed signature request will still open a dialog button


KillariDev:
- to test, make malformed `SafeTx` request for example and see a dialog still popping, we should refuse this before window opens

PR #392, opened 2023-05-05
fix EIP712 parsing

KillariDev:
Now instead of failing to parse EIP712 message if it doesn't fit to one one ouf identified types, show it as non-identified

Fix https://github.com/DarkFlorist/TheInterceptor/issues/389

### DIFF
--- a/app/ts/background/windows/personalSign.ts
+++ b/app/ts/background/windows/personalSign.ts
@@ -4,7 +4,8 @@ import { stringifyJSONWithBigInts } from '../../utils/bigint.js'
 import { METAMASK_ERROR_USER_REJECTED_REQUEST } from '../../utils/constants.js'
 import { Future } from '../../utils/future.js'
 import { HandleSimulationModeReturnValue, InterceptedRequest, PersonalSign, ExternalPopupMessage, Settings, UserAddressBook, PersonalSignRequest } from '../../utils/interceptor-messages.js'
-import { EIP2612Message, OpenSeaOrder, OpenSeaOrderMessage, Permit2, SafeTx } from '../../utils/personal-message-definitions.js'
+import { OpenSeaOrderMessage, PersonalSignRequestIdentifiedEIP712Message } from '../../utils/personal-message-definitions.js'
+import { assertNever } from '../../utils/typescript.js'
 import { AddressBookEntry, SignerName, Website, WebsiteSocket, WebsiteTabConnections } from '../../utils/user-interface-types.js'
 import { OldSignTypedDataParams, PersonalSignParams, SignTypedDataParams } from '../../utils/wire-types.js'
 import { personalSignWithSimulator, sendMessageToContentScript } from '../background.js'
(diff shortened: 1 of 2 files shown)


app/ts/background/windows/personalSign.ts
```diff
@@ -112,54 +113,86 @@ export async function craftPersonalSignPopupMessage(ethereumClientService: Ether
...
-
-	if (namedParams.param.primaryType === 'Permit') {
-		const parsed = EIP2612Message.parse(namedParams.param)
-		const token = await getTokenMetadata(ethereumClientService, parsed.domain.verifyingContract)
-		const owner = getAddressMetaData(parsed.message.owner, userAddressBook)
-		if (token.type === 'NFT') throw 'Attempted to perform Permit to an NFT'
+	
+	let parsed 
+	try {
+		parsed = PersonalSignRequestIdentifiedEIP712Message.parse(namedParams.param)
```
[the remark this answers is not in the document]
KillariDev: yeah, but then I would be catching a lot possible other errors that I do not want to be catching

app/ts/background/windows/personalSign.ts
```diff
@@ -112,54 +113,86 @@ export async function craftPersonalSignPopupMessage(ethereumClientService: Ether
...
-					owner,
-					spender: getAddressMetaData(parsed.message.spender, userAddressBook),
-					verifyingContract: token,
-				},
-				...await getQuarrantineCodes(BigInt(parsed.domain.chainId), account, activeAddressWithMetadata, owner),
+				quarantine: false,
+				quarantineCodes: []
 			}
 		} as const
 	}
```
MicahZoltu, line 134: Don't use try/catch blocks for flow control.  `safeParse` exists for exactly this scenario, where you aren't confident that parsing will succeed and you need to branch on success/failure.  If necessary (doesn't look like it is here), you can access the error details in `maybeParsed.message`, `maybeParsed.fullError`, and `maybeParsed.key` inside of the failure code path.
```suggestion
	const maybeParsed = PersonalSignRequestIdentifiedEIP712Message.safeParse(namedParams.param)
	if (!maybeParsed.success) {
		// if we fail to parse the message, that means it's a message type we do not identify, let's just show it as a nonidentified EIP712 message
		return {
			method: 'popup_personal_sign_request',
			data: {
				originalParams,
				...basicParams,
				type: 'EIP712' as const,
				message: namedParams.param,
				account,
				quarantine: false,
				quarantineCodes: []
			}
		} as const
	}
	const parsed = maybeParsed.value
```

app/ts/background/windows/personalSign.ts
```diff
@@ -112,54 +113,86 @@ export async function craftPersonalSignPopupMessage(ethereumClientService: Ether
...
 	const namedParams = { param: originalParams.params[1], account: originalParams.params[0] }
 	const account = getAddressMetaData(namedParams.account, userAddressBook)
-
-	if (namedParams.param.primaryType === 'Permit') {
-		const parsed = EIP2612Message.parse(namedParams.param)
-		const token = await getTokenMetadata(ethereumClientService, parsed.domain.verifyingContract)
-		const owner = getAddressMetaData(parsed.message.owner, userAddressBook)
-		if (token.type === 'NFT') throw 'Attempted to perform Permit to an NFT'
+
```
MicahZoltu, line 116: Nitpick.  I thought our .editorconfig dealt with this automatically?
```suggestion

```

MicahZoltu (approved):
Approved, but I *strongly* recommend applying the change I suggested or something similar.  Flow control via exception handling is generally a bad idea.

app/ts/background/windows/personalSign.ts
KillariDev: hmm I think it should

PR #395, opened 2023-05-05
Show raw message along with decoded one

KillariDev:
image


MicahZoltu (approved):
Consider making this a toggle, so the user can switch between Raw and UTF-8 Decoded, rather than displaying them both at the same time (consumes double real-estate when only one is actually useful).

We could pretty easily "guess" the right one to show as default by just attempting UTF-8 decoding and then checking to see if there are any control characters in the text or if it is not a valid UTF-8 string (e.g., it ends with one byte of a two-byte character).

jubalm (approved):

KillariDev:
yeah, well with these messages, there's nothing else to show, but yeah, I guess switch makes sense, like how etherscan does

Issue #404, opened 2023-05-08
add special handing (and parsing) for special domain eip-712 fields


KillariDev:
string name the user readable name of signing domain, i.e. the name of the DApp or the protocol.
string version the current major version of the signing domain. Signatures from different versions are not compatible.
uint256 chainId the [EIP-155](https://eips.ethereum.org/EIPS/eip-155) chain id. The user-agent should refuse signing if it does not match the currently active chain.
address verifyingContract the address of the contract that will verify the signature. The user-agent may do contract specific phishing prevention.
bytes32 salt an disambiguating salt for the protocol. This can be used as a domain separator of last resort.

PR #405, opened 2023-05-08
Parse and validate EIP-712 messages and show metadata on them

KillariDev:
- parses message field: checks that the type information is available and then add address metadata to it.
- Visualize arbitrary EIP-712 messages
- This only applies o EIP-712 messages that we do not directly recognize
- Does not include domain fields parsing, I added issue about it (https://github.com/DarkFlorist/TheInterceptor/issues/404)

image
image

### DIFF
--- a/app/ts/utils/eip712Parsing.ts
+++ b/app/ts/utils/eip712Parsing.ts
@@ -1,0 +1,340 @@
+import { UserAddressBook } from "./interceptor-messages.js"
+import { assertNever } from "./typescript.js"
+import { AddressBookEntry } from "./user-interface-types.js"
+import { EIP712Message, EthereumAddress, EthereumData, EthereumQuantity, JSONEncodeableObject, NonHexBigInt } from "./wire-types.js"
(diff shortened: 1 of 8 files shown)


app/ts/utils/eip712Parsing.ts
```diff
@@ -0,0 +1,339 @@
...
+	funtypes.Literal('bytes28'),
+	funtypes.Literal('bytes29'),
+	funtypes.Literal('bytes30'),
+	funtypes.Literal('bytes31'),
+	funtypes.Literal('bytes32'),
+	funtypes.Literal('bool'),
+	funtypes.Literal('address'),
+	funtypes.Literal('string'),
+	funtypes.Literal('bytes'),
+)
```
MicahZoltu, line 110: Putting the "fun" in funtypes.

MicahZoltu (approved):

app/ts/utils/eip712Parsing.ts
KillariDev: Literally

PR #412, opened 2023-05-10
fix simulation mode and signing mode mixup

KillariDev:
Fix https://github.com/DarkFlorist/TheInterceptor/issues/409


MicahZoltu (approved):
I'm still concerned that we have had a number of bugs of this class (data mixups) and it feels like we should consider some architectural changes to make it harder to get such mixups.

KillariDev:
> I'm still concerned that we have had a number of bugs of this class (data mixups) and it feels like we should consider some architectural changes to make it harder to get such mixups.

yeah, I haven't figured out what is the best way to handle these cases, as the user can flip between the signing and simulation states quite easily while using the same dapp

PR #414, opened 2023-05-10
No signer connection UI update

KillariDev:
fixes https://github.com/DarkFlorist/TheInterceptor/issues/400

Makes UI more clear about informing that there's no wallet connection

image

image

### DIFF
--- a/app/ts/components/pages/Home.tsx
+++ b/app/ts/components/pages/Home.tsx
@@ -31,29 +31,24 @@ type SignerExplanationParams = {
 function SignerExplanation(param: SignerExplanationParams) {
 	if (param.activeAddress !== undefined) return <></>
 	if (param.tabIcon === ICON_NOT_ACTIVE) {
-		return <div class = 'content'>
-			<p class = 'paragraph' style = 'color: var(--subtitle-text-color)'>(The page you are looking at has not yet connected to the wallet, or you do not have a browser wallet installed)</p>
-		</div>
+		return <>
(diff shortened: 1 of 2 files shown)


jubalm (approved):
Maybe the disabled button "change" isn't necessary anymore as there's already a message about changing address in MetaMask.

Also the message "...or you do not have a browser wallet installed" shows even while trying to connect. could be dynamic while awaiting

KillariDev:
I feel the disabled button is good to attract the viewers attention as the button is disabled and then the user would read the texts why. I added issue about the other line, I agree we shouldn't think there's no wallet installed if we have information that there's some wallet.

app/ts/components/pages/Home.tsx
```diff
@@ -83,58 +78,77 @@ function FirstCard(param: FirstCardParams) {
...
+	function connectToSigner() {
+		sendPopupMessageToBackgroundPage({ method: 'popup_requestAccountsFromSigner', options: true })
+	}
+
+	if (param.signerName === 'NoSigner' && param.simulationMode === false) {
+		return <>
+			<div class = 'card' style = 'margin: 10px;'>
+				<FirstCardHeader { ...param }/>
+				<div class = 'card-content'>
+					<DinoSays text = { 'No signer connnected. You can use Interceptor in simulation mode without a signer, but signing mode requires a browser wallet.' } />
```
MicahZoltu, line 95: Text can be written in HTML
```suggestiotn
<DinoSays text = 'No signer connnected. You can use Interceptor in simulation mode without a signer, but signing mode requires a browser wallet.' />
```

PR #416, opened 2023-05-10
handle `No tab with id` error here in case of race condition

KillariDev:


app/ts/background/popupMessageHandlers.ts
```diff
@@ -270,7 +270,14 @@ export async function openAddressBook() {
 	const addressBookTab = allTabs.find((tab) => tab.id === tabId)
 
 	if (addressBookTab?.id === undefined) return await openInNewTab()
-	return await browser.tabs.update(addressBookTab.id, { active: true })
+	try {
+		return await browser.tabs.update(addressBookTab.id, { active: true })
+	} catch (error) {
+		if (!(error instanceof Error)) throw error
+		if (!error.message?.includes('No tab with id')) throw error
```
jubalm, line 277: could be a constant

jubalm (approved):

app/ts/background/popupMessageHandlers.ts
KillariDev: hmm what could be constant?

KillariDev: ah, I got it. Instead of a random string, we make it a constant and use the same constant everywhere. I agree, I'll change that

Issue #371, opened 2023-04-27
Add option to queue transactions


KillariDev:
- when dapp sends transaction, allow them to send another that is queud
- add queue view to confirm transaction
- when confirm transaction view is closed, all transactions queued are canceled
- option to reject current transaction and all transactions
- when new transaction comes, instead of blocking it, add it to queue

PR #441, opened 2023-05-17
Confirm transactions window allows pending transactions instead of rejecting transactions if the window is already open

KillariDev:
- fixes: https://github.com/DarkFlorist/TheInterceptor/issues/371
- I moved the website that created the transaction to the transaction card as each transaction can have a different website
- the contract being called was moved to bottom of the transaction
- Added shadows to card to separate them from each other more clearly'
- When new transaction is added to the queue, the confirm transaction view is focused. (this is not yet very clear that the transaction was added to the queue and the transaction user sees is not the one they were prompted with. There's issue for this: https://github.com/DarkFlorist/TheInterceptor/issues/436 )

image


app/ts/components/pages/ConfirmTransaction.tsx
```diff
@@ -1,24 +1,49 @@
...
+import { QuarantineCodes, SenderReceiver, TransactionImportanceBlock } from '../simulationExplaining/Transactions.js'
 import { identifyTransaction } from '../simulationExplaining/identifyTransaction.js'
 import { SomeTimeAgo } from '../subcomponents/SomeTimeAgo.js'
 import { TIME_BETWEEN_BLOCKS } from '../../utils/constants.js'
 
+type UnderTransactionsParams = {
+	pendingTransactions: ConfirmTransactionTransactionSingleVisualizationArray
+}
+
+const HALF_HEADER_HEIGHT = 48 / 2
```
MicahZoltu, line 21: It feels like this should use a variable defined elsewhere rather than `48`.  That way if we change the header, this adjusts correctly.

MicahZoltu (approved):

PR #444, opened 2023-05-17
Transaction created timestamp

KillariDev:
fix https://github.com/DarkFlorist/TheInterceptor/issues/435

image


KillariDev:
I am not super fan of putting it in that location, but I couldn't come up with better place for it 🤔 . I also did not want to highlight it that much, as I feel its not super important field, but nice to know sometimes

MicahZoltu (approved):
I like the chosen position.

PR #446, opened 2023-05-18
fix gas limit and expand senderreceiver

KillariDev:
fix https://github.com/DarkFlorist/TheInterceptor/issues/437
image

### DIFF
--- a/app/ts/background/simulationModeHanders.ts
+++ b/app/ts/background/simulationModeHanders.ts
@@ -55,7 +55,7 @@ export async function sendTransaction(
 	website: Website,
 	settings: Settings,
 ) {
-	async function formTransaction() {
+	const formTransaction = async() => {
 		const simulationState = (await getSimulationResults()).simulationState
 		if (simulationState === undefined) return undefined
 		const block = getSimulatedBlock(ethereumClientService, simulationState)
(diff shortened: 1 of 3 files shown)


app/ts/background/simulationModeHanders.ts
```diff
@@ -65,19 +65,23 @@ export async function sendTransaction(
...
-				nonce: await transactionCount,
-				maxFeePerGas: sendTransactionParams.params[0].maxFeePerGas ? sendTransactionParams.params[0].maxFeePerGas : maxFeePerGas,
-				maxPriorityFeePerGas: sendTransactionParams.params[0].maxPriorityFeePerGas ? sendTransactionParams.params[0].maxPriorityFeePerGas : 1n,
-				gas: sendTransactionParams.params[0].gas ? sendTransactionParams.params[0].gas : 90000n,
-				to: sendTransactionParams.params[0].to === undefined ? null : sendTransactionParams.params[0].to,
-				value: sendTransactionParams.params[0].value ? sendTransactionParams.params[0].value : 0n,
-				input: 'data' in sendTransactionParams.params[0] && sendTransactionParams.params[0].data !== undefined ? sendTransactionParams.params[0].data : new Uint8Array(),
-				accessList: []
+				...transactionWithoutGas,
+				...(transactionDetails.gas !== undefined ? { gas: transactionDetails.gas } : { gas: await simulateEstimateGas(ethereumClientService, simulationState, transactionWithoutGas) })
```
MicahZoltu, line 84: Can simplify this:
```suggestion
				gas: transactionDetails.gas || await simulateEstimateGas(ethereumClientService, simulationState, transactionWithoutGas)
```

Note: this will result in a gas estimation if `transactionDetails.gas === 0n`, but that is probably the right thing to do in that scenario anyway.

MicahZoltu (approved):

app/ts/background/simulationModeHanders.ts
KillariDev: I feel that's not the right thing to do? I feel it should just fail

PR #447, opened 2023-05-18
tighten blockie

KillariDev:
fix https://github.com/DarkFlorist/TheInterceptor/issues/433

### DIFF
--- a/app/ts/components/subcomponents/PreactBlocky.tsx
+++ b/app/ts/components/subcomponents/PreactBlocky.tsx
@@ -1,15 +1,13 @@
 import { useEffect, useState } from 'preact/hooks'
 import { DataURLCache } from './DataURLCache.js'
 import { JSX } from 'preact/jsx-runtime'
+import { addressString } from '../../utils/bigint.js'
 const dataURLCache = new DataURLCache()
 
 interface BlockieProps {
-	seed: string,
-	size?: number,
+	address: bigint,
 	scale?: number,
 	color?: string,
-	bgColor?: string,
-	spotColor?: string,
 	borderRadius?: string,
 	style?: JSX.CSSProperties
 }
@@ -111,16 +109,15 @@ function generateIdenticon(options: BlockieProps, canvasRef: HTMLCanvasElement) 
 	}
 
 	const opts = options || {}
-	const size = opts.size || 8
 	const scale = opts.scale || 4
-	const seed = opts.seed || Math.floor((Math.random()*Math.pow(10,16))).toString(16)
+	const seed = addressString(opts.address)
 
 	seedrand(seed)
 
 	const color = opts.color || createColor()
-	const bgcolor = opts.bgColor || createColor()
-	const spotcolor = opts.spotColor || createColor()
-	const imageData = createImageData(size)
+	const bgcolor = createColor()
+	const spotcolor = createColor()
+	const imageData = createImageData(8)
 	const canvas = setCanvas(canvasRef, imageData, color, scale, bgcolor, spotcolor)
 
 	return canvas
@@ -128,23 +125,23 @@ function generateIdenticon(options: BlockieProps, canvasRef: HTMLCanvasElement) 
 
 export default function Blockie(props: BlockieProps) {
 	const scale = props.scale || 4
-	const dimension = (props.size || 8) * scale
-	const [seed, setSeed] = useState<string | undefined>(props.seed)
-	const [dataURL, setDataURL] = useState<string | undefined>(dataURLCache.get(`${ props.seed }!${ dimension }`))
+	const dimension = 8 * scale
+	const [address, setAddress] = useState<bigint | undefined>(props.address)
+	const [dataURL, setDataURL] = useState<string | undefined>(dataURLCache.get(`${ props.address }!${ dimension }`))
 
 	useEffect(() => {
-		if (dataURL === undefined || seed !== props.seed) {
-			setSeed(props.seed)
+		if (dataURL === undefined || address !== props.address) {
+			setAddress(props.address)
 			const element = document.createElement('canvas')
 			generateIdenticon(props, element)
 			element.toBlob((blob) => {
 				if (!blob) return
 				const dataUrl = URL.createObjectURL(blob)
 				setDataURL(dataUrl)
-				dataURLCache.set(dataUrl, `${ props.seed }!${ dimension }`)
+				dataURLCache.set(dataUrl, `${ props.address }!${ dimension }`)
 			})
 		}
-	}, [props.seed])
+	}, [props.address])
 	return <img
 		src = { dataURL === undefined ? 'data:image/gif;base64,R0lGODlhAQABAAAAACwAAAAAAQABAAA=' : dataURL}
 		style = {
--- a/app/ts/components/subcomponents/PreactBlocky.tsx
+++ b/app/ts/components/subcomponents/PreactBlocky.tsx
@@ -1,18 +1,18 @@
-import { useEffect, useState } from 'preact/hooks'
 import { DataURLCache } from './DataURLCache.js'
 import { JSX } from 'preact/jsx-runtime'
 import { addressString } from '../../utils/bigint.js'
+import { useAsyncState } from '../../utils/preact-utilities.js'
+import { Signal, useSignalEffect } from '@preact/signals'
+import { Future } from '../../utils/future.js'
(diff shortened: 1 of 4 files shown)


app/ts/components/subcomponents/PreactBlocky.tsx
```diff
@@ -111,40 +109,39 @@ function generateIdenticon(options: BlockieProps, canvasRef: HTMLCanvasElement)
...
 
 	const opts = options || {}
-	const size = opts.size || 8
 	const scale = opts.scale || 4
-	const seed = opts.seed || Math.floor((Math.random()*Math.pow(10,16))).toString(16)
+	const seed = addressString(opts.address)
 
 	seedrand(seed)
 
 	const color = opts.color || createColor()
```
MicahZoltu, line 117: What does this do if it doesn't affect the output blockie?

app/ts/components/subcomponents/PreactBlocky.tsx
```diff
@@ -111,40 +109,39 @@ function generateIdenticon(options: BlockieProps, canvasRef: HTMLCanvasElement)
...
+		if (dataURL === undefined || address !== props.address) {
+			setAddress(props.address)
 			const element = document.createElement('canvas')
 			generateIdenticon(props, element)
 			element.toBlob((blob) => {
 				if (!blob) return
 				const dataUrl = URL.createObjectURL(blob)
 				setDataURL(dataUrl)
-				dataURLCache.set(dataUrl, `${ props.seed }!${ dimension }`)
+				dataURLCache.set(dataUrl, `${ props.address }!${ dimension }`)
```
MicahZoltu, line 141: This is going to be something like `1234n!32`, that won't be a problem (bigint native string serialization vs hex serialization)?

app/ts/components/subcomponents/PreactBlocky.tsx
```diff
@@ -1,15 +1,13 @@
 import { useEffect, useState } from 'preact/hooks'
 import { DataURLCache } from './DataURLCache.js'
 import { JSX } from 'preact/jsx-runtime'
+import { addressString } from '../../utils/bigint.js'
 const dataURLCache = new DataURLCache()
 
 interface BlockieProps {
-	seed: string,
-	size?: number,
+	address: bigint,
```
MicahZoltu, line 8: Consider a signal here since you have to touch all of the call sites anyway.
```suggestion
	address: Signal<bigint>,
```
Same for all of the properties, but this one is the most likely to change.

app/ts/components/subcomponents/PreactBlocky.tsx
```diff
@@ -111,40 +109,39 @@ function generateIdenticon(options: BlockieProps, canvasRef: HTMLCanvasElement)
...
 	return canvas
 }
 
 export default function Blockie(props: BlockieProps) {
 	const scale = props.scale || 4
-	const dimension = (props.size || 8) * scale
-	const [seed, setSeed] = useState<string | undefined>(props.seed)
-	const [dataURL, setDataURL] = useState<string | undefined>(dataURLCache.get(`${ props.seed }!${ dimension }`))
+	const dimension = 8 * scale
+	const [address, setAddress] = useState<bigint | undefined>(props.address)
```
MicahZoltu, line 129: Why do you store the address (which is already coming from props) in local state?  This will make it so a change of address on the `<Blockie .../>` will not correctly propagate to update the blockie.

app/ts/components/subcomponents/PreactBlocky.tsx
```diff
@@ -111,40 +109,39 @@ function generateIdenticon(options: BlockieProps, canvasRef: HTMLCanvasElement)
...
+	const [dataURL, setDataURL] = useState<string | undefined>(dataURLCache.get(`${ props.address }!${ dimension }`))
 
 	useEffect(() => {
-		if (dataURL === undefined || seed !== props.seed) {
-			setSeed(props.seed)
+		if (dataURL === undefined || address !== props.address) {
+			setAddress(props.address)
 			const element = document.createElement('canvas')
```
MicahZoltu, line 135: I believe this element needs to be removed after we are done generating a blob out of it?  Otherwise we will get an unbounded number of these elements over time.

app/ts/components/subcomponents/PreactBlocky.tsx
```diff
@@ -111,40 +109,39 @@ function generateIdenticon(options: BlockieProps, canvasRef: HTMLCanvasElement)
...
 			element.toBlob((blob) => {
 				if (!blob) return
 				const dataUrl = URL.createObjectURL(blob)
 				setDataURL(dataUrl)
-				dataURLCache.set(dataUrl, `${ props.seed }!${ dimension }`)
+				dataURLCache.set(dataUrl, `${ props.address }!${ dimension }`)
 			})
 		}
-	}, [props.seed])
+	}, [props.address])
```
MicahZoltu, line 144: If you switch to signals for the properties, you can "simplify" this to:
```suggestion
	const dimension = useComputed(() => 8 * (props.scale.value || 4))
	const dataUrl = useComputed(() => {
		const url = `${props.address.value}!${dimension.value}`
		const cachedUrl = dataURLCache.get(url)
		if (cachedUrl !== undefined) return cacheUrl
		setAddress(props.address)
		const element = document.createElement('canvas')
		generateIdenticon(props, element)
		element.toBlob((blob) => {
			if (!blob) return
			const dataUrl = URL.createObjectURL(blob)
			setDataURL(dataUrl)
			dataURLCache.set(dataUrl, url)
		})
		document.removeChild(element)
		// TODO: figure out if this can ever return undefined after the above code
		return dataURLCache.get(url)
	})
```

MicahZoltu (approved):

app/ts/components/subcomponents/PreactBlocky.tsx
KillariDev: no problem there, its just to invalidate the cache if the address changes, it doesn't really matter what's the format

app/ts/components/subcomponents/PreactBlocky.tsx
KillariDev: hmm What's the question? Previously if the seed was missing, it generated a random seed

app/ts/components/subcomponents/PreactBlocky.tsx
KillariDev: this is only called when the element is created I think?

app/ts/components/subcomponents/PreactBlocky.tsx
MicahZoltu: If the user overrides the color, what happens?  The color of each pixel in the blockie needs to match exactly what is on Etherscan, MetaMask, etc. so I don't understand what setting this could do that wouldn't affect the validity of the blockie.

app/ts/components/subcomponents/PreactBlocky.tsx
MicahZoltu: This example component has its address changed every second(ish), but the Blockie won't update with the code as you have it now I don't think.
```tsx
function MyComponent() {
	const address = useSignal(0n)
	useEffect(() => {
		const id = setTimeout(() => address.value = Date.now % 2 ? 1n : 0n = , 1000)
		return () => clearTimeout(id)
	})
	return <Blockie address={address.value}/>
}
```

app/ts/components/subcomponents/PreactBlocky.tsx
KillariDev: oh, you are refering the `opts.color`? you are right it should be removed

app/ts/components/subcomponents/PreactBlocky.tsx
```diff
@@ -110,51 +108,51 @@ function generateIdenticon(options: BlockieProps, canvasRef: HTMLCanvasElement)
...
+		future.resolve(dataUrl)
+	})
+	return await future
+}
+
+export function Blockie(props: BlockieProps) {
+	const dimension = 8 * (props.scale?.value || 4)
+    const { value: dataURL, waitFor } = useAsyncState<string>()
+    useSignalEffect(() => {
+		props.address.value
```
MicahZoltu, line 144: This shouldn't be necessary, I suggested adding it just for testing but IIRC it didn't fix anything.  As long as this function touches `props.address.value` before it goes async (which it appears to do) then the signal effect will work.

PR #448, opened 2023-05-18
eth_sendRawTransaction

KillariDev:
fix https://github.com/DarkFlorist/TheInterceptor/issues/432


app/ts/background/simulationModeHanders.ts
```diff
@@ -98,6 +99,36 @@ export async function sendTransaction(
...
+			data: stringToUint8Array(transaction.data),
+			...transaction.gasLimit === null ? { gas: transaction.gasLimit } : {},
+			value: transaction.value,
+			...transaction.to === null ? {} : { to: EthereumAddress.parse(transaction.to) },
+			...transaction.gasPrice === null ? {} : { gasPrice: transaction.gasPrice },
+			...transaction.maxPriorityFeePerGas === null ? {} : { maxPriorityFeePerGas: transaction.maxPriorityFeePerGas },
+			...transaction.maxFeePerGas === null ? {} : { maxFeePerGas: transaction.maxFeePerGas },
+		}]
+	}
+	return sendTransaction(websiteTabConnections, getActiveAddressForDomain, ethereumClientService, sendTransactionParams, socket, request, simulationMode, website, settings)
```
MicahZoltu, line 129: I'm not sure this is what we want to do.  If the dapp gives us a raw transaction, we shouldn't be calling `eth_sendTransaction` on the connected wallet, we should be calling either `eth_sendRawTransaction` against our RPC or we should be calling `eth_sendRawTransaction` on MetaMask (whichever is appropriate).

In this situation, we don't actually need a connected wallet, so we could do this even in simulation mode.

MicahZoltu (changes requested):
I *think* the behavior proposed here is incorrect, but I also may be misunderstanding something.

app/ts/background/simulationModeHanders.ts
KillariDev: We are converting `eth_sendRawTransaction` to `eth_sendTransaction` here. Is there a reason why we wouldn't want to be doing that?

MicahZoltu: It is very possible (and arguably to-spec) that wallets will ignore any signature properties included in an `eth_sendTransaction` call, which would result in the user getting prompted to sign the transaction in their wallet.  The transaction being submitted is likely not `from` the wallet account, so this likely would result in stuff breaking if we implement it this way.

KillariDev: yeah, I am prompting user to sign the transaction here. Right, that's incorrect. I feel we should still present the transaction to user and the user will then decide to forward or not forward it

MicahZoltu: I think it is reasonable to show the transaction to the user before broadcasting it, we just need to broadcast it as a signed transaction (`eth_sendRawTransaction`) not as an unsigned transaction (`eth_sendTransaction`).

app/ts/background/background.ts
```diff
@@ -301,8 +304,14 @@ async function handleSigningMode(
 		case 'eth_signTypedData_v3':
 		case 'eth_signTypedData_v4': return await personalSign(ethereumClientService, websiteTabConnections, socket, parsedRequest, request, false, website, settings)
 		case 'wallet_switchEthereumChain': return await switchEthereumChain(websiteTabConnections, socket, ethereumClientService, parsedRequest, request, false, website)
+		case 'eth_sendRawTransaction': {
+			if (isSupportedChain(settings.activeChain.toString()) ) {
+				return sendRawTransaction(ethereumClientService, parsedRequest, socket, request, false, website, settings)
+			}
```
MicahZoltu, line 310: Should we be checking to see if the raw transaction's chainId matches the active chain's?

app/ts/background/simulationModeHanders.ts
```diff
@@ -95,6 +96,68 @@ export async function sendTransaction(
...
+		const transaction = ethers.Transaction.from(dataStringWith0xStart(sendRawTransactionParams.params[0]))
+		const transactionDetails = {
+			from: EthereumAddress.parse(transaction.from),
+			data: stringToUint8Array(transaction.data),
+			...transaction.gasLimit === null ? { gas: transaction.gasLimit } : {},
+			value: transaction.value,
+			...transaction.to === null ? {} : { to: EthereumAddress.parse(transaction.to) },
+			...transaction.gasPrice === null ? {} : { gasPrice: transaction.gasPrice },
+			...transaction.maxPriorityFeePerGas === null ? {} : { maxPriorityFeePerGas: transaction.maxPriorityFeePerGas },
+			...transaction.maxFeePerGas === null ? {} : { maxFeePerGas: transaction.maxFeePerGas },
```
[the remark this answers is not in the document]
MicahZoltu: Seems fine enough to me.

MicahZoltu (approved):

Issue #372, opened 2023-04-27
Add tab view for confirm transaction


KillariDev:
- make it optional setting

PR #451, opened 2023-05-19
Tab optionality

KillariDev:
Fix https://github.com/DarkFlorist/TheInterceptor/issues/372

- adds option to open all popups in a tab instead
- when the tab or popup is closed with reject/approval, activate the tab that made the request
- this PR does not add possibility to enable tab mode in the UI. You can enable it in console with: `await browser.storage.local.set({ useTabsInsteadOfPopup: true })`. I'll add settings menu in a future PR.


MicahZoltu (approved):
🥳

PR #461, opened 2023-05-23
if Gas estimation fails, throw an error to website instead of using max gas

KillariDev:
Fixes https://github.com/DarkFlorist/TheInterceptor/issues/457 . But instead of fixing the issue by adding +20% gas, we handle the real issue, as we already add 20%


MicahZoltu (approved):
Will the same change need to be made to the code added in eth_sendRawTransaction?  I thought I saw a new gas estimate happening in there.

PR #465, opened 2023-05-24
do not update `eth_sendRawTransaction`'s nonces if its invalid

KillariDev:


MicahZoltu (approved):
I normally like small PRs, but it feels like this PR should be included with #464 as #464 introduces a bug without this.

KillariDev:
hmm yeah. But logically they are separate functionalities, so I feel its easier to audit them like this

Issue #467, opened 2023-05-25
ERC721 Transfer gets displayed as ERC20 Transfer From


0xjimmy:
ERC721 Transfer gets displayed as ERC20 Transfer From
Would have to assume it displays as 'ERC20 Transfer From' because both ERC20 and ERC721 use functions with the same `transferFrom(address,address,uint256)` signature.

![image](https://github.com/DarkFlorist/TheInterceptor/assets/52896585/7fe4e7b2-9f0a-4a81-b507-8d49e5e35a0a)

PR #471, opened 2023-05-25
add ERC721 to fourbyte explanation for Transfer Form

KillariDev:
Fix https://github.com/DarkFlorist/TheInterceptor/issues/467

Other option is just call it "Transfer From" but I am not sure if "ERC20/ERC721 Transfer From" is more clear

### DIFF
--- a/app/ts/utils/constants.ts
+++ b/app/ts/utils/constants.ts
@@ -84,7 +84,7 @@ function get4Byte(functionAbi: string) {
 }
 
 export const FourByteExplanations = new Map<number, string >([
-	[get4Byte('transferFrom(address,address,uint256)'), 'ERC20 Transfer From'],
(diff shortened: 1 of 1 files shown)


MicahZoltu (approved):

MicahZoltu:
Lots of stuff may use `transferFrom`, so I recommend just calling it `Transfer From`.  We could take it a step further and turn it into a complete sentence like:
```
Transfer X TOK from Y to Z
```

KillariDev:
> Lots of stuff may use `transferFrom`, so I recommend just calling it `Transfer From`. We could take it a step further and turn it into a complete sentence like:
> 
> ```
> Transfer X TOK from Y to Z
> ```

This requires to be able to simulate the transactions thought, with 4byte we can get some name for failing transactions too. And jimmys transaction is not pure transfer function as there's approve there, thats why it doesn't get more accurate title

PR #477, opened 2023-05-26
add some missing fields to raw transaction details

KillariDev:
fix https://github.com/DarkFlorist/TheInterceptor/issues/438
image

### DIFF
--- a/app/ts/components/simulationExplaining/SimulationSummary.tsx
+++ b/app/ts/components/simulationExplaining/SimulationSummary.tsx
@@ -6,12 +6,12 @@ import { ERC721Token, Ether, EtherAmount, EtherSymbol, Token, TokenAmount, Token
 import { LogAnalysis } from './Transactions.js'
 import { CopyToClipboard } from '../subcomponents/CopyToClipboard.js'
 import { SomeTimeAgo, humanReadableDateDeltaLessDetailed } from '../subcomponents/SomeTimeAgo.js'
-import { CHAINS, MAKE_YOU_RICH_TRANSACTION } from '../../utils/constants.js'
-import { addressString, dataStringWith0xStart, nanoString } from '../../utils/bigint.js'
+import { CHAINS, MAKE_YOU_RICH_TRANSACTION, getChainName } from '../../utils/constants.js'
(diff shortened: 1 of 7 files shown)


app/ts/components/simulationExplaining/SimulationSummary.tsx
```diff
@@ -688,13 +686,26 @@ export function ExtraDetailsTransactionCard({ transaction }: ExtraDetailsTransac
...
+						<CellElement text = 'Gas used: '/>
+						<CellElement text = { `${ gasSpent.toString(10) } gas (${ Number(gasSpent * 10000n / transaction.gas) / 100 }%)` }/>
 						<CellElement text = 'Gas limit: '/>
 						<CellElement text = { `${ transaction.gas.toString(10) } gas` }/>
+						<CellElement text = 'Nonce: '/>
+						<CellElement text = { transaction.nonce.toString(10) }/>
+						<CellElement text = 'Chain: '/>
+						<CellElement text = { getChainName(BigInt(transaction.chainId)) }/>
+						<CellElement text = 'Unsigned transaction hash: '/>
+						<CellElement text = { bytes32String(transaction.hash) }/>
```
MicahZoltu, line 707: That transaction hash is taking up a lot of space and isn't necessary for anything.  It can be calculated from the rest of the transaction details and I don't think any tools use it for anything.

Perhaps keep it, but greatly constrain how much space it takes up and let people click to copy it if they want the whole thing, or hover tooltip for whole thing?

MicahZoltu (approved):

app/ts/components/simulationExplaining/SimulationSummary.tsx
KillariDev: I imagine that if you want to sign that transaction offchain, that's the only information you need? Sure Interceptor could lie about it

KillariDev: well there's space for it, and if there's not it gets ellipsed

MicahZoltu: The problem is that it causes the rest of the fields to be separated from their labels by a lot.  What if we right justify the left column and left justify the right column?  How does that look?

Issue #284, opened 2023-03-30
unify style of multiline ternaries


KillariDev:
Hmm, I'm seeing this style of ternary in a number of places now.  I thought the style for multiline ternaries we were using in this project was this?
```ts
condition
	? firstBranch
	: secondBranch
```

_Originally posted by @MicahZoltu in https://github.com/DarkFlorist/TheInterceptor/pull/273#discussion_r1151890576_

PR #479, opened 2023-05-26
align multiline trinary code style

KillariDev:
fix https://github.com/DarkFlorist/TheInterceptor/issues/284

### DIFF
--- a/extension/.vscode/settings.json
+++ b/extension/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "typescript.tsdk": "node_modules\\typescript\\lib"
-}
--- a/app/ts/AddressBook.tsx
+++ b/app/ts/AddressBook.tsx
@@ -99,8 +99,9 @@ export function ListElement(entry: ListElementParam) {
 				<div class = 'media' style = 'height: 100%;'>
 					<div class = 'media-content' style = 'overflow-y: visible; overflow-x: unset; height: 100%;'>
 						<div style = 'padding-bottom: 10px; height: 40px'>
-							{ entry.type === 'empty' ? <></> :
-								<BigAddress
+							{ entry.type === 'empty'
+								? <></>
+								: <BigAddress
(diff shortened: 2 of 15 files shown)


extension/.vscode/settings.json
```diff
@@ -1,3 +0,0 @@
-{
-    "typescript.tsdk": "node_modules\\typescript\\lib"
-}
```
MicahZoltu, line 3: I don't think you want to remove this...

app/ts/utils/json-rpc-requests.ts
MicahZoltu, line 1: Is this supposed to be an empty file?

MicahZoltu (changes requested):
Last two files seem out of place.  A mistake?

extension/.vscode/settings.json
KillariDev: ah commited that by mistake

app/ts/utils/json-rpc-requests.ts
KillariDev: yeah, did not intend to commit it yet

extension/.vscode/settings.json
KillariDev: actually that file should no longer be needed

MicahZoltu: Why not?

MicahZoltu: Hmm, your assumption is that VSCode will "do the right thing" in that case?  Happen to have any docs that support that, because I feel like in my experience VSCode still used its built-in even when I had a `node_modules` in the root.

KillariDev: well I don't think that file did anything as it was alone in `extension/` folder. Not sure why vscode would look it from there?

MicahZoltu: Oh, I see the problem now!  Sorry, I thought this was `<root>/.vscode/settings.json`

MicahZoltu (approved):

Issue #480, opened 2023-05-30
Bug: Tab mode sometimes results in inability to request_accounts.


MicahZoltu:
When I have the setting enabled to open prompts in a new tab, an account request pops up even though I already have the account approved for the site, and approving in that dialog sends a rejection to the dapp.

KillariDev:
should be fixed as the permissions was rewamped

PR #491, opened 2023-06-06
Get rid of notification center and replace it with support for multiple access requests

KillariDev:
fix https://github.com/DarkFlorist/TheInterceptor/issues/490

image


MicahZoltu (approved):

app/ts/background/windows/interceptorAccess.ts
```diff
@@ -114,212 +99,178 @@ export async function requestAccessFromUser(
...
-	if (pendingInterceptorAccess !== undefined) return rejectReply()
-
+	let justAddToPending = false
+	if (pendingInterceptorAccess.size !== 0) justAddToPending = true
+	
 	// check if we need to ask address access or not. If address is put to never need to have address specific permision, we don't need to ask for it
 	const askForAddressAccess = requestAccessToAddress !== undefined && settings.userAddressBook.addressInfos.find((x) => x.address === requestAccessToAddress.address)?.askForAddressAccess !== false
 	const accessAddress = askForAddressAccess ? requestAccessToAddress : undefined
+	const future = new Future<InterceptorAccessReply>()
+	const requestId = request === undefined ? -Math.random() : request.requestId // if there's no particular request requesting this access, generate random ID for it
```
MicahZoltu, line 111: Why not just a file-local constant that increments over time?  Does this need to be unique across runs or something?

MicahZoltu (approved):
This is a pretty big change, would be good to get more eyes on it.

app/ts/background/windows/interceptorAccess.ts
KillariDev: I'll realized it has to be consistent between sessions too. I'll keep it as random

KillariDev:
I refactored pr quite a bit and this should also be solved https://github.com/DarkFlorist/TheInterceptor/issues/480

app/ts/background/windows/interceptorAccess.ts
KillariDev: I also realized I can generate that Id deterministically :)

app/ts/components/pages/InterceptorAccess.tsx
```diff
@@ -1,20 +1,58 @@
...
+		}) }
+	</div>
+}
+
 function AssociatedTogether({ associatedAddresses, renameAddressCallBack }: { associatedAddresses: readonly AddressInfoEntry[], renameAddressCallBack: RenameAddressCallBack } ) {
 	const [showLogs, setShowLogs] = useState<boolean>(associatedAddresses.length > 1)
 
 	return <>
-		<div class = 'card' style = 'margin-top: 10px; margin-bottom: 10px'>
+		<div class = 'card' style = 'margin-top: 10px; margin-bottom: 10px;'>
```
jubalm, line 55: there are `<p>` tags inside `<p>` elements within this code block which could potentially display incorrectly. consider using `<span>` to be safe

app/ts/components/pages/InterceptorAccess.tsx
```diff
@@ -1,20 +1,58 @@
...
+	</header>
+}
+
+type UnderTransactionsParams = {
+	reversedPendingAccessRequestArray: PendingAccessRequestArray
+}
+
+function UnderAccesses(param: UnderTransactionsParams) {
+	const nTx = param.reversedPendingAccessRequestArray.length
+	return <div style = {`position: relative; top: ${ nTx * -HALF_HEADER_HEIGHT }px;`}>
```
jubalm, line 40: consider using objects as style properties here
```suggestion
	return <div style = {{ position: 'relative', top: nTx * -HALF_HEADER_HEIGHT }}>
```

jubalm (commented):
Consider using object notation for inline styles. As a bonus, TypeScript can help ensure spelling and add unit `px` to values.

```ts
// string
<p style = `fon-size: 13px; margin-bottom: ${SOMEREFERENCE / 2}px`>

// object notation
<p style = {{ fontSize: 13, marginBottom: SOMEREFERENCE / 2 }}>
```

app/ts/components/pages/InterceptorAccess.tsx
KillariDev: I created issue about this. I didn't know that its not allowed

app/ts/components/pages/InterceptorAccess.tsx
KillariDev: added issue about this, I have been planning to move into this styling but I havent gotten aroudn it

PR #520, opened 2023-06-13
Import export settings

KillariDev:
fix: https://github.com/DarkFlorist/TheInterceptor/issues/207

image

### DIFF
--- a/app/css/interceptor.css
+++ b/app/css/interceptor.css
@@ -707,7 +707,3 @@ svg.spinner > circle {
 	padding-bottom: 10px;
 	padding-top: 10px;
 }
-
-input[type = 'file'] {
-	opacity: 0    
-}
--- a/app/css/interceptor.css
+++ b/app/css/interceptor.css
@@ -707,3 +707,7 @@ svg.spinner > circle {
 	padding-bottom: 10px;
 	padding-top: 10px;
 }
+
+input[type = 'file'] {
+	opacity: 0    
+}
--- a/app/ts/background/popupMessageHandlers.ts
+++ b/app/ts/background/popupMessageHandlers.ts
@@ -342,7 +342,7 @@ export async function importSettings(settingsData: ImportSettings) {
 	if (!parsed.success) {
 		return await sendPopupMessageToOpenWindows({
 			method: 'popup_initiate_export_settings_reply',
-			data: { success: false, errorMessage: 'Failed to read the file. Is it a valid interceptor settings file?' }
+			data: { success: false, errorMessage: 'Failed to read the file. It is not a valid interceptor settings file' }
 		})
 	}
 	await importSettingsAndAddressBook(parsed.value)
@@ -354,7 +354,6 @@ export async function importSettings(settingsData: ImportSettings) {
 
 export async function exportSettings() {
 	const exportedSettings = await exportSettingsAndAddressBook()
-	//const blobData = new Blob([JSON.stringify(ExportedSettings.serialize(exportedSettings))], { type: 'text/json; charset=utf-8' })
 	await sendPopupMessageToOpenWindows({
 		method: 'popup_initiate_export_settings',
 		data: { fileContents: JSON.stringify(ExportedSettings.serialize(exportedSettings), undefined, 4) }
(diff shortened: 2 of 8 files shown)


app/ts/components/pages/SettingsView.tsx
```diff
@@ -22,6 +26,74 @@ function CheckBoxSetting(param: CheckBoxSettingParam) {
...
+			<div style = 'margin: 10px; background-color: var(--bg-color);'>
+				<DinoSays text = { 'Settings and address book loaded!' }/>
+			</div>
+		: <></> }
+		<div class = 'popup-button-row'>
+			<div style = 'display: flex; flex-direction: row;'>
+				<button className = 'button is-primary is-danger' style = 'flex-grow: 1; margin-left: 5px; margin-right: 5px;'>
+					Import settings
+					<input type = 'file' accept = '.json' onInput = { importSettings } style = 'position: absolute; width: 100%; height: 100%;'/>
+				</button>
```
MicahZoltu, line 88: I don't think this should be a button, as it doesn't function like a button.  The input is what matters here.  I also think you should just put the opacity here rather than in CSS since this is basically the only place it applies.  CSS is useful (for readability and performance) when you need to apply the same style to many elements.  However, if the style only applies to one element then inline helps with readability IMO.
```suggestion
				<label className = 'button is-primary is-danger' style = 'flex-grow: 1; margin-left: 5px; margin-right: 5px;'>
					Import settings
					<input type = 'file' accept = '.json' onInput = { importSettings } style = 'position: absolute; width: 100%; height: 100%; opacity: 0;'/>
				</label>
```

app/css/interceptor.css
```diff
@@ -707,3 +707,7 @@ svg.spinner > circle {
 	padding-bottom: 10px;
 	padding-top: 10px;
 }
+
+input[type = 'file'] {
+	opacity: 0    
+}
```
MicahZoltu, line 713: See comment below about why I think this should be deleted.
```suggestion
```

app/ts/background/popupMessageHandlers.ts
```diff
@@ -328,3 +329,33 @@ export async function changeSettings(simulator: Simulator, parsedRequest: Change
...
+		method: 'popup_initiate_export_settings_reply',
+		data: { success: true }
+	})
+}
+
+export async function exportSettings() {
+	const exportedSettings = await exportSettingsAndAddressBook()
+	await sendPopupMessageToOpenWindows({
+		method: 'popup_initiate_export_settings',
+		data: { fileContents: JSON.stringify(ExportedSettings.serialize(exportedSettings), undefined, 4) }
```
MicahZoltu, line 359: No bigints or Uint8Arrays will ever be serialized by this?  In my project I have these two functions to help with (de)serializing stuff that may contain bigints or Uint8Arrays:
```ts
export function jsonStringify(value: unknown, space?: string | number | undefined): string {
    return JSON.stringify(value, (_key, value) => {
		if (typeof value === 'bigint') return `0x${value.toString(16)}n`
		if (value instanceof Uint8Array) return `b'${Array.from(value).map(x => x.toString(16).padStart(2, '0')).join('')}'`
		return value
    }, space)
}
export function jsonParse(text: string): unknown {
	return JSON.parse(text, (_key: string, value: unknown) => {
		if (typeof value !== 'string') return value
		if (/^0x[a-fA-F0-9]+n$/.test(value)) return BigInt(value.slice(0, -1))
		const bytesMatch = /^b'(:<hex>[a-fA-F0-9])+'$/.exec(value)
		if (bytesMatch && 'groups' in bytesMatch && bytesMatch.groups && 'hex' in bytesMatch.groups) return hexToBytes(`0x${bytesMatch.groups['hex']}`)
		return value
	})
}
```

MicahZoltu (approved):

app/ts/components/pages/SettingsView.tsx
KillariDev: hmm. I feel it functions like a button. You press it and file selection dialog appears

app/ts/background/popupMessageHandlers.ts
KillariDev: serialize converts everything to dictionaries, numbers, strings ad arrays. so its all json compatible. If you want to serialize bigint, its converted to string with serialize. uint8Array's look bit funky as they become arrays of numbers, but are serializeables

app/ts/components/pages/SettingsView.tsx
MicahZoltu: But you aren't clicking the button, you are clicking the `Input` field.  The button doesn't actually behave like a button, it behaves like a label in the current code.  This is, in theory, particularly important for screen readers and such which will read this as a button wrapping an input field which is confusing.  A labelled input field on the other hand is much more clear as to what the association is.

KillariDev: ah, yeah. I guess a better way would be to make the input look like a button instead with CSS

MicahZoltu: Yes, that would be ideal.  The problem is that styling `input` fields is *really* hard.  You can do it, but it is a huge pain.

KillariDev: and you want me to go through huge pain, instead of using tricks like this? D:

MicahZoltu: 😆 no.  I actually started writing a comment suggesting that you style the input as a button but realized it would be too hard which is why I switched to just suggesting swapping the `button` element into a `label` element.

Issue #516, opened 2023-06-12
Error constants have incorrect values


sambacha:
https://github.com/DarkFlorist/TheInterceptor/blob/9ce3769113b31aadb3191318ce6ea8c28d02ad18/app/ts/utils/constants.ts#L66

`METAMASK_ERROR_INCOMPLETE_REQUEST` is suppose to be `PARSE_ERROR` see  https://github.com/ethereum/EIPs/blob/master/EIPS/eip-234.md

Moreover the Error constants for JSON-RPC are suppose to be *negative*. In the constant above, it should actually be `-32700` not `32700`. 

## Error Reference (annotations mine)

```ts

/**
 *
 * @export
 * @interface ErrorCodes
 * @see {@link https://github.com/MetaMask/rpc-errors/blob/main/src/error-constants.ts}
 *
 */

interface ErrorCodes {
  readonly rpc: {
    readonly static invalidInput       : -32_000;
    readonly static resourceNotFound   : -32_001;
    readonly static resourceUnavailable: -32_002;
    readonly static transactionRejected: -32_003;
    readonly static methodNotSupported : -32_004;
    readonly static limitExceeded      : -32_005;
    readonly static parse              : -32_700;
    readonly static invalidRequest     : -32_600;
    readonly static methodNotFound     : -32_601;
    readonly static invalidParams      : -32_602;
    readonly static internal           : -32_603;
  };
  readonly provider: {
    readonly static userRejectedRequest: 4_001;
    readonly static unauthorized       : 4_100;
    readonly static unsupportedMethod  : 4_200;
    readonly static disconnected       : 4_900;
    readonly static chainDisconnected  : 4_901;
  };
}

export const errorCodes: ErrorCodes = {
  rpc: {
    invalidInput       : -32_000,
    resourceNotFound   : -32_001,
    resourceUnavailable: -32_002,
    transactionRejected: -32_003,
    methodNotSupported : -32_004,
    limitExceeded      : -32_005,
    parse              : -32_700,
    invalidRequest     : -32_600,
    methodNotFound     : -32_601,
    invalidParams      : -32_602, /* Invalid transaction envelope type: specified type \"0x02\" but including maxFeePerGas and maxPriorityFeePerGas requires type: \"0x2\"", data: None })) */
    internal           : -32_603, /* Cannot read properties of undefined (reading 'message', data: Some(Object({"originalError": Object({}) })) */
  },
  provider: {
    userRejectedRequest: 4_001,
[...]

PR #523, opened 2023-06-14
make error constants negative

KillariDev:
fix https://github.com/DarkFlorist/TheInterceptor/issues/516


MicahZoltu (changes requested):
#516 also suggests correcting some numbers (not just negative sign).  We should either fix those as well in this PR, or remove the `fixes #516` from the description so that bug remains open.

KillariDev:
> #516 also suggests correcting some numbers (not just negative sign). We should either fix those as well in this PR, or remove the `fixes #516` from the description so that bug remains open.

I added question to the issue

KillariDev:
fixed it

MicahZoltu (approved):

Issue #484, opened 2023-06-04
Bug: When in signing mode without MM on a tab transaction is added to simulation stack.


MicahZoltu:
1. Make sure you are in signing mode in Interceptor.
2. Install MetaMask in Brave and restrict extension permissions to On click
![image](https://github.com/DarkFlorist/TheInterceptor/assets/886059/27a82618-3335-4ec5-bf8e-a30b9948657a)
3. Go into a tab and enable MetaMask by clicking on the extension icon.
4. Open a different dapp in another tab that has previously been approved access (MM + Interceptor) to the currently selected MM account.
5. Do **not** click on the MetaMask icon in the new tab (so MetaMask is not injected into the page).
6. Interact with the dapp and approve the transaction in Interceptor.
7. Notice that you don't get a MetaMask popup (expected).
8. Attempt to interact with the dapp again and notice that you get a nonce error.

The nonce error is because of #483.  That bug however lets us easily see the problem here, which is that your transaction is silently getting added to the simulation stack even though you are in signing mode.

PR #531, opened 2023-06-15
Allow each tab have their own signer account

KillariDev:
Currently interceptor forces signer to only have one address active at a time (and when this is not the case, interceptor behaves bit weirdly).

This change makes it so that every tab can have their own signer account.

This change also makes access management a bit simpler, now access is checked at the beginning of the RPC query and the active address doesn't change during lifespan of the request

Should fix https://github.com/DarkFlorist/TheInterceptor/issues/484

### DIFF
--- a/app/ts/background/background.ts
+++ b/app/ts/background/background.ts
@@ -4,7 +4,7 @@ import { Simulator } from '../simulation/simulator.js'
 import { EthereumJsonRpcRequest, EthereumQuantity, OldSignTypedDataParams, PersonalSignParams, SignTypedDataParams } from '../utils/wire-types.js'
 import { clearTabStates, getSignerName, getSimulationResults, removeTabState, setIsConnected, updateSimulationResults, updateTabState } from './storageVariables.js'
 import { changeSimulationMode, getSettings, getMakeMeRich } from './settings.js'
-import { blockNumber, call, chainId, estimateGas, gasPrice, getAccounts, getBalance, getBlockByNumber, getCode, getLogs, getPermissions, getSimulationStack, getTransactionByHash, getTransactionCount, getTransactionReceipt, personalSign, requestPermissions, sendRawTransaction, sendTransaction, subscribe, switchEthereumChain, unsubscribe } from './simulationModeHanders.js'
+import { blockNumber, call, chainId, estimateGas, gasPrice, getAccounts, getBalance, getBlockByNumber, getCode, getLogs, getPermissions, getSimulationStack, getTransactionByHash, getTransactionCount, getTransactionReceipt, personalSign, sendRawTransaction, sendTransaction, subscribe, switchEthereumChain, unsubscribe } from './simulationModeHanders.js'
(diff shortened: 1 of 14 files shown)


app/ts/background/background.ts
```diff
@@ -475,17 +477,17 @@ export function sendMessageToContentScript(websiteTabConnections: WebsiteTabConn
 	})
 }
 
-export async function handleContentScriptMessage(websiteTabConnections: WebsiteTabConnections, socket: WebsiteSocket, request: InterceptedRequest, website: Website) {
+export async function handleContentScriptMessage(websiteTabConnections: WebsiteTabConnections, socket: WebsiteSocket, request: InterceptedRequest, website: Website, activeAddress: bigint | undefined) {
```
MicahZoltu, line 480: Would optional more accurately represent the intent of this parameter, or is an explicit `undefined` expected?
```suggestion
export async function handleContentScriptMessage(websiteTabConnections: WebsiteTabConnections, socket: WebsiteSocket, request: InterceptedRequest, website: Website, activeAddress?: bigint) {
```

app/ts/background/simulationModeHanders.ts
```diff
@@ -32,35 +32,32 @@ export async function getTransactionReceipt(ethereumClientService: EthereumClien
 	return { result: EthTransactionReceiptResponse.serialize(await getSimulatedTransactionReceipt(ethereumClientService, simulationState, request.params[0])) }
 }
 
-function getFromField(websiteTabConnections: WebsiteTabConnections, simulationMode: boolean, transactionFrom: bigint | undefined, getActiveAddressForDomain: (websiteAccess: WebsiteAccessArray, websiteOrigin: string, settings: Settings) => bigint | undefined, socket: WebsiteSocket, settings: Settings) {
+function getFromField(websiteTabConnections: WebsiteTabConnections, simulationMode: boolean, transactionFrom: bigint | undefined, activeAddress: bigint | undefined, socket: WebsiteSocket) {
```
MicahZoltu, line 35: Consider re-arranging these parameters so you can have optional parameters if that is the intent.  My personal policy on parameter ordering is that "dependencies" should come first, followed by "parameters".  This is driven largely by the fact that this makes optional parameters possible.
```suggestion
function getFromField(websiteTabConnections: WebsiteTabConnections, socket: WebsiteSocket, simulationMode: boolean, transactionFrom?: bigint, activeAddress?: bigint) {
```

app/ts/background/simulationModeHanders.ts
```diff
@@ -32,35 +32,32 @@ export async function getTransactionReceipt(ethereumClientService: EthereumClien
...
-		return from
+		if (activeAddress === undefined) throw new Error('Access to active address is denied')
+		return activeAddress
 	}
 }
 
 export async function sendTransaction(
 	websiteTabConnections: WebsiteTabConnections,
-	getActiveAddressForDomain: (websiteAccess: WebsiteAccessArray, websiteOrigin: string, settings: Settings) => bigint | undefined,
+	activeAddress: bigint | undefined,
```
MicahZoltu, line 48: Same comment as above (and elsewhere), does optionality better express what you want, or do you expect an explicitly undefined `activeAddress`?

MicahZoltu (approved):

app/ts/background/background.ts
KillariDev: I guess with better wording. I want the active address be mandatory when it exists, and it would be a bug to leave it out when it exists

PR #532, opened 2023-06-16
fix etherscan

KillariDev:


app/ts/background/simulationModeHanders.ts
```diff
@@ -70,8 +70,8 @@ export async function sendTransaction(
 			from,
 			chainId: ethereumClientService.getChainId(),
 			nonce: await transactionCount,
-			maxFeePerGas: transactionDetails.maxFeePerGas != undefined ? transactionDetails.maxFeePerGas : parentBlock.baseFeePerGas * 2n,
-			maxPriorityFeePerGas: transactionDetails.maxPriorityFeePerGas != undefined  ? transactionDetails.maxPriorityFeePerGas : 10n**8n, // 0.1 nanoEth/gas
+			maxFeePerGas: transactionDetails.maxFeePerGas !== undefined && transactionDetails.maxFeePerGas !== null ? transactionDetails.maxFeePerGas : parentBlock.baseFeePerGas * 2n,
+			maxPriorityFeePerGas: transactionDetails.maxPriorityFeePerGas !== undefined && transactionDetails.maxPriorityFeePerGas !== null ? transactionDetails.maxPriorityFeePerGas : 10n**8n, // 0.1 nanoEth/gas
```
MicahZoltu, line 74: I wonder if it is worth adding a comment indicating that EtherScan sets these to null sometimes which is why we have them, that way if they eventually fix the bug we can go back and delete this cruft.

MicahZoltu (approved):

MicahZoltu:
We should cut a new release as soon as this is merged.

app/ts/background/simulationModeHanders.ts
KillariDev: I added comment to the type

PR #533, opened 2023-06-19
fix serialization issue of `input` field

KillariDev:


app/ts/utils/wire-types.ts
```diff
@@ -549,8 +549,9 @@ export const DappRequestTransaction = funtypes.Intersect(
 		maxFeePerGas: funtypes.Union(EthereumQuantity, funtypes.Null), // etherscan sets this field to null, remove this if etherscan fixes this
 	}),
 	funtypes.Union(
-		funtypes.ReadonlyPartial({ data: EthereumData }),
-		funtypes.ReadonlyPartial({ input: EthereumData })
+		funtypes.ReadonlyObject({ data: EthereumData }),
+		funtypes.ReadonlyObject({ input: EthereumData }),
+		funtypes.ReadonlyObject({})
```
MicahZoltu, line 554: Hmm, this feels risky.  I don't know that evaluation order is well defined in `funtypes`, so this may work today and break with a patch version bump to `funtypes`.  Since funtypes will ignore excess keys, I wonder if we would be better off handling this the same way we handle gas price, where we just put both of these on the partial above this.  Worst case scenario, we (de)serialize both and only one of them is used.

It would be *nice* to get automatic checking to make sure that we don't have both an `input` and a `data` property on the same object, just as it would be nice to verify that we don't have both a `gasPrice` and `max*FeePerGas` on the same object.  However, I am not sure that is reasonable given current tooling.

MicahZoltu (approved):

PR #540, opened 2023-06-21
Check if v#.#.# tag is being cut from head of release

epheph:
The idea is that we protect `release` behind two approvals. then we cut releases off release, and if you try to tag anything other than HEAD of release, we bail.

### DIFF
--- a/.github/workflows/release-extension.yml
+++ b/.github/workflows/release-extension.yml
@@ -19,6 +19,15 @@ jobs:
           GIT_COMMIT_SHA="${{ github.sha }}"
           TAG=$(basename "${{ github.ref }}")
           VERSION=${TAG#v}
+
+          RELEASE_SHA=$(git rev-parse origin/release)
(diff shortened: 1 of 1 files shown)


MicahZoltu (changes requested):
This doesn't align with my understanding of the goal.  My understanding is that we still can cut new GitHub releases easily (just tag `v#.#.#`) and this would generate artifacts and add them to GitHub releases.  the `Upload Firefox extension to mozilla addon` is the only part that shouldn't happen automatically when someone creates a version tag.

I believe we can just move the `Upload Firefox extension to mozilla addon` to a separate job, and move `MOZILLA_ADDON_USER` and `MOZILLA_ADDON_SECRET` to GitHub Environment Secrets (I already created a `Production` environment we can add to).  This will trigger approval gating on the job (already configured in GitHub environments) so even if the job is *triggered* by a version tag, it won't actually execute until after a human has gone in and manually approved the job to run and the necessary time delay (currently 7 days) has passed.

Issue #538, opened 2023-06-21
fix change chain dialog


KillariDev:
![image](https://github.com/DarkFlorist/TheInterceptor/assets/13102010/10d6c5d6-d570-4160-a290-59f06417f8bc)

PR #543, opened 2023-06-22
fix font in change chain

KillariDev:
fix https://github.com/DarkFlorist/TheInterceptor/issues/538
image

### DIFF
--- a/app/ts/components/pages/ChangeChain.tsx
+++ b/app/ts/components/pages/ChangeChain.tsx
@@ -87,7 +87,7 @@ export function ChangeChain() {
 					</article>
 					<div class = 'media-content' style = 'padding-bottom: 10px'>
 						<div class = 'content'>
-							<p className = 'title' style = 'white-space: normal; text-align: center;'>
+							<p className = 'title' style = 'white-space: normal; text-align: center; padding: 10px;'>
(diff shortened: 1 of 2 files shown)


app/ts/components/pages/ChangeChain.tsx
```diff
@@ -87,31 +87,27 @@ export function ChangeChain() {
...
-											text = { 'This chain is not supported by The Interceptor. Would you like to disable The Interceptor and attempt to connect anyway?' }
-											checked = { connectAnyway }
-											onInput = { setConnectAnyway }
-										/>
-									</div>
-								: <></> }
 							</p>
+							{ !chainChangeData.isInterceptorSupport && chainChangeData.simulationMode ?
+								<ErrorContainer
+									text = { 'This chain is not supported by The Interceptor. If you want to use this chain anyway. Select Signing mode instead of Simulation mode and attempt to change the chain again. You will then be able to disable The Interceptor and send transactions without protection of The Interceptor.' }
```
MicahZoltu, line 101:
```suggestion
									text = { 'This chain is not supported by The Interceptor. If you want to use this chain anyway. Select Signing mode instead of Simulation mode and attempt to change the chain again. You will then be able to disable The Interceptor and send transactions without its protection.' }
```

MicahZoltu (approved):

PR #544, opened 2023-06-22
add constraint to DappRequestTransaction

KillariDev:
fix https://github.com/DarkFlorist/TheInterceptor/pull/533

### DIFF
--- a/app/ts/utils/wire-types.ts
+++ b/app/ts/utils/wire-types.ts
@@ -503,19 +503,18 @@ export const DappRequestTransaction = funtypes.ReadonlyPartial({
 	maxFeePerGas: funtypes.Union(EthereumQuantity, funtypes.Null), // etherscan sets this field to null, remove this if etherscan fixes this
 	data: EthereumData,
 	input: EthereumData,
-})
-	.withConstraint((dappRequestTransaction) => dappRequestTransaction.input !== undefined && dappRequestTransaction.data !== undefined ? areEqual(dappRequestTransaction.input, dappRequestTransaction.data) : true)
-	.withConstraint((x) => {
-		if (x.gasPrice !== undefined) {
-			return x.maxPriorityFeePerGas === undefined && x.maxFeePerGas === undefined
-		} else if (x.maxPriorityFeePerGas !== undefined) {
-			return x.maxFeePerGas !== undefined && x.gasPrice === undefined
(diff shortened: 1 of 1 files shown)


app/ts/utils/wire-types.ts
```diff
@@ -492,22 +493,17 @@ export type ToWireType<T> =
...
+	from: EthereumAddress,
+	gas: EthereumQuantity,
+	value: EthereumQuantity,
+	to: funtypes.Union(EthereumAddress, funtypes.Null),
+	gasPrice: EthereumQuantity,
+	maxPriorityFeePerGas: funtypes.Union(EthereumQuantity, funtypes.Null), // etherscan sets this field to null, remove this if etherscan fixes this
+	maxFeePerGas: funtypes.Union(EthereumQuantity, funtypes.Null), // etherscan sets this field to null, remove this if etherscan fixes this
+	data: EthereumData,
+	input: EthereumData,
+}).withConstraint((dappRequestTransaction) => dappRequestTransaction.input !== undefined && dappRequestTransaction.data !== undefined ? areEqual(dappRequestTransaction.input, dappRequestTransaction.data) : true)
```
MicahZoltu, line 506: You might be able to simplify these suggested conditionals, but be careful to ensure the code is still readable.
```suggestion
})
	.withConstraint((dappRequestTransaction) => dappRequestTransaction.input !== undefined && dappRequestTransaction.data !== undefined ? areEqual(dappRequestTransaction.input, dappRequestTransaction.data) : true)
	.withConstraint((x) => {
		if (x.gasPrice !== undefined) {
			return x.maxPriorityFeePerGas === undefined && x.maxFeePerGas === undefined
		} else if (x.maxPriorityFeePerGas !== undefined) {
			return x.maxFeePerGas !== undefined && x.gasPrice === undefined
		} else if (x.maxFeePerGas !== undefined) {
			return x.maxPriorityFeePerGas !== undefined && x.gasPrice === undefined
		} else {
			return true
		}
	}
```

MicahZoltu (approved):
I think we should fix both of these while we are in here, but the current change is not buggy so approved if you decide to do this in separate PRs.

app/ts/utils/wire-types.ts
KillariDev: I think this is quite readable and simple enough

PR #546, opened 2023-06-23
get rid of anys in inpage

KillariDev:


app/inpage/ts/inpage.ts
```diff
@@ -362,19 +363,24 @@ class InterceptorMessageListener {
...
 			if (forwardRequest.requestId === undefined) return
 			if (error instanceof Error) {
-				if (!('code' in error)) (error as any).code = -32603
-				if (!('data' in error) || (error as any).data === undefined || (error as any).data === null) (error as any).data = { request: forwardRequest }
-				else if (!('request' in (error as any).data)) (error as any).data.request = forwardRequest
-				return this.outstandingRequests.get(forwardRequest.requestId)!.reject(error)
+				const inputedErrorObject = {
+					...error,
+					code: ('code' in error && typeof error.code !== 'number') ? error.code : -32603,
+					data: !('data' in error) || error.data === undefined || error.data === null || typeof error.data !== 'object' ? { request: forwardRequest } : { ...error.data, request: forwardRequest }
```
MicahZoltu, line 375: If `error.data` is a byte array or a string, this will overwrite it with a `{ request }`.  Is that really what we want?

MicahZoltu (approved):

app/inpage/ts/inpage.ts
KillariDev: objects that can be whatever are annoying :(

KillariDev: I decided that in that case we make error.data an object with request and old error.data field. I guess that's a good solution to this?

MicahZoltu: Works for me.

PR #552, opened 2023-06-26
fix(build): dependency transitive fixes and docker update

sambacha:
Building the extension on my local machine proved missing transitive dependencies:

```bash

> interceptor-extension@ setup-chrome /Users/janitor/TheInterceptor
> npm run vendor && npm run inpage && npm run build-chrome


> vendor
> cd build && npm ci --ignore-scripts && npm run vendor


added 3 packages, and audited 4 packages in 453ms

found 0 vulnerabilities

> vendor
> tsc && node --enable-source-maps ./vendor.mjs && node --input-type=module -e "import { promises as fs } from 'fs'; await fs.rm('./bundler.mjs'); await fs.rm('./vendor.mjs');"

[Error: ENOENT: no such file or directory, scandir '/Users/janitor/TheInterceptor/node_modules/@noble/secp256k1'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'scandir',
  path: '/Users/janitor/TheInterceptor/node_modules/@noble/secp256k1'
}
 ELIFECYCLE  Command failed with exit code 1.

```

This was only one of the errors.

Also fix Dockerfile `RUN` command arguments. Also changed pinned docker image to the alpine version that is available cached on GitHub actions machine by default.

TOODLES!


pnpm-lock.yaml
MicahZoltu, line 1: We don't use pnpm, so this should be left out.

Dockerfile
```diff
@@ -1,6 +1,8 @@
-FROM node:18.16.0-alpine3.17@sha256:44aaf1ccc80eaed6572a0f2ef7d6b5a2982d54481e4255480041ac92221e2f11
+# syntax=docker/dockerfile:1.4
 
-RUN apk --no-cache add zip
+FROM node:18-alpine@sha256:9036ddb8252ba7089c2c83eb2b0dcaf74ff1069e8ddf86fe2bd6dc5fecc9492d
```
MicahZoltu, line 3: I'm pretty hesitant to change this.  As you can see, it breaks some things (like apparently not having `zip` on it).  At the least we should specify the exact version rather than using `18-alpine`, just so it makes it easier to verify the sha256 in the future.

Dockerfile
```diff
@@ -1,6 +1,8 @@
...
+
+RUN apk add --no-cache zip
```
MicahZoltu, line 5: Hmm, `zip` should exist in the image already.  Did you run into a problem with it being missing?  Possibly related to changing the base image.

docker-bake.hcl
MicahZoltu, line 1: What is a `.hcl` file?

package-lock.json
MicahZoltu, line 1: package-lock.json files should not be removed for applications.  It prevents reproducible builds.

build/package-lock.json
MicahZoltu, line 1: package-lock.json files should be retained so builds are reproducible.

package.json
```diff
@@ -17,16 +17,19 @@
 	},
 	"dependencies": {
 		"@darkflorist/address-metadata": "0.2.0",
-		"@noble/curves": "0.8.0",
+		"@noble/curves": "1.1.0",
 		"@noble/hashes": "1.2.0",
+		"@noble/secp256k1": "2.0.0",
 		"@preact/signals": "1.1.3",
+		"@preact/signals-core": "1.3.0",
```
MicahZoltu, line 24: `@noble/curves` is a superset of `@noble/secp256k1`.  While I wouldn't mind switching from curves to secp256k1, should not need both.  `signals-core` should be a transitive dependency of `signals`, though if we reference `signals-core` directly anywhere we should include it in `package.json` dependencies.

package.json
```diff
@@ -17,16 +17,19 @@
...
 		"ethers": "6.4.0",
 		"funtypes": "5.0.3",
 		"preact": "10.8.1",
 		"webextension-polyfill": "0.10.0"
 	},
 	"devDependencies": {
 		"@types/firefox-webext-browser": "109.0.0",
+		"@types/node": "^18",
```
MicahZoltu, line 32: Adding `@types/node` needs extra care to be included in a web project, else you won't get compiler errors on trying to reference Node-only globals and imports from the web project (which will be runtime failures).  I think thus far we have managed to avoid it.

.github/workflows/docker-release.yml
```diff
@@ -0,0 +1,90 @@
...
+                  context: .
+                  push: true
+                  tags: ${{ steps.docker_tagging.outputs.docker_tags }}
+                  labels: ${{ steps.meta.outputs.labels }}
+                  cache-from: type=gha
+                  cache-to: type=gha,mode=max
+                  build-args: |
+                      BUILDTIME=${{ fromJSON(steps.meta.outputs.json).labels['org.opencontainers.image.created'] }}
+                      VERSION=${{ fromJSON(steps.meta.outputs.json).labels['org.opencontainers.image.version'] }}
+                      REVISION=${{ fromJSON(steps.meta.outputs.json).labels['org.opencontainers.image.revision'] }}
```
MicahZoltu, line 90: What is the purpose of pushing this as a docker image?  This project builds a browser extension, which doesn't really make sense to use inside of a docker image that I can think of.  We already do builds inside of docker for reproducibility/consistency, but the image is discarded upon build completion.

Dockerfile
```diff
@@ -36,4 +38,15 @@ WORKDIR /
...
+
+LABEL org.label-schema.build-date=$BUILD_DATE \
+      org.label-schema.name="The Interceptor Browser Extenstion" \
+      org.label-schema.description="TheInterceptor" \
+      org.label-schema.url="https://dark.florist/" \
+      org.label-schema.vcs-ref=$VCS_REF \
+      org.label-schema.vcs-url="https://github.com/DarkFlorist/TheInterceptor.git" \
+      org.label-schema.vendor="Dark Florist" \
+      org.label-schema.version=$VERSION \
+      org.label-schema.schema-version="1.0"
```
MicahZoltu, line 52: Similar to comment above, these seem unnecessary as the docker image is only used transiently during the build process but it is not published/retained.

.dockerignore
```diff
@@ -0,0 +1,5 @@
+target
```
MicahZoltu, line 1: We don't generate a `target` folder I don't believe.
```suggestion
```

.dockerignore
```diff
@@ -0,0 +1,5 @@
...
+.github
+.git
+node_modules
+.env
```
MicahZoltu, line 5: We don't have/use `.env` files in this project.
```suggestion
```

MicahZoltu (changes requested):
Regarding the build failure, did you do `npm ci --ignore-scripts` prior to trying to run the other scripts?  That should have picked up the necessary transitive dependencies that it seems like you are missing.

KillariDev:
closing as wont-fix and stale

sambacha:
> closing as wont-fix and stale

Was discussed in discord iirc, should have closed after. Danke

PR #555, opened 2023-06-27
Get rid of hard coded RPC list and instead fetch it from storage

KillariDev:
fix https://github.com/DarkFlorist/TheInterceptor/issues/535
- Also refactor in a way that pure `chainId`'s are not moved around, but instead network objects that contain RPC url, name, native token info etc


app/ts/background/settings.ts
```diff
@@ -20,6 +20,61 @@ export const defaultAddresses = [
...
+		name: 'Eth (neth-multi)',
+		chainId: 1n,
+		https_rpc: 'https://rpc.dark.florist/birdchalkrenewtip',
+		currencyName: 'Ether',
+		currencyTicker: 'ETH',
+		primary: false,
+		minimized: true,
+		weth: 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2n,
+	},
+	*/
```
MicahZoltu, line 75: Feels like we should leave these in and just discourage their use somehow.  Would rather not have to wait for a plugin update just to troubleshoot them.

MicahZoltu (approved):

app/ts/background/settings.ts
KillariDev: well, they are completely broken currently as they don't support our multicall method. So I don't think they should be added

Issue #557, opened 2023-06-28
writing random text to address field when adding new active address breaks the dialog and produces following error


KillariDev:
![image](https://github.com/DarkFlorist/TheInterceptor/assets/13102010/607d65e8-d00f-4881-841d-4db28f5b72a9)

PR #561, opened 2023-06-29
Trim address on active address adding and don't crash on non-valid address

KillariDev:
fix https://github.com/DarkFlorist/TheInterceptor/issues/557

### DIFF
--- a/app/ts/utils/bigint.ts
+++ b/app/ts/utils/bigint.ts
@@ -84,6 +84,13 @@ export function checksummedAddress(address: bigint) {
 	return ethers.getAddress(addressString(address))
 }
 
+export function stringToAddress(addressString: string | undefined) {
+	if (addressString === undefined) return undefined
+	const trimmedAddress = addressString.trim()
+	if (!ethers.isAddress(trimmedAddress)) return undefined
+	return BigInt(trimmedAddress)
+}
+
 export function bytes32String(bytes32: bigint) {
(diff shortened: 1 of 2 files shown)


app/ts/utils/bigint.ts
```diff
@@ -84,6 +84,13 @@ export function checksummedAddress(address: bigint) {
 	return ethers.getAddress(addressString(address))
 }
 
+export function stringToAddress(addressString: string | undefined) {
+	if (addressString === undefined) return undefined
+	const trimmedAddress = addressString.trim()
+	if (!ethers.isAddress(trimmedAddress)) return undefined
```
MicahZoltu, line 90: Feels like this could just be a regex instead.

KillariDev: I feel regex is more complex than using ethers method?

MicahZoltu: 🤷 
```ts
/0x[a-fA-F0-9]{40}/.test(trimmedAddress)
ethers.isAddress(trimmedAddress)
```
If Ethers is doing checksumming then I agree with you.

KillariDev: yeah, I feel `isAddress` is more clear than regexp there. If done with regexp, I would wrap that in "isAddress" function anyway to comment it

MicahZoltu (approved):
Sorry, not sure why this was a comment instead of an approval.

PR #562, opened 2023-06-29
Fix view refreshing when using tabs

KillariDev:
fix https://github.com/DarkFlorist/TheInterceptor/issues/492
when tab is refreshed it loses it's state. This PR makes it so that on refresh the tab will request the data from storage again. This system is added to all popup views that we have (addressbook works with refresh)

### DIFF
--- a/app/ts/background/windows/confirmTransaction.ts
+++ b/app/ts/background/windows/confirmTransaction.ts
@@ -3,7 +3,7 @@ import { EthereumClientService } from '../../simulation/services/EthereumClientS
 import { appendTransaction } from '../../simulation/services/SimulationModeEthereumClientService.js'
 import { ERROR_INTERCEPTOR_NO_ACTIVE_ADDRESS, METAMASK_ERROR_NOT_CONNECTED_TO_CHAIN, METAMASK_ERROR_USER_REJECTED_REQUEST } from '../../utils/constants.js'
 import { Future } from '../../utils/future.js'
-import { ConfirmTransactionTransactionSingleVisualization, ExternalPopupMessage, InterceptedRequest, PendingTransaction, TransactionConfirmation } from '../../utils/interceptor-messages.js'
+import { InterceptedRequest, TransactionConfirmation } from '../../utils/interceptor-messages.js'
 import { Semaphore } from '../../utils/semaphore.js'
 import { WebsiteSocket, WebsiteTabConnections } from '../../utils/user-interface-types.js'
 import { EstimateGasError, WebsiteCreatedEthereumUnsignedTransaction } from '../../utils/visualizer-types.js'
@@ -17,15 +17,21 @@ let openedDialog: PopupOrTab | undefined = undefined
 let pendingTransactions = new Map<number, Future<Confirmation>>()
 const pendingConfirmationSemaphore = new Semaphore(1)
 
-async function updateConfirmTransactionViewWithPendingTransactionOrClose() {
+export async function updateConfirmTransactionViewWithPendingTransaction() {
 	const promises = await getPendingTransactions()
 	if (promises.length >= 1) {
-		return sendPopupMessageToOpenWindows({
+		await sendPopupMessageToOpenWindows({
 			method: 'popup_update_confirm_transaction_dialog',
 			data: promises.map((p) => p.simulationResults),
 		})
+		return true
 	}
-	if (openedDialog) closePopupOrTab(openedDialog)
+	return false
+}
+
+async function updateConfirmTransactionViewWithPendingTransactionOrClose() {
+	if (await updateConfirmTransactionViewWithPendingTransaction() === true) return
+	if (openedDialog) await closePopupOrTab(openedDialog)
 	openedDialog = undefined
 }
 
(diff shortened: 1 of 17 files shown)


app/ts/components/App.tsx
```diff
@@ -38,6 +38,7 @@ export function App() {
 	const [isConnected, setIsConnected] = useState<IsConnected>(undefined)
 	const [useTabsInsteadOfPopup, setUseTabsInsteadOfPopup] = useState<boolean | undefined>(undefined)
 	const [currentTabId, setCurrentTabId] = useState<number | undefined>(undefined)
+	const [rpcEntries, setRpcEntries] = useState<RpcEntries>([])
```
MicahZoltu, line 41: Still resisting `Signal`s?

app/ts/components/pages/InterceptorAccess.tsx
```diff
@@ -190,16 +190,16 @@ export function InterceptorAccess() {
...
 				return
 			}
 		}
 		browser.runtime.onMessage.addListener(popupMessageListener)
-		sendPopupMessageToBackgroundPage({ method: 'popup_interceptorAccessReadyAndListening' })
 		return () => browser.runtime.onMessage.removeListener(popupMessageListener)
 	})
 
+	
+	useEffect(() => { sendPopupMessageToBackgroundPage({ method: 'popup_interceptorAccessReadyAndListening' }) }, [])
```
MicahZoltu, line 201: `[]` as second parameter to `useEffect` means it will run once when the component is first instantiated.  What does missing parameter mean (like the previous `useEffect`)?  The docs don't seem to specify.  Is it the same as `[]`?  If so, we should probably be consistent on which one we use.

MicahZoltu (approved):
This is a fairly large change and I only gave it a cursory review.  Recommend getting another set of eyes on it before merging.

app/ts/components/App.tsx
KillariDev: no, but I feel it requires bigger change to port over :/

app/ts/components/pages/InterceptorAccess.tsx
KillariDev: they are not the same, [] means it runs once, and missing parameter means its run everytime the component refreshes

KillariDev: you can read about that eg in here: <https://stackoverflow.com/questions/55240526/useeffect-second-argument-variations-in-react-hook>

app/ts/background/windows/confirmTransaction.ts
```diff
@@ -122,7 +106,17 @@ export async function openConfirmTransactionDialog(
 				})
 			}
 			if (openedDialog?.windowOrTab.id === undefined) return false
-			await resolveAppendPromise(openedDialog.windowOrTab.id, await refreshSimulationPromise)
+			await appendPendingTransaction({
+				dialogId: openedDialog.windowOrTab.id,
+				socket: socket,
```
jubalm, line 111: can use a shorthand property

app/ts/background/windows/interceptorAccess.ts
```diff
@@ -181,17 +174,17 @@ export async function requestAccessFromUser(
...
+			return await tryFocusingTabOrWindow({ type: openedDialog.popupOrTab.type === 'tab' ? 'tab' : 'window', id: openedDialog.popupOrTab.windowOrTab.id })
 		}
 		pendingAccessRequests.resolve(requests)
 	})
 }
 
 async function updateViewOrClose() {
 	const promises = await getPendingAccessRequests()
-	if (promises.length > 0) return sendPopupMessageToOpenWindows({ method: 'popup_update_access_dialog', data: promises })
+	if (promises.length > 0) return sendPopupMessageToOpenWindows({ method: 'popup_interceptorAccessDialog', data: promises })
```
jubalm, line 186: why the change from snake case to camel case on this method?

app/ts/background/windows/personalSign.ts
```diff
@@ -30,9 +30,17 @@ export async function resolvePersonalSign(websiteTabConnections: WebsiteTabConne
 		const resolved = await resolve(confirmation, data.simulationMode, data.params)
 		postMessageIfStillConnected(websiteTabConnections, data.socket, { ...data.params, ...resolved, requestId: confirmation.data.requestId })
 	}
+	if (openedDialog) await closePopupOrTab(openedDialog)
 	openedDialog = undefined
 }
 
+export async function updatePendingPersonalSignViewWithPendingRequests(ethereumClientService: EthereumClientService) {
+	const request = await getPendingPersonalSignPromise()
+	if (request != undefined) {
```
jubalm, line 39: looks like not all conditions are returned from this function?

app/ts/components/subcomponents/ChainSelector.tsx
```diff
@@ -42,7 +31,7 @@ export function ChainSelector(params: ChainSelectorParams) {
 		<div class = 'dropdown-menu' id = 'dropdown-menu' role = 'menu' style = 'right: -10px; min-width: 160px; left: unset'>
 			<div class = 'dropdown-content'>
 				{
-					rpcEntries === undefined ? <></> : rpcEntries.map((rpcEntry) => { return (
+					params.rpcEntries.map((rpcEntry) => { return (
 						<a href = '#' class = { `dropdown-item ${ rpcEntry.httpsRpc === params.rpcNetwork.httpsRpc ? 'is-active' : '' }` } onClick = { () => changeRpc(rpcEntry) } >
```
jubalm, line 35: this feels like a good place for a button element instead of anchor

app/ts/background/windows/interceptorAccess.ts
KillariDev: the idea is to keep `popup_` prefix and then move to camelcase for all messages. As they are currently quite mixed. Not sue what is the best strategy with these 🤔

app/ts/background/windows/personalSign.ts
KillariDev: it returns void so it doesn't need to return in all conditions

app/ts/components/subcomponents/ChainSelector.tsx
KillariDev: hmm, you feel the styling of button would be better than anchors?

PR #563, opened 2023-06-30
add 25% to gas estimations and set gas prices for estimate gas if set

KillariDev:


app/ts/simulation/services/SimulationModeEthereumClientService.ts
```diff
@@ -75,13 +75,21 @@ export const simulateEstimateGas = async (ethereumClientService: EthereumClientS
 	const transactionCount = getSimulatedTransactionCount(ethereumClientService, simulationState, sendAddress)
 	const block = await ethereumClientService.getBlock()
 	const maxGas = simulationGasLeft(simulationState, block)
+
+	const getGasPriceFields = (data: DappRequestTransaction) => {
+		if (data.gasPrice !== undefined) return { maxFeePerGas: data.gasPrice, maxPriorityFeePerGas: data.gasPrice }
```
MicahZoltu, line 80:
```suggestion
		if (data.gasPrice !== undefined && data.gasPrice !== null) return { maxFeePerGas: data.gasPrice, maxPriorityFeePerGas: data.gasPrice }
```

app/ts/simulation/services/SimulationModeEthereumClientService.ts
```diff
@@ -75,13 +75,21 @@ export const simulateEstimateGas = async (ethereumClientService: EthereumClientS
...
+		if (data.maxPriorityFeePerGas !== undefined && data.maxPriorityFeePerGas !== null && data.maxFeePerGas !== undefined && data.maxFeePerGas !== null) {
+			return { maxFeePerGas: data.maxFeePerGas, maxPriorityFeePerGas: data.maxPriorityFeePerGas }
+		}
+		return { maxFeePerGas: 0n, maxPriorityFeePerGas: 0n }
```
MicahZoltu, line 84: I feel like these fields shouldn't be included if they aren't set, rather than setting them to 0.  Setting to 0 could cause some client to incorrectly try to run the transaction with 0 gas price which would fail.
```suggestion
		return {}
```

MicahZoltu (approved):
Feels like we should do just the 25% change as a hotfix, and the improvement to gas price handling as a separate PR against main.

app/ts/simulation/services/SimulationModeEthereumClientService.ts
KillariDev: gasprice cannot be null by its current type

app/ts/simulation/services/SimulationModeEthereumClientService.ts
KillariDev: because we are using our multicall, that works with transactions, we need to format the calls to be proper transactions. Transaction needs to have its gasprices defined

MicahZoltu: Shouldn't these be set to valid gas prices then?  `0` is an invalid `maxFeePerGas`.

KillariDev: it's not invalid gas price in our plugin

MicahZoltu: Hmm, interesting.  These will likely become invalid once we switch to official client, so we may want to put in reasonable values.  Do we already have the latest block in memory so we could just use the base fee from that at least?  The priority fee doesn't matter much, but 0.1 seems to be standard now.

KillariDev: If user estimates gas with no gas prices set, I don't feel we should add gas prices to the estimation. Not sure how the normal estimate gas function handles this

KillariDev: I don't think it would become invalid, as I feel those values are not checked. I thought we agreed so?

MicahZoltu: > If user estimates gas with no gas prices set, I don't feel we should add gas prices to the estimation. Not sure how the normal estimate gas function handles this

I agree with this, but you aren't leaving them out here.  You are explicitly setting them.  `0 !== undefined`.

MicahZoltu: > I don't think it would become invalid, as I feel those values are not checked. I thought we agreed so?

In non-validating mode they won't be checked.  In validating mode they will be, and I think we will want to use validating mode.

PR #566, opened 2023-07-03
V3 spot prices for estimate ethereum prices for tokens

### LINKED ISSUE #514: Add Uniswap V3 (and other?) exchanges to token price estimator
https://github.com/DarkFlorist/TheInterceptor/blob/9ce3769113b31aadb3191318ce6ea8c28d02ad18/app/ts/simulation/priceEstimator.ts#L24

0xjimmy:
Uses spot price from pool with highest TVL.
Does not like USDT, need to fix.

Resolves #514


]
MicahZoltu:1]: Why is this file named `]`?

app/ts/simulation/priceEstimator.ts
```diff
@@ -21,99 +22,96 @@ export class PriceEstimator {
...
+	public async estimateEthereumPricesForTokens(tokens: TokenDecimals[], quote?: TokenDecimals) : Promise<TokenPriceEstimate[]> {
 		if (tokens.length == 0) return []
 
-		const chainId = await this.ethereum.getChainId()
+		const chainId = this.ethereum.getChainId()
 		const chainString = chainId.toString()
 		if (!isSupportedChain(chainString)) return []
+		// Support only mainnet for now
+		// TODO: fix this
+		if (chainId !== 1n) return []
```
MicahZoltu, line 33: Should this be fixed before we merge?

app/ts/simulation/priceEstimator.ts
```diff
@@ -21,99 +22,96 @@ export class PriceEstimator {
...
-		let inOutResults: TokenPriceEstimate[] = []
-		const swapInterface = new ethers.Interface(ABI)
-		for ( const token of tokens) {
-			if ( token.token === CHAINS[chainString].weth ) {
-				inOutResults.push({
-					token: addressString(CHAINS[chainString].weth),
-					inOutAmount: [10n ** 18n, 10n ** 18n],
-					decimals: 18n,
+		const quoteToken = quote ?? {
+			address: BigInt(CHAINS[chainString].weth),
```
MicahZoltu, line 36: Would we use `weth` as the quote token for other chains as well (e.g., Optimism/Arbitrum/BNB)?  I wonder `CHAINS` should contain a `quoteToken` so we can define a different quote token for each chain?

app/ts/simulation/priceEstimator.ts
```diff
@@ -21,99 +22,96 @@ export class PriceEstimator {
...
+			const callData = stringToUint8Array(IMulticall3.encodeFunctionData('aggregate3', [uniswapSpotCalls]))
+			const callTransaction = {
+				type: '1559',
+				from: MOCK_ADDRESS,
+				to: MULTICALL3,
+				value: 0n,
+				input: callData,
+				maxFeePerGas: 0n,
+				maxPriorityFeePerGas: 0n,
+				gasLimit: 15_000_000n,
```
MicahZoltu, line 67: This limit seems excessive, and it could result in unnecessary failures during multicalls with large simulation stacks.  Can we come up with an upper bound for this that is a bit more reasonable?  I don't know what a "normal" version of these calls costs, but maybe we can just take normal and multiply it by 2 or 3 or something?

app/ts/utils/uniswap.ts
```diff
@@ -0,0 +1,159 @@
...
+		keccak256(solidityPacked(['address', 'address'], [token0, token1])),
+		'0x96e8ac4277198ff8b6f785478aa9a39f403cb768dd02cbee326c3e7da348845f'
+	)
+
+	const v3 = {
+		'100': getCreate2Address(
+			addressString(UNISWAP_V3_FACTORY_ADDRESS),
+			keccak256(
+				abi.encode(['address', 'address', 'uint24'], [token0, token1, 100])
+			),
```
MicahZoltu, line 39: These newlines don't seem to be adding value, recommend removing them.
```suggestion
			keccak256(abi.encode(['address', 'address', 'uint24'], [token0, token1, 100])),
```

app/ts/utils/uniswap.ts
```diff
@@ -0,0 +1,159 @@
...
+			'0xe34f199b19b2b4f47f68442619d555527d244f78a3297ea89325f843f87b8b54'
+		),
+		'500': getCreate2Address(
+			addressString(UNISWAP_V3_FACTORY_ADDRESS),
+			keccak256(
+				abi.encode(['address', 'address', 'uint24'], [token0, token1, 500])
+			),
```
MicahZoltu, line 46:
```suggestion
			keccak256(abi.encode(['address', 'address', 'uint24'], [token0, token1, 500])),
```

app/ts/utils/uniswap.ts
```diff
@@ -0,0 +1,159 @@
...
+			'0xe34f199b19b2b4f47f68442619d555527d244f78a3297ea89325f843f87b8b54'
+		),
+		'3000': getCreate2Address(
+			addressString(UNISWAP_V3_FACTORY_ADDRESS),
+			keccak256(
+				abi.encode(['address', 'address', 'uint24'], [token0, token1, 3000])
+			),
```
MicahZoltu, line 53:
```suggestion
			keccak256(abi.encode(['address', 'address', 'uint24'], [token0, token1, 3000])),
```

app/ts/utils/uniswap.ts
```diff
@@ -0,0 +1,159 @@
...
+			'0xe34f199b19b2b4f47f68442619d555527d244f78a3297ea89325f843f87b8b54'
+		),
+		'10000': getCreate2Address(
+			addressString(UNISWAP_V3_FACTORY_ADDRESS),
+			keccak256(
+				abi.encode(['address', 'address', 'uint24'], [token0, token1, 10000])
+			),
```
MicahZoltu, line 60:
```suggestion
			keccak256(abi.encode(['address', 'address', 'uint24'], [token0, token1, 10000])),
```

app/ts/utils/uniswap.ts
```diff
@@ -0,0 +1,159 @@
...
+	}
+}
+
+interface Multicall3Call {
+	target: string
+	allowFailure: boolean
+	callData: string
+}
+
+export function calculateUniswapPools(tokenA: EthereumAddress, tokenB: EthereumAddress): UniswapPools {
```
MicahZoltu, line 24: I believe this function only works on mainnet, should we consider renaming this to something like `calculateUniswapPoolsEthereumMainnet` or something?  Perhaps we should have the factory address as a parameter so this function can be reused on any chain?

MicahZoltu (commented):
Will leave this for @KillariDev to review when he gets back, but I left some initial feedback.

While I'm not against shipping support for Ethereum mainnet only for now, we should probably think about and have a plan for how we will eventually extend this to other blockchains (like testnets) in the future.

app/ts/simulation/priceEstimator.ts
0xjimmy: Yeah needs to be fixed before merge.

epheph:
#514

app/ts/simulation/priceEstimator.ts
```diff
@@ -21,99 +20,92 @@ export class PriceEstimator {
...
-		let inOutResults: TokenPriceEstimate[] = []
-		const swapInterface = new ethers.Interface(ABI)
-		for ( const token of tokens) {
-			if ( token.token === CHAINS[chainString].weth ) {
-				inOutResults.push({
-					token: addressString(CHAINS[chainString].weth),
-					inOutAmount: [10n ** 18n, 10n ** 18n],
-					decimals: 18n,
+		// TODO: Support only mainnet for now as we only know Uniswap V2 and V3 on mainnet for pricing.
+		if (chainId !== 1n) return []
```
KillariDev, line 31: we cannot support goerli and sepolia?

app/ts/simulation/priceEstimator.ts
KillariDev: currently this `weth´ needs to correspond something that ~equals the chains main coin. I feel the strategy would be to always quote in this, and then add a separate token that we would swap to users preference token, which would be a separate setting. I feel this change should not be part of this PR but another one

app/ts/simulation/priceEstimator.ts
```diff
@@ -21,99 +20,92 @@ export class PriceEstimator {
...
+			const callData = stringToUint8Array(IMulticall3.encodeFunctionData('aggregate3', [uniswapSpotCalls]))
+			const callTransaction = {
+				type: '1559',
+				to: MULTICALL3,
+				value: 0n,
+				input: callData,
 			}
+			const multicallReturnData: { success: boolean, returnData: string }[] = IMulticall3.decodeFunctionResult('aggregate3', await this.ethereum.call(callTransaction, 'latest'))[0]
+
+			const prices = multicallReturnData.reduce((prices, { success, returnData }, index): { price: bigint, k: bigint }[] => {
```
KillariDev, line 65: what is this index? could it be named better

app/ts/simulation/priceEstimator.ts
```diff
@@ -21,99 +20,92 @@ export class PriceEstimator {
...
+				if (index > 4) return prices
```
KillariDev, line 66: Where does its `4` come from?

app/ts/simulation/priceEstimator.ts
```diff
@@ -21,99 +20,92 @@ export class PriceEstimator {
...
+
+				// V2 Pool
+				if (index === 0 && success) {
+					const { reserve0, reserve1 } = IUniswapV2Pool.decodeFunctionResult('getReserves', returnData)
+
+					const price = poolAddresses.token0IsQuote
+						? (reserve1 * (10n ** BigInt(token.decimals))) / reserve0
```
KillariDev, line 73: there's decimals should already be Bigints

app/ts/simulation/priceEstimator.ts
```diff
@@ -21,99 +20,92 @@ export class PriceEstimator {
...
+						: (reserve0 * (10n ** BigInt(token.decimals))) / reserve1
+
+					return [...prices, { price, k: BigInt(reserve0 * reserve1) }]
```
KillariDev, line 76: I feel these reserves should already be bigints?

app/ts/simulation/priceEstimator.ts
```diff
@@ -21,99 +20,92 @@ export class PriceEstimator {
...
+				}
+
+				// V3 Pool
+				if (success && multicallReturnData[index * 2 + 3].success && multicallReturnData[index * 2 + 4].success) {
+					const { sqrtPriceX96 } = IUniswapV3Pool.decodeFunctionResult('slot0', returnData)
+
+					const price = poolAddresses.token0IsQuote
```
KillariDev, line 83: could this logic be wrapped in a function with a refence to a source where it comes from?

KillariDev: the decimals also should already be bigints

app/ts/simulation/priceEstimator.ts
```diff
@@ -21,99 +20,92 @@ export class PriceEstimator {
...
+						? ((sqrtPriceX96 * (10n ** BigInt(token.decimals)) / (2n ** 96n)) ** 2n) / (10n ** BigInt(token.decimals))
+						: (10n ** (18n)) / ((sqrtPriceX96 / (2n ** 96n)) ** 2n)
+
+					const reserve0 = BigInt(multicallReturnData[index * 2 + 3].returnData)
```
KillariDev, line 87: can we parse this with funtypes instead of casting?

app/ts/simulation/priceEstimator.ts
```diff
@@ -21,99 +20,92 @@ export class PriceEstimator {
...
+					const reserve1 = BigInt(multicallReturnData[index * 2 + 4].returnData)
+
+					return [...prices, { price, k: reserve0 * reserve1 }]
```
KillariDev, line 90: can we have a better name than `k` for this?

app/ts/simulation/priceEstimator.ts
```diff
@@ -21,99 +20,92 @@ export class PriceEstimator {
...
+				}
+
+				return prices
+			}, [] as { price: bigint, k: bigint }[])
```
KillariDev, line 94: What does the `price: bigint` contain? What if the price of token A is less than a wei? I feel if we have prices they should be numbers and not bigints, unless you represent them with a fraction (A: bigint / B: bigint)

app/ts/simulation/priceEstimator.ts
```diff
@@ -21,99 +20,92 @@ export class PriceEstimator {
...
+
+			// Use pool with most TVL
+			const price = prices.length > 0 ? prices.reduce((highestLiq, p) => p.k > highestLiq.k ? p : highestLiq).price : undefined
+
+			if (price) tokenPrices.push({
```
KillariDev, line 99: if the price is zero, we are not pushing that here. I feel this should check if its undefined instead

app/ts/simulation/priceEstimator.ts
```diff
@@ -21,99 +20,92 @@ export class PriceEstimator {
...
 	}
 }
 
-export function getTokenPrice(ethInOutAmounts: readonly bigint[], tokenDecimals: bigint, precision: number) {
-	return Number(ethInOutAmounts[1] * 10n ** BigInt(precision) / ethInOutAmounts[0]) / Number(10n ** BigInt(precision)* 10n ** tokenDecimals / 10n ** 18n)
-}
-
 export function getTokenAmountsWorth(tokenAmount: bigint, tokenPriceEstimate: TokenPriceEstimate) {
-	return (tokenPriceEstimate.inOutAmount[0] * tokenAmount / tokenPriceEstimate.inOutAmount[1])
+	return (tokenPriceEstimate.price * tokenAmount) / (10n ** tokenPriceEstimate.token.decimals)
```
KillariDev, line 110: I feel this is wrong if the price of a token is small enough? as the price would just be 0

app/ts/utils/constants.ts
```diff
@@ -1,13 +1,14 @@
...
 // common contract addresses
-export const UNISWAP_FACTORY_ADDRESS = 0x5C69bEe701ef814a2B6a3EDD4B1652CB9cc5aA6fn
+export const UNISWAP_V2_FACTORY_ADDRESS = 0x5C69bEe701ef814a2B6a3EDD4B1652CB9cc5aA6fn
 export const SUSHISWAP_FACTORY_ADDRESS = 0xC0AEe478e3658e2610c5F7A4A2E1777cE9e4f2Acn
 export const UNISWAP_V2_ROUTER_ADDRESS = 0x7a250d5630B4cF539739dF2C5dAcb4c659F2488Dn
 export const SUSHISWAP_V2_ROUTER_ADDRESS = 0xd9e1cE17f2641f24aE83637ab66a2cca9C378B9Fn
 export const UNISWAP_V3_FACTORY_ADDRESS = 0x1F98431c8aD98523631AE4a59f267346ea31F984n
 export const UNISWAP_V3_ROUTER = 0xE592427A0AEce92De3Edee1F18E0157C05861564n
 export const UNISWAP_V3_NFT_ROUTER = 0xC36442b4a4522E871399CD717aBDD847Ab11FE88n
+export const MULTICALL3 = 0xcA11bde05977b3631167028862bE2a173976CA11n
```
KillariDev, line 11: can we add a source where this comes from? (we should add that to the other contracts too, but shouldn't be part of this pr)

app/ts/utils/uniswap.ts
```diff
@@ -0,0 +1,145 @@
...
+	const [token0, token1] = tokenA < tokenB ? [addressString(tokenA), addressString(tokenB)] : [addressString(tokenB), addressString(tokenA)]
+	const abi = new AbiCoder()
+
+	const v2 = getCreate2Address(
+		addressString(UNISWAP_V2_FACTORY_ADDRESS),
+		keccak256(solidityPacked(['address', 'address'], [token0, token1])),
+		'0x96e8ac4277198ff8b6f785478aa9a39f403cb768dd02cbee326c3e7da348845f'
```
KillariDev, line 30: I feel these should be consts with a name that explain what they are

app/ts/utils/uniswap.ts
```diff
@@ -0,0 +1,145 @@
...
+			'0xe34f199b19b2b4f47f68442619d555527d244f78a3297ea89325f843f87b8b54'
+		),
+		'10000': getCreate2Address(
+			addressString(UNISWAP_V3_FACTORY_ADDRESS),
+			keccak256(abi.encode(['address', 'address', 'uint24'], [token0, token1, 10000])),
+			'0xe34f199b19b2b4f47f68442619d555527d244f78a3297ea89325f843f87b8b54'
+		)
+	}
+
+	return {
```
KillariDev, line 56: make `UniswapPools` as a funtype, so we don't need to repeat the parsing here, but parse the whole thing at the same time

app/ts/utils/uniswap.ts
```diff
@@ -0,0 +1,145 @@
...
+		token0IsQuote: BigInt(token0) === tokenA,
```
KillariDev, line 57: hmm why do we need to cast to bigint?

KillariDev (changes requested):
Improve some clarity (in comments). I feel its wrong to represent price as a bigint, as tokens are often worth less than the same amount of eth. Does this also work if any of the pools are missing?

app/ts/simulation/priceEstimator.ts
0xjimmy: There are 5 pool variants, that have calls + then balanceOf calls to query v3 pool balances, so calls after index 4 (5th and last pool) are balanceOf calls.

app/ts/simulation/priceEstimator.ts
0xjimmy: Might be a good idea to use a ratio.
For now the price is amount of WEI per 1 whole token, which is still a good amount of precision.
For example it right now the quote price of PEPE is 801242161 WEI per token.

app/ts/utils/uniswap.ts
```diff
@@ -0,0 +1,122 @@
...
+export function calculateUniswapLikePools(token: EthereumAddress, quoteToken: EthereumAddress, chainString: string): UniswapPools | undefined {
+	const [token0, token1] = token < quoteToken ? [addressString(token), addressString(quoteToken)] : [addressString(quoteToken), addressString(token)]
+	const abi = new AbiCoder()
+
+	if (!isSupportedChain(chainString)) return undefined
+
+	const v2Pools = CHAINS[chainString].priceSources.uniswapV2like.map(({ factory, initCodeHash }) => EthereumAddress.parse(getCreate2Address(addressString(factory), keccak256(solidityPacked(['address', 'address'], [token0, token1])), initCodeHash)))
+
+	const v3Pools = CHAINS[chainString].priceSources.uniswapV3Like.map(({ factory, initCodeHash }) => [
+		getCreate2Address(addressString(factory), keccak256(abi.encode(['address', 'address', 'uint24'], [token0, token1, 100])), initCodeHash),
```
KillariDev, line 29: could create helper function for `getCreate2Address` that would take bigint instead of string

0xjimmy:
https://github.com/DarkFlorist/TheInterceptor/pull/566/files#diff-3c149a3d286ea025bf3f159322a627e1ae9eff10cf1049c54da115b6834dff9cR91-R95
`Ether` component needs to be replaced with `Token` component inside the `TokenPrice`.
NetworkPrice config object probably also needs to be updated to contain more metadata needed to pass to the `Token` component.

app/ts/background/settings.ts
```diff
@@ -49,7 +79,7 @@ export const defaultRpcs: RpcEntries = [
 		currencyTicker: 'SEETH',
 		primary: true,
 		minimized: true,
-		weth: 0x105083929bf9bb22c26cb1777ec92661170d4285n,
+		weth: 0x105083929bf9bb22c26cb1777ec92661170d4285n
```
KillariDev, line 82: commas could be added here. For some reason github doesn't allow me to edit them directly

KillariDev (approved):

KillariDev:
> https://github.com/DarkFlorist/TheInterceptor/pull/566/files#diff-3c149a3d286ea025bf3f159322a627e1ae9eff10cf1049c54da115b6834dff9cR91-R95 `Ether` component needs to be replaced with `Token` component inside the `TokenPrice`. NetworkPrice config object probably also needs to be updated to contain more metadata needed to pass to the `Token` component.

add issue about this

app/ts/background/settings.ts
MicahZoltu:
```suggestion
		weth: 0x105083929bf9bb22c26cb1777ec92661170d4285n,
```

PR #570, opened 2023-07-10
Block time warning

KillariDev:
image

Fix https://github.com/DarkFlorist/TheInterceptor/issues/397

### DIFF
--- a/app/ts/background/iconHandler.ts
+++ b/app/ts/background/iconHandler.ts
@@ -7,8 +7,8 @@ import { imageToUri } from '../utils/imageToUri.js'
 import { Future } from '../utils/future.js'
 import { WebsiteSocket, WebsiteTabConnections } from '../utils/user-interface-types.js'
 import { getSettings } from './settings.js'
-import { getIsConnected, getSignerName, updateTabState } from './storageVariables.js'
-import { TabIcon, TabState } from '../utils/interceptor-messages.js'
+import { getRpcConnectionStatus, getSignerName, updateTabState } from './storageVariables.js'
+import { RpcConnectionStatus, TabIcon, TabState } from '../utils/interceptor-messages.js'
 import { getLastKnownCurrentTabId } from './popupMessageHandlers.js'
 
 async function setInterceptorIcon(websiteTabConnections: WebsiteTabConnections, tabId: number, icon: TabIcon, iconReason: string) {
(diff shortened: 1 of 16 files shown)


app/ts/background/iconHandler.ts
```diff
@@ -62,8 +62,13 @@ export async function updateExtensionIcon(websiteTabConnections: WebsiteTabConne
 	return setInterceptorIcon(websiteTabConnections, socket.tabId, ICON_SIGNING, `The Interceptor forwards your transactions to ${ getPrettySignerName(await getSignerName()) } once sent.`)
 }
 
+export function noNewBlockForOverTwoMins(connectionStatus: RpcConnectionStatus) {
+	return connectionStatus && connectionStatus.latestBlock && (connectionStatus.lastConnnectionAttempt.getTime() - connectionStatus.latestBlock.timestamp.getTime()) > 2 * 60 * 1000
```
MicahZoltu, line 66: For Ethereum Mainnet, we could be quite a bit more aggressive than this I think.  1 block missing is rare (1% last I heard), 2 blocks in a row missing is even more rare (0.1%), 3 blocks 0.01%, etc.  If this error appears for 12 seconds once every 33 hours that feels fine to me.  We could bump that up 4 blocks in a row (60 seconds) missing to reach once every ~2 weeks (and only for 12 seconds).

app/ts/components/pages/Home.tsx
```diff
@@ -188,13 +189,35 @@ function SimulationResults(param: SimulationStateParam) {
...
+		{ rpcConnectionStatus.isConnected === false ?
+			<div style = 'margin: 10px; background-color: var(--bg-color);'>
+				<Error warning = { true } text = { <>Unable to connect to { rpcConnectionStatus.rpcNetwork.name }. Retrying in <SomeTimeAgo priorTimestamp = { priorDate } countBackwards = { true }/>.</> }/>
+			</div>
+		: <></> }
+
+		{ rpcConnectionStatus.latestBlock !== undefined && noNewBlockForOverTwoMins(rpcConnectionStatus) ?
+			<div style = 'margin: 10px; background-color: var(--bg-color);'>
+				<Error warning = { true } text = { <>The connected RPC ({ rpcConnectionStatus.rpcNetwork.name }) seem to be stuck at block { rpcConnectionStatus.latestBlock.number }. Retrying in <SomeTimeAgo priorTimestamp = { priorDate } countBackwards = { true }/>.</> }/>
+			</div>
```
MicahZoltu, line 216: We should include `rpcConnectionStatus.latestBlock.timestamp` somewhere in the warning, so the user knows just how old the block is.  In fact, I would remove the number and replace it with timestamp I think.

MicahZoltu (approved):

app/ts/background/iconHandler.ts
KillariDev: couple blocks missing can happen by our node lagging a bit behind, I wouldn't want to alert all users if the node lags a bit. I feel this error should happen only when its quite clear the node appears to be frozen

app/ts/components/pages/Home.tsx
KillariDev: image
I feel the block number is useful too, as if fork happens and somene forgots to update, the block number would somewhat explain why

MicahZoltu: Recommend changing "occured on" to "produced at" for a bit more clarity.

Issue #567, opened 2023-07-05
Lunaria simulation not adding to the simulation stack #testing


jubalm:
> Hi All, these are some of the intermittent issues I encountered during the use of interceptor which has been experienced by others as well. These are all in Google Chrome. Not sure if this would be the correct channel

**Lunaria simulation not adding to the simulation stack**
1. On Goerli testnet, simulated a transaction of claiming token from Uniswap V3 NFT
2. Simulation stack properly displayed transaction of claiming token A and token B landing on the compromised wallet
3. Upon simulation on Lunaria, token A was simulated properly and was added to the stack to be transferred to recipient.
4. Same was done for token B, confirmation dialog appears and simulation appears to be fine and successful (buttons and prompts). However, the simulation doesn't appear in the simulation stack as if nothing was done

Workaround: Instead of just confirming/simulating as the confirmation dialog appears, waited 10-20 seconds before confirming and simulation was recorded to the stack as expected.

**Multiple wallets conflicting on bouquet upon importing payload from the interceptor**
1. Wallet A and Wallet B is being used in the same browser
2. Initially tested with wallet A, connected to bouquet and initiated simulations without problem
3. Switched to wallet B, both in interceptor and bouquet
4. Simulated transaction with interceptor using wallet B with no problem. Proper address appears in the simulation stack ( in the extension)
5. Upon importing payload to bouquet, address being reflected on the stack on the bouquet is that of Wallet A instead of wallet B. Interceptor simulation stack still shows wallet B. Sometimes, pressing on importing payload again switches it to the correct wallet, wallet B

Workaround: Started with a clean browser were only wallet B has connected

MicahZoltu:
We should split this into two separate bugs.

KillariDev:
these bugs are probably fixed in master, but let's keep this open if this still happens

PR #572, opened 2023-07-12
Instead of using requestID as unique identifier across connections (its not), add uniqueRequestIdentifier that is unique

KillariDev:
One possible fix for the first issue of https://github.com/DarkFlorist/TheInterceptor/issues/567 . However, I think its unlikely it was caused by this, so I continue looking for that.

### DIFF
--- a/app/ts/background/messageSending.ts
+++ b/app/ts/background/messageSending.ts
@@ -1,0 +1,55 @@
+import { InterceptedRequestForward, InterceptorMessageToInpage, SubscriptionReplyOrCallBack } from "../utils/interceptor-messages.js"
+import { WebsiteSocket, WebsiteTabConnections } from "../utils/user-interface-types.js"
+import { websiteSocketToString } from "./backgroundUtils.js"
+
+function postMessageToPortIfConnected(port: browser.runtime.Port, message: InterceptorMessageToInpage) {
+	try {
+		port.postMessage(InterceptorMessageToInpage.serialize(message) as Object)
+	} catch (error) {
+		if (error instanceof Error) {
+			if (error.message?.includes('Attempting to use a disconnected port object')) {
+				return
+			}
+			if (error.message?.includes('Could not establish connection. Receiving end does not exist')) {
+				return
+			}
+		}
+		throw error
+	}
+}
+export function postMessageIfStillConnected(websiteTabConnections: WebsiteTabConnections, socket: WebsiteSocket, message: InterceptorMessageToInpage) {
+	const tabConnection = websiteTabConnections.get(socket.tabId)
+	const identifier = websiteSocketToString(socket)
+	if (tabConnection === undefined) return false
+	for (const [socketAsString, connection] of Object.entries(tabConnection.connections)) {
+		if (socketAsString !== identifier) continue
+		postMessageToPortIfConnected(connection.port, message)
+	}
+	return true
+}
+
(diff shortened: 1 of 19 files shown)


app/ts/background/messageSending.ts
```diff
@@ -0,0 +1,55 @@
...
+			}
+		}
+		throw error
+	}
+}
+export function postMessageIfStillConnected(websiteTabConnections: WebsiteTabConnections, socket: WebsiteSocket, message: InterceptorMessageToInpage) {
+	const tabConnection = websiteTabConnections.get(socket.tabId)
+	const identifier = websiteSocketToString(socket)
+	if (tabConnection === undefined) return false
+	for (const [socketAsString, connection] of Object.entries(tabConnection.connections)) {
```
MicahZoltu, line 24: This is dangerous as the `Object.entries` return type is not sound.  Take the following example:
```ts
type Fruit = { type: 'apple' | 'banana', color: 'red' | 'yellow' }
function eat(fruit: Fruit) {
	const entries = Object.entries(fruit)
	for (const [ key, value ] of entries) {
		// compiletime: typeof key === 'type' | 'color'
		// runtime: typeof key === 'type' | 'color' | 'seed'
	}
}
const apple = { type: 'apple', color: 'red', seeds: true } as const
eat(apple)
```
Generally speaking, you should avoid using `Object.entries` unless you are *absolutely certain* that there is no way (now or in the future) that extra properties can't be added to by someone in the future, because the type checker won't validate it correctly.  Since `tabConnection.connections` is coming from somewhere else, I strongly recommend not using `Object.entries` here and instead iterating over opaque keys and narrowing inside the loop.

app/ts/background/messageSending.ts
```diff
@@ -0,0 +1,55 @@
...
+		postMessageToPortIfConnected(connection.port, message)
+	}
+	return true
+}
+
+export function replyToInterceptedRequest(websiteTabConnections: WebsiteTabConnections, message: InterceptedRequestForward) {
+	const tabConnection = websiteTabConnections.get(message.uniqueRequestIdentifier.requestSocket.tabId)
+	const identifier = websiteSocketToString(message.uniqueRequestIdentifier.requestSocket)
+	if (tabConnection === undefined) return false
+	for (const [socketAsString, connection] of Object.entries(tabConnection.connections)) {
```
MicahZoltu, line 35: Same comment as above.

app/ts/background/messageSending.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export function sendSubscriptionReplyOrCallBackToPort(port: browser.runtime.Port, message: SubscriptionReplyOrCallBack) {
+	postMessageToPortIfConnected(port, {...message, interceptorApproved: true })
+}
+
+export function sendSubscriptionReplyOrCallBack(websiteTabConnections: WebsiteTabConnections, socket: WebsiteSocket, message: SubscriptionReplyOrCallBack) {
+	const tabConnection = websiteTabConnections.get(socket.tabId)
+	const identifier = websiteSocketToString(socket)
+	if (tabConnection === undefined) return false
+	for (const [socketAsString, connection] of Object.entries(tabConnection.connections)) {
```
MicahZoltu, line 50: Same comment as above.

MicahZoltu (approved):
Strongly recommend addressing the `Object.entries` issue as it is a dangerous pattern, but I'm trying to take a more hands-off approach so will approve this even though I think it is a bad idea.

PR #577, opened 2023-07-12
blur simulation results when they are being updated

KillariDev:
image

### DIFF
--- a/app/css/interceptor.css
+++ b/app/css/interceptor.css
@@ -707,3 +707,7 @@ svg.spinner > circle {
 	padding-bottom: 10px;
 	padding-top: 10px;
 }
+
+.blur {
+	filter: blur(3px);
+}
--- a/app/ts/background/background.ts
+++ b/app/ts/background/background.ts
@@ -1,44 +1,30 @@
-import { ConfirmTransactionTransactionSingleVisualization, InpageScriptRequest, PopupMessage, RPCReply, Settings, TabState } from '../utils/interceptor-messages.js'
+import { ConfirmTransactionTransactionSingleVisualization, PopupMessage, RPCReply, Settings } from '../utils/interceptor-messages.js'
 import 'webextension-polyfill'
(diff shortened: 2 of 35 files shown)


app/ts/background/background.ts
jubalm, line 1: i just noticed commented lines on #583 (missing methods) which could maybe an issue so you can track later?

jubalm (approved):

app/css/interceptor.css
```diff
@@ -707,3 +707,7 @@ svg.spinner > circle {
 	padding-bottom: 10px;
 	padding-top: 10px;
 }
+
+.blur {
+	filter: blur(3px);
```
MicahZoltu, line 712: Can we make this a bit less blurry so you can still mostly make out the values behind the blur?  In most cases, when re-simulating the results won't change and it will be annoying if you are trying to read it and every 12 seconds it blurs and you have to pause for a second to read.

MicahZoltu (commented):
I'm not particularly a fan of this, though also not openly opposed to it.  In many cases, the user may be taking time to review the contents of the simulation and having it blur out every 12 seconds (even when no changes are happening) is somewhat annoying.

app/css/interceptor.css
KillariDev: one option is to make it blur only on clear situations when we know the simulation results are wrong. Eg, when you change account, change chain, add new transaction

MicahZoltu: I would be far more in favor of that.

KillariDev: but not like that either? My problem is that the simulation takes a bit time, and also the more stuff we add to it, the longer it takes, and we need to somehow display to users that it waits for simulation

MicahZoltu: I am OK with a blur if we are reasonably confident that the results are going to change.  The thing I'm minor 👎 on is having it blur every 12 seconds when we re-simulate just because a new block arrived.

PR #578, opened 2023-07-13
adds workaround for constraint checking

KillariDev:
workaround for https://github.com/ForbesLindesay/funtypes/issues/62

### DIFF
--- a/app/ts/background/windows/confirmTransaction.ts
+++ b/app/ts/background/windows/confirmTransaction.ts
@@ -118,7 +118,7 @@ export async function openConfirmTransactionDialog(
 				simulationResults: await refreshSimulationPromise,
 			})
 			await updateConfirmTransactionViewWithPendingTransaction()
-			if (justAddToPending) sendPopupMessageToOpenWindows({ method: 'popup_confirm_transaction_dialog_pending_changed', data: (await getPendingTransactions()).map((p) => p.simulationResults) })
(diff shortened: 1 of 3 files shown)


app/ts/background/windows/confirmTransaction.ts
```diff
@@ -118,7 +118,7 @@ export async function openConfirmTransactionDialog(
 				simulationResults: await refreshSimulationPromise,
 			})
 			await updateConfirmTransactionViewWithPendingTransaction()
-			if (justAddToPending) sendPopupMessageToOpenWindows({ method: 'popup_confirm_transaction_dialog_pending_changed', data: (await getPendingTransactions()).map((p) => p.simulationResults) })
+			if (justAddToPending) await sendPopupMessageToOpenWindows({ method: 'popup_confirm_transaction_dialog_pending_changed', data: (await getPendingTransactions()).map((p) => p.simulationResults) })
```
MicahZoltu, line 121: Unrelated bug fix?

MicahZoltu (approved):
Feels like this should be two PRs, or at least both should be mentioned in the PR description.  That being said, probably not worth the effort to split them now since this is a pretty small changeset overall.

app/ts/background/windows/confirmTransaction.ts
KillariDev: yaeh

PR #579, opened 2023-07-13
Executionspec 383 types and a wrapper

KillariDev:
- Adds support for geth normal client operation
- add execution spec 383 types
- refactors json-rpc's types to new file
- adds a wrapper of new multicall to old. It doesn't support eth balance changes thought, and some downstream breaks for that. I need to think about how I modify that

There's no UI enabled to enable new multicall client for now (intended as it doesn't fully work)


app/ts/simulation/services/EthereumClientService.ts
```diff
@@ -192,8 +195,65 @@ export class EthereumClientService {
...
 		return MulticallResponse.parse(unvalidatedResult)
 	}
+
+	public readonly executionSpec383MultiCall = async (calls: readonly BlockCalls[], blockTag: EthereumBlockTag) => {
+		const call = { method: 'eth_multicallV1', params: [calls, blockTag] } as const
+		console.log(calls)
+		console.log(JSON.stringify(ExecutionSpec383MultiCallParams.serialize(call)))
+		const unvalidatedResult = await this.requestHandler.jsonRpcRequest(call)
+		console.log([0, calls, blockTag])
+		console.log(unvalidatedResult)
```
MicahZoltu, line 215: Recommend adding a TODO to remove these log lines, since they appear to just for debugging.  this way you can just search all files for `TODO` to easily find them, rather than having to dig around trying to find all of the stuff that needs to be removed.

It may be valuable to have a comment prefix specific to the multicall stuff like `// MULTICALL TODO: ...` so you can easily find it all.

MicahZoltu (approved):

PR #594, opened 2023-08-03
Don't prompt access request again for the same request

KillariDev:
The ens site is calling eth_requestAccounts in a loop all the time and this caused interceptor to behave badly. Now if dapp requests this account, we throw `Access request pending already` error. 

There was also another race condition bug caused by this, when you approve the request, it takes couple frames for access rights to propagate and during this time the dapp requested access again while the accesses were already.

And there's a third bug, I did not know tabs `id` and `windowId` are separate things, so now I am just using `id` to track things, instead of both mixed. 

fix https://github.com/DarkFlorist/TheInterceptor/issues/591

### DIFF
--- a/app/ts/background/storageVariables.ts
+++ b/app/ts/background/storageVariables.ts
@@ -157,9 +157,10 @@ export async function updateTabState(tabId: number, updateFunc: (prevState: TabS
 const pendingAccessRequestsSemaphore = new Semaphore(1)
 export async function updatePendingAccessRequests(updateFunc: (prevState: PendingAccessRequestArray) => Promise<PendingAccessRequestArray>) {
 	return await pendingAccessRequestsSemaphore.execute(async () => {
-		const pendingAccessRequests = await updateFunc(PendingAccessRequestArray.parse(await browserStorageLocalSingleGetWithDefault('pendingInterceptorAccessRequests', [])))
+		const previous = PendingAccessRequestArray.parse(await browserStorageLocalSingleGetWithDefault('pendingInterceptorAccessRequests', []))
+		const pendingAccessRequests = await updateFunc(previous)
 		await browserStorageLocalSet('pendingInterceptorAccessRequests', PendingAccessRequestArray.serialize(pendingAccessRequests) as string)
-		return pendingAccessRequests
+		return { previous: previous, current: pendingAccessRequests }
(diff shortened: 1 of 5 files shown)


app/ts/background/storageVariables.ts
```diff
@@ -157,9 +157,10 @@ export async function updateTabState(tabId: number, updateFunc: (prevState: TabS
 const pendingAccessRequestsSemaphore = new Semaphore(1)
 export async function updatePendingAccessRequests(updateFunc: (prevState: PendingAccessRequestArray) => Promise<PendingAccessRequestArray>) {
 	return await pendingAccessRequestsSemaphore.execute(async () => {
-		const pendingAccessRequests = await updateFunc(PendingAccessRequestArray.parse(await browserStorageLocalSingleGetWithDefault('pendingInterceptorAccessRequests', [])))
+		const previous = PendingAccessRequestArray.parse(await browserStorageLocalSingleGetWithDefault('pendingInterceptorAccessRequests', []))
+		const pendingAccessRequests = await updateFunc(previous)
 		await browserStorageLocalSet('pendingInterceptorAccessRequests', PendingAccessRequestArray.serialize(pendingAccessRequests) as string)
```
jubalm, line 162: Could use an assertion to ensure serialize returns string

app/ts/components/ui-utils.tsx
```diff
@@ -100,13 +100,17 @@ export async function openPopupOrTab(createData: browser.windows._CreateCreateDa
 	return { type: 'popup', windowOrTab: window }
 }
 
+export async function browserTabsQueryById(id: number) {
+	return (await browser.tabs.query({})).find((x) => x.id === id)
```
jubalm, line 104: `query` function could use a default here so it can just be called without the empty object

jubalm (approved):

app/ts/background/storageVariables.ts
KillariDev: Good catch! The type here is wrong, I need to refactor these by making browserStorageLocalSet to take an object that includes the proper type (PendingAccessRequestArray) with the key "pendingInterceptorAccessRequests" and then performing the serialization just before storing then. Similar refactor that I did to json rpc method return values

Issue #569, opened 2023-07-05
funtypes preventing wallet_addEthereumChain from being passed through to metamask


epheph:

epheph:
https://www.flashbotsrpc.com/

for unknown methods, we should pass along to the wallet.

```
Uncaught (in promise) EthereumJsonRpcError: Unable to assign {method: "wallet_addEthereumChain" ... } to { readonly method: "eth_getBlockByNumber"; readonly params: [ParsedValue<string> | ParsedValue<string> | "latest" | "pending", boolean]; } | { readonly method: "eth_getBalance"; readonly params: [ParsedValue<string>, ParsedValue<string> | ParsedValue<string> | "latest" | "pending"]; } | { readonly method: "eth_estimateGas"; readonly params: [{ readonly from?: ParsedValue<string>; readonly gas?: ParsedValue<string>; readonly value?: ParsedValue<string>; readonly to?: ParsedValue<string> | null; readonly gasPrice?: ParsedValue<string>; readonly maxPriorityFeePerGas
```

MicahZoltu:
In simulation mode, unknown methods should result in an error, but it should be more clear to the user than a funtype error.  e.g., "`wallet_addEthereumChain` is not supported by Interceptor."

In signing mode, I agree with @epheph that we should just forward the request to the underlying wallet.

KillariDev:
hmm, I don't feel we should just blindly forward everything 🤔 . If you have a typo in method, it sounds silly that its just passed to wallet. Or if there's a new signing method that we don't support, you could bypass interceptor using that.

MicahZoltu:
There will almost certainly be times when a wallet implements some new feature that we haven't integrated yet, and we don't want the solution for the user to be "disable Interceptor".  Perhaps we can let the user know about the unexpected request and ask the user if they want to pass the request on to the wallet (with a remember me button)?  We can give them a sufficient warning saying that we don't recommend clicking yes.

PR #598, opened 2023-08-04
Support for unsupported methods

KillariDev:
- forward `wallet_addEthereumChain` to wallet
- forward all unknown json rpc methods to a wallet that we are not supporting. This is done so that we are checking that the method is not one we support, and only then forward. If the dapp is sending a request method that we support, but with invalid parameters, we will block it and throw error

fix https://github.com/DarkFlorist/TheInterceptor/issues/569

### DIFF
--- a/app/ts/background/background.ts
+++ b/app/ts/background/background.ts
@@ -24,7 +24,7 @@ import { updateChainChangeViewWithPendingRequest } from './windows/changeChain.j
 import { updatePendingPersonalSignViewWithPendingRequests } from './windows/personalSign.js'
 import { InterceptedRequest, UniqueRequestIdentifier } from '../utils/requests.js'
 import { replyToInterceptedRequest } from './messageSending.js'
-import { EthereumJsonRpcRequest, SendRawTransaction, SendTransactionParams } from '../utils/JsonRpc-types.js'
+import { EthGetStorageAtParams, EthereumJsonRpcRequest, SendRawTransaction, SendTransactionParams, SupportedEthereumJsonRpcRequestMethods, WalletAddEthereumChain } from '../utils/JsonRpc-types.js'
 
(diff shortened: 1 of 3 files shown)


app/ts/background/background.ts
```diff
@@ -192,21 +192,36 @@ async function handleRPCRequest(
...
 		return {
 			method: request.method,
 			error: {
-				message: maybeParsedRequest.fullError === undefined ? 'Failed to parse RPC request ' : maybeParsedRequest.fullError.toString(),
+				message: `Failed to parse RPC request: ${ JSON.stringify(request) }`,
+				data: maybeParsedRequest.fullError === undefined ? 'Failed to parse RPC request' : maybeParsedRequest.fullError.toString(),
 				code: METAMASK_ERROR_FAILED_TO_PARSE_REQUEST,
 			}
 		}
 	}
```
MicahZoltu, line 224: This can be simplified with some conditional inversion and combining two of the return paths.
```suggestion
if (maybeParsedRequest.success === false) {
	console.log(request)
	console.warn(maybeParsedRequest.fullError)
	const maybePartiallyParsedRequest = SupportedEthereumJsonRpcRequestMethods.safeParse(request)
	// the method is some method that we are not supporting, forward it to the wallet if signer is available
	if (maybePartiallyParsedRequest.success === false && forwardToSigner) return { forward: true as const, unknownMethod: true, ...request }
	return {
		method: request.method,
		error: {
			message: `Failed to parse RPC request: ${ JSON.stringify(request) }`,
			data: maybeParsedRequest.fullError === undefined ? 'Failed to parse RPC request' : maybeParsedRequest.fullError.toString(),
			code: METAMASK_ERROR_FAILED_TO_PARSE_REQUEST,
		}
	}
}
```

MicahZoltu (approved):

Issue #603, opened 2023-08-09
ERC1155 swap


KillariDev:
When swap is performed, currently ERC1155 tokens are not considered

PR #604, opened 2023-08-09
Add Erc1155 support

KillariDev:
image

Adds support for ERC1155. If you perform a swap with ERC1155, it's not identified as a swap currently, all the other cases should be handled. (added issue about this: https://github.com/DarkFlorist/TheInterceptor/issues/603 )


app/ts/AddressBook.tsx
```diff
@@ -346,8 +346,8 @@ export function AddressBook() {
 		switch(filter) {
 			case 'My Active Addresses': return setAddingNewAddressType({ addingAddress: true, type: 'addressInfo' })
 			case 'My Contacts': return setAddingNewAddressType({ addingAddress: true, type: 'contact' })
-			case 'Erc20Tokens': return setAddingNewAddressType({ addingAddress: true, type: 'Erc20Token' })
-			case 'Non Fungible Tokens': return setAddingNewAddressType({ addingAddress: true, type: 'NFT' })
+			case 'Erc20Tokens': return setAddingNewAddressType({ addingAddress: true, type: 'ERC20' })
+			case 'Non Fungible Tokens': return setAddingNewAddressType({ addingAddress: true, type: 'ERC721' })
```
0xjimmy, line 350: So no ERC1155 in address book?

app/ts/components/formVisualizerResults.ts
```diff
@@ -25,25 +25,44 @@ export function formSimulatedAndVisualizedTransaction(simState: SimulationState,
...
+				return {
+					...change,
+					from: fromEntry,
+					to: toEntry,
+					token: tokenEntry,
+				}
+			}
+			console.warn('unknown token in token results:')
+			console.log(change)
+			console.log(tokenEntry)
```
0xjimmy, line 64: This intended to be in prod?

app/ts/components/simulationExplaining/customExplainers/OpenSeaOrder.tsx
```diff
@@ -16,8 +16,8 @@ type VisualizeOpenSeaAssetParams = {
 }
 
 function ValueField({ orderOrConsideration }: { orderOrConsideration: SeaPortSingleOfferWithAddressBookEntries | SeaPortSingleConsiderationWithAddressBookEntries }) {
-	if (orderOrConsideration.itemType === 'ERC721' || orderOrConsideration.itemType === 'ERC1155') {
-		return <Token721AmountField { ...orderOrConsideration.token } tokenId = { orderOrConsideration.identifierOrCriteria } type = { 'NFT' } style = { tokenStyle }/>
+	if (orderOrConsideration.itemType === 'ERC721') {
+		return <></>
```
0xjimmy, line 20:
```suggestion
		return null
```

0xjimmy (approved):

app/ts/AddressBook.tsx
0xjimmy: Never mind, I see the difference in the NFT entries

app/ts/components/formVisualizerResults.ts
KillariDev: yeah, it can still happen if we have identified token as erc20 and then for some reason it sends erc721 logs for example, so I want to log those cases if they happen, but they shouldn't be happen with normal operation. I'll add issue to handle them better

PR #605, opened 2023-08-09
If simulation fails on startup, eg there's no internet, throw error but don't block the view from rendering

KillariDev:


0xjimmy:
image
Currently it doesn't render anything if I cut network, but will start working again when network is back

KillariDev:
> image Currently it doesn't render anything if I cut network, but will start working again when network is back

hmm, I need to take a look what could cause that. Thank you for noticing

jubalm (approved):

PR #609, opened 2023-08-11
Transaction sending rework

KillariDev:
- partially fixes https://github.com/DarkFlorist/TheInterceptor/issues/602 in a way that if when a new block is confirmed, the nonce is checked. However, we are not still counting pending transactions, only confirmed ones
- makes sure that error's data object is propagated to the dapp if available
-fix https://github.com/DarkFlorist/TheInterceptor/issues/606 : 
image

### DIFF
--- a/app/ts/background/background.ts
+++ b/app/ts/background/background.ts
@@ -3,9 +3,9 @@ import 'webextension-polyfill'
 import { Simulator } from '../simulation/simulator.js'
 import { getEthDonator, getSignerName, getSimulationResults, updateSimulationResults } from './storageVariables.js'
 import { changeSimulationMode, getSettings, getMakeMeRich } from './settings.js'
-import { blockNumber, call, chainId, estimateGas, gasPrice, getAccounts, getBalance, getBlockByNumber, getCode, getLogs, getPermissions, getSimulationStack, getTransactionByHash, getTransactionCount, getTransactionReceipt, personalSign, sendRawTransaction, sendTransaction, subscribe, switchEthereumChain, unsubscribe } from './simulationModeHanders.js'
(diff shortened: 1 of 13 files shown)


app/ts/background/popupMessageHandlers.ts
```diff
@@ -177,29 +177,39 @@ export async function refreshPopupConfirmTransactionMetadata(ethereumClientServi
...
+	if (first.simulationResults === undefined || first.simulationResults.statusCode !== 'success') return
 	return await sendPopupMessageToOpenWindows({
 		method: 'popup_update_confirm_transaction_dialog',
 		data: [{
-			statusCode: 'success',
-			data: {
-				...first.simulationResults.data,
-				simulatedAndVisualizedTransactions: formSimulatedAndVisualizedTransaction(first.simulationResults.data.simulationState, first.simulationResults.data.visualizerResults, addressBookEntries),
-				addressBookEntries,
+			...first,
```
jubalm, line 184: I may be mistaken but seems this may be a `Promise.all`?

jubalm (approved):

app/ts/background/background.ts
```diff
@@ -158,7 +158,7 @@ export async function refreshConfirmTransactionSimulation(
 }
 
 // returns true if simulation state was changed
-export async function getPrependTrasactions(ethereumClientService: EthereumClientService, settings: Settings, richMode: boolean) {
+export async function getPrependTrasactions(ethereumClientService: EthereumClientService, settings: Settings, richMode: boolean): Promise<WebsiteCreatedEthereumUnsignedTransaction[]> {
```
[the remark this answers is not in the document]
KillariDev: oh, I have never heard of `satisfies` keyword. I think I could find that useful in some places 🤔 

I often wants to specify the return type of function, as sometimes I am crafting the complex type and then its nice to see that I have constructed the type correctly as I get error otherwise

app/ts/background/popupMessageHandlers.ts
KillariDev: hmm. I don't understand what do you mean?

PR #633, opened 2023-08-23
prevent modification of interceptor and dark florist data

KillariDev:
fix https://github.com/DarkFlorist/TheInterceptor/issues/627

image


0xjimmy (approved):
Only minor grammar suggestion. "incomplete" is single word, names like ``InCompleteAddressBookEntry`` could be ``IncompleteAddressBookEntry`` instead.

PR #635, opened 2023-08-23
Duplicate name or symbol

KillariDev:
fixes https://github.com/DarkFlorist/TheInterceptor/issues/626

Do not allow user to add addresses with duplicate name or symbol in them

image


app/ts/background/medataSearch.ts
```diff
@@ -135,3 +135,29 @@ export async function getMetadataForAddressBookData(filter: GetAddressBookDataFi
 		maxDataLength: filtered.length,
 	}
 }
+
+export async function findEntryWithSymbolOrName(symbol: string | undefined, name: string | undefined, userAddressBook: UserAddressBook): Promise<AddressBookEntry | undefined> {
```
jubalm, line 139: matchers can use a helper here:
```ts
function likeStrings(a: string, b: string) {
	return a.toLowerCase() === b.toLowerCase()
}
```

jubalm (approved):

Issue #581, opened 2023-07-18
allow user to add unknown ??? token to addressbook by clicking it and accepting


KillariDev:

KillariDev:
add same button that we have wtih addresses

PR #642, opened 2023-08-28
Token renaming from transactions

KillariDev:
fix https://github.com/DarkFlorist/TheInterceptor/issues/581
Allows user to rename tokens when they appear

image


app/ts/utils/constants.ts
```diff
@@ -197,3 +197,5 @@ export const PRIMARY_COLOR = '#58a5b3'
 
 export const CHROME_NO_TAB_WITH_ID_ERROR = 'No tab with id'
 export const CANNOT_SIMULATE_OFF_LEGACY_BLOCK = 'Cannot simulate off a legacy block'
+
+export const BIG_FONT_SIZE = '28px'
```
jubalm, line 201: this looks like something that can be handled in CSS?

jubalm (approved):

PR #650, opened 2023-09-03
fix `eth_send`

KillariDev:
fix how eth_send is supposed to work. This makes peepeth work better, however, peepeth still uses filter methods that are not implemented, but peepeth works with metamask+interceptor in signing mode (not in simulation)


app/inpage/ts/inpage.ts
```diff
@@ -202,13 +204,19 @@ class InterceptorMessageListener {
...
+			if (window.ethereum === undefined) throw new Error('window.ethereum is missing')
+			switch (payload.method) {
+				case 'eth_coinbase': 
+				case 'eth_accounts': return { jsonrpc: '2.0', id: payload.id, result: window.ethereum.selectedAddress === undefined || window.ethereum.selectedAddress === null ? [] : [window.ethereum.selectedAddress] }
+				case 'net_version': return { jsonrpc: '2.0', id: payload.id, result: window.ethereum.networkVersion }
+				case 'eth_chainId': return { jsonrpc: '2.0', id: payload.id, result: window.ethereum.chainId }
+				default: throw new EthereumJsonRpcError(METAMASK_INVALID_METHOD_PARAMS, `Invalid method parameter for window.ethereum.send: ${ payload.method }`)
+			}
+		}
+		throw new EthereumJsonRpcError(METAMASK_METHOD_NOT_SUPPORTED, 'Method not supported (wnidow.ethereum.send).')
```
0xjimmy, line 219:
```suggestion
		throw new EthereumJsonRpcError(METAMASK_METHOD_NOT_SUPPORTED, 'Method not supported (window.ethereum.send).')
```

0xjimmy (approved):

PR #663, opened 2023-09-27
Ecrecover faking

KillariDev:
- fixes https://github.com/DarkFlorist/TheInterceptor/issues/662
- Allows user to fake permit2 messages (probably others as well, not fully tested)

image


app/ts/background/popupMessageHandlers.ts
```diff
@@ -333,7 +338,7 @@ export async function homeOpened(simulator: Simulator) {
 			tabId,
 			rpcEntries: await getRpcList(),
 		}
-	})
+	}) as PartiallyParsedPersonalSignRequest)
```
KillariDev, line 341: Figure out if there's a better way to do this

KillariDev:
sorry for the long commit, I had to do some refactoring to fit the new types. We need to figure out what to do with the big types better too

app/ts/background/popupMessageHandlers.ts
0xjimmy: I don't understand this, `PartiallyParsedPersonalSignRequest` and `UpdateHomePage` have different method's, and some of the expected `UpdateHomePage` values in the data.simulation don't match the  object.

Might need to use one type only and make it part of the ``MessageToPopup`` union?

0xjimmy (approved):

app/ts/background/popupMessageHandlers.ts
KillariDev: good catch, I fix that here: https://github.com/DarkFlorist/TheInterceptor/pull/669/files#diff-f3d3b02da582a58752f45484338f583264b08a4705fb534f7ef5c87666d24cc0R321

app/ts/simulation/services/EthereumJSONRpcRequestHandler.ts
```diff
@@ -28,7 +28,12 @@ export class EthereumJSONRpcRequestHandler {
 			},
 			body: JSON.stringify(requestBodyJson)
 		})
-		if (!response.ok) throw new FetchResponseError(response, requestBodyJson.id)
+		if (!response.ok) {
+			console.log('req failed')
+			console.log(response)
+			console.log(rpcRequest)
```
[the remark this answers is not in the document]
KillariDev: they are still useful for failing requests (they are only printed on failed requests), so I'll like to keep them in still

PR #699, opened 2023-10-18
catch error on new block and print it

KillariDev:
this was uncaught error otherwise, so lets print it and print it properly


app/ts/background/background-startup.ts
```diff
@@ -147,7 +148,11 @@ async function newBlockAttemptCallback(blockheader: EthereumBlockHeader, ethereu
 	await sendPopupMessageToOpenWindows({ method: 'popup_new_block_arrived', data: { rpcConnectionStatus } })
 	if (isNewBlock) {
 		const settings = await getSettings()
```
0xjimmy, line 150:
```ts
  await setRpcConnectionStatus(rpcConnectionStatus)
	await updateExtensionBadge()
	await sendPopupMessageToOpenWindows({ method: 'popup_new_block_arrived', data: { rpcConnectionStatus } })
	if (isNewBlock) {
		const settings = await getSettings()
```
If these throw its fine? Just the promises inside the try/catch is intended to to be a non-blocking error?

0xjimmy (approved):

app/ts/background/background-startup.ts
KillariDev: You are right its not fine. I just didn't expect them to throw, but so I ignored them. But we should also catch errors from them as well if they happen to throw. I fixed that

PR #703, opened 2023-10-18
catch tab id errors and ignore them

KillariDev:
fix errors: "Unchecked runtime.lastError: No tab with id: 942650812." when we try to reach for tab id that has been closed

### DIFF
--- a/app/ts/components/pages/ConfirmTransaction.tsx
+++ b/app/ts/components/pages/ConfirmTransaction.tsx
@@ -169,13 +169,18 @@ export function ConfirmTransaction() {
 			}
 			if (message.method === 'popup_confirm_transaction_dialog_pending_changed') {
 				updatePendingTransactions(message)
-				const currentWindowId = (await browser.windows.getCurrent()).id
-				const currentTabId = (await browser.tabs.getCurrent()).id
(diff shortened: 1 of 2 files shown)


app/ts/components/pages/ConfirmTransaction.tsx
```diff
@@ -169,13 +169,18 @@ export function ConfirmTransaction() {
...
-				const currentTabId = (await browser.tabs.getCurrent()).id
-				if (currentWindowId === undefined) throw new Error('could not get current window Id!')
-				if (currentTabId === undefined) throw new Error('could not get current tab Id!')
 				setPendingTransactionAddedNotification(true)
-				browser.windows.update(currentWindowId, { focused: true })
-				browser.tabs.update(currentTabId, { active: true })
+				try {
+					const currentWindowId = (await browser.windows.getCurrent()).id
+					const currentTabId = (await browser.tabs.getCurrent()).id
+					if (currentWindowId === undefined) throw new Error('could not get current window Id!')
```
jubalm, line 176: maybe throw early so that it doesn't need to await both tab and window id

```suggestion
					if (currentWindowId === undefined) throw new Error('could not get current window Id!')
					const currentTabId = (await browser.tabs.getCurrent()).id
```

jubalm (approved):

PR #704, opened 2023-10-18
clean up confirm transaction pending transaction management

KillariDev:


app/ts/background/popupMessageHandlers.ts
```diff
@@ -207,14 +205,12 @@ export async function refreshPopupConfirmTransactionMetadata(ethereumClientServi
 	})
 }
 
-export async function refreshPopupConfirmTransactionSimulation(simulator: Simulator, ethereumClientService: EthereumClientService, { data }: RefreshConfirmTransactionDialogSimulation) {
-	const transactionToSimulate = data.originalRequestParameters.method === 'eth_sendTransaction' ? await formEthSendTransaction(ethereumClientService, data.activeAddress, data.simulationMode, data.website, data.originalRequestParameters, data.created) : await formSendRawTransaction(ethereumClientService, data.originalRequestParameters, data.website, data.created)
+export async function refreshPopupConfirmTransactionSimulation(simulator: Simulator, ethereumClientService: EthereumClientService) {
 	const promises = await getPendingTransactions()
```
jubalm, line 209: just an alternative syntax so there's no need to splice the parts later to stitch back the new array

```ts
	const [firstTxn, ...remainingTxns] = await getPendingTransactions()
	// firstTxn should be same as promises[0]
	// ... to later be used like
	const newTransaction = [{ ...firstTxn, /* modified parts */ }, ...remainingTxns]
```

**Github won't allow selection of deleted lines preventing use of suggestion snippet here*

jubalm (approved):

Issue #691, opened 2023-10-08
look into governance systems that we could support


KillariDev:
- gitcoin
- gnosis safe
- others?

PR #712, opened 2023-10-25
When user is about to sign a vote on compound -like governance contracts, print simulation

KillariDev:
fix https://github.com/DarkFlorist/TheInterceptor/issues/691

When user is about to make compound governance vote, eg vote here: https://www.tally.xyz/gov/ondo-dao/proposal/9?chart=0
We'll identify this and make additional simulation for the vote outcome. Currently we just print the results to console, but we are not displaying them


app/ts/background/background.ts
```diff
@@ -50,13 +55,32 @@ async function updateMetadataForSimulation(simulationState: SimulationState, eth
...
+		if (simulatedTransaction.multicallResponse.statusCode !== 'success') return
+		const fourByte = get4Byte(simulatedTransaction.signedTransaction.input)
+		if (fourByte === undefined) return
+		const explanation = FourByteExplanations[fourByte]
+		if ((explanation !== 'Submit vote' && explanation !== 'Cast vote') || simulatedTransaction.multicallResponse.events.length !== 1) return
+		if (simulatedTransaction.signedTransaction.to === null) return
+		const governanceContractInterface = new Interface(CompoundGovernanceAbi)
+		const params = governanceContractInterface.decodeFunctionData(explanation === 'Submit vote' ? 'submitVote' : 'castVote', dataStringWith0xStart(simulatedTransaction.signedTransaction.input))
+		const contractExecutionResult = await simulateCompoundGovernanceExecution(ethereum, simulatedTransaction.signedTransaction.to, params[0])
+		console.log(contractExecutionResult)
```
jubalm, line 70: looks like some debugging code left here, also this silently fails?

jubalm (approved):

app/ts/background/background.ts
KillariDev: > looks like some debugging code left here, also this silently fails?

its not debugging code, its the output of the thing, thats all the function does, prints the results to console. Yeah, it should fail silently for some cases, but not others,I'll add issue about it

Issue #716, opened 2023-10-27
detect when its okay for gov simulaton to fail and when not. Inform user about t


KillariDev:
looks like some debugging code left here, also this silently fails?

_Originally posted by @jubalm in https://github.com/DarkFlorist/TheInterceptor/pull/712#discussion_r1372707857_

PR #717, opened 2023-10-30
Visualize governance vote simulations

KillariDev:
fix https://github.com/DarkFlorist/TheInterceptor/issues/716

image
image
image


app/ts/simulation/compoundGovernanceFaking.ts
```diff
@@ -41,32 +45,32 @@ export const simulateCompoundGovernanceExecution = async (ethereumClientService:
...
 	]
 	const governanceContractCalls = await ethereumClientService.multicall(calls, [], parentBlock.number)
+	governanceContractCalls.forEach((call) => { if (call.statusCode !== 'success') throw new Error('Failed to retrieve governance contracts information') })
 	if (governanceContractCalls[0]?.returnValue === undefined) return undefined
 	const timeLockContractResult = compoundGovernanceAbi.decodeFunctionResult('timelock', governanceContractCalls[0].returnValue)
 	const timeLockContract = EthereumAddress.parse(timeLockContractResult[0])
 	if (governanceContractCalls[1] === undefined) throw new Error('proposals return value was undefined')
-	const [_id, _proposer, eta, _startBlock, _endBlock, _forVotes, _againstVotes, _abstainVotes, canceled, executed] = compoundGovernanceAbi.decodeFunctionResult('proposals', governanceContractCalls[1].returnValue)
+	const proposal = compoundGovernanceAbi.decodeFunctionResult('proposals', governanceContractCalls[1].returnValue)
+	const eta: bigint = funtypes.BigInt.parse(proposal[0].eta)
```
jubalm, line 54: maybe `safeParse` here so it doesn't throw for whatever unknown reason

jubalm (approved):

app/ts/components/simulationExplaining/customExplainers/GovernanceVoteVisualizer.tsx
```diff
@@ -0,0 +1,107 @@
...
+								renameAddressCallBack = { param.renameAddressCallBack }
+								activeAddress = { param.simulationAndVisualisationResults.activeAddress }
+								resetButton = { false }
+								currentBlockNumber = { currentBlockNumber }
+								rpcConnectionStatus = { rpcConnectionStatus }
+							/>
+						</> }
+					</div>
+				</> : <div style = 'display: flex; justify-content: center;'>
+					{ !(param.simulationAndVisualisationResults.rpcNetwork.httpsRpc === 'https://rpc.dark.florist/birdchalkrenewtip' // todo remove this check
```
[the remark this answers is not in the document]
KillariDev: I added issue about this

app/ts/simulation/compoundGovernanceFaking.ts
KillariDev: I want it to throw in this case and stop the execution as that means the contract behaves differently than I assume, and that probably means I am doing something wrong. We should not show simulation results if the contract behaves differently than I assume, as I would perform wrong simulation

Issue #706, opened 2023-10-20
Allow contracts to have ABI in them


KillariDev:
- button to import ABI from
- save imported ABI's
- show ABI in logs / contract methods
- verify imported ABI
- allow pasting the ABI from somewhere

PR #720, opened 2023-10-30
Retrieve ABI from etherscan

KillariDev:
Fixes  partially https://github.com/DarkFlorist/TheInterceptor/issues/706

- Adds button to import ABI from etherscan
- ABI is saved to addressbook

still missing that user could see the abi somewhere and that user could paste the abi themselves (we need verification for this), and user should be able to change eherscan api key

image

### DIFF
--- a/app/ts/simulation/services/EtherScanAbiFetcher.ts
+++ b/app/ts/simulation/services/EtherScanAbiFetcher.ts
@@ -5,7 +5,9 @@ import { addressString } from '../../utils/bigint.js'
 const EtherScanABIKey = 'PSW8C433Q667DVEX5BCRMGNAH9FSGFZ7Q8'
 
 export async function fetchAbi(contractAddress: EthereumAddress) {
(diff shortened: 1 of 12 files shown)


MicahZoltu (approved):
Beware that merging this PR is going to close #706.  Be sure to re-open it since it isn't fully fixed, or open new issues with the unfinished parts.

I definitely think we need a way to let the user provide their own ABI.  I don't think we need to validate it though.  We can just trust that they gave the right one.  We would want untrusted if it was coming from a dApp.

app/ts/simulation/services/EtherScanAbiFetcher.ts
```diff
@@ -0,0 +1,32 @@
+import { EtherscanGetABIResult, EtherscanSourceCodeResult } from '../../types/etherscan.js'
+import { EthereumAddress } from '../../types/wire-types.js'
+import { addressString } from '../../utils/bigint.js'
+
+const EtherScanABIKey = 'PSW8C433Q667DVEX5BCRMGNAH9FSGFZ7Q8'
+
+async function fetchJsonOrUndefined(url: string): Promise<string | undefined> {
+	const response = await fetch(url)
+	if (!response.ok) return undefined // TODO, here we might want to propagate the error of not being able to reach etherscan to user
```
KillariDev, line 9: get error if we cannot reach etherscan and propagate to user

PR #724, opened 2023-10-31
Check for abi on governance vote and remove hard coded abi's (just use user added ones)

KillariDev:
image


app/ts/background/background.ts
```diff
@@ -55,8 +55,8 @@ async function updateMetadataForSimulation(simulationState: SimulationState, eth
 	}
 }
 
-export const simulateGovernanceContractExecution = async (pendingTransaction: PendingTransaction, ethereum: EthereumClientService) => {
-	const returnError = (text: string) => ({ error: { message: text} })
+export const simulateGovernanceContractExecution = async (pendingTransaction: PendingTransaction, ethereum: EthereumClientService, userAddressBook: UserAddressBook) => {
+	const returnError = (text: string) => ({ error: { type: 'Other' as const, message: text } })
```
MicahZoltu, line 59: Recommend a discriminated union if you are going to be returning multiple different types of objects.
```suggestion
	const returnError = (text: string) => ({ success: false, error: { type: 'Other' as const, message: text } })
```
A similar change would need to be made to the happy-path return so it returns `{ success: true, ... }`.

Typescript has special support for discriminated unions (a union of objects where one key on the object is unique for each subtype in the union), and dealing with discriminated unions is generally a better developer experience in TS than dealing with unions that have different shapes but no discriminant.

app/ts/background/background.ts
```diff
@@ -65,13 +65,16 @@ export const simulateGovernanceContractExecution = async (pendingTransaction: Pe
...
 		const params = governanceContractInterface.decodeFunctionData(explanation === 'Submit vote' ? 'submitVote' : 'castVote', dataStringWith0xStart(pendingTransaction.transactionToSimulate.transaction.input))
-		const contractExecutionResult = await simulateCompoundGovernanceExecution(ethereum, pendingTransaction.transactionToSimulate.transaction.to, params[0])
+		const addr = await identifyAddress(ethereum, userAddressBook, pendingTransaction.transactionToSimulate.transaction.to)
+		if (!('abi' in addr) || addr.abi === undefined) return { error: { type: 'MissingAbi' as const, message: 'ABi for the governance contract is missing', addressBookEntry: addr } }
+		const contractExecutionResult = await simulateCompoundGovernanceExecution(ethereum, addr, params[0])
 		if (contractExecutionResult === undefined) return returnError('Failed to simulate governacne execution')
 		const parentBlock = await ethereum.getBlock()
-		if (parentBlock.baseFeePerGas === undefined) throw new Error('cannot build simulation from legacy block')
+		if (parentBlock.baseFeePerGas === undefined) return returnError('cannot build simulation from legacy block')
```
MicahZoltu, line 77: I wonder if, at this point, we could just remove support for pre-London blocks entirely?  Is there anything we do in Interceptor that actually needs to support them?

app/ts/background/background.ts
```diff
@@ -102,6 +105,7 @@ export const simulateGovernanceContractExecution = async (pendingTransaction: Pe
 		}
 		return await visualizeSimulatorState(governanceContractSimulationState, ethereum)
 	} catch(error) {
+		console.warn(error)
 		if (error instanceof Error) return returnError(error.message)
 		throw error
```
MicahZoltu, line 110: Why do you throw an error if it isn't an `instanceof Error` rather than returning it like all of the other errors?

app/ts/background/background.ts
```diff
@@ -102,6 +105,7 @@ export const simulateGovernanceContractExecution = async (pendingTransaction: Pe
 		}
 		return await visualizeSimulatorState(governanceContractSimulationState, ethereum)
 	} catch(error) {
+		console.warn(error)
```
MicahZoltu, line 108: Debugging code, or intentionally left in?  Since this isn't the terminal handler of errors, my instinct is to not log it to the console.  I generally reserve that for errors that are unhandled anywhere else (the last point of handling).

app/ts/components/simulationExplaining/customExplainers/GovernanceVoteVisualizer.tsx
```diff
@@ -20,6 +21,23 @@ export function GovernanceTransactionExecution(param: TransactionCardParams) {
 	</>
 }
 
+export type MissingAbiParams = {
+	errorMessage: String
```
MicahZoltu, line 25:
```suggestion
	errorMessage: string
```
`String != string`

app/ts/components/simulationExplaining/customExplainers/GovernanceVoteVisualizer.tsx
```diff
@@ -57,7 +75,8 @@ export function GovernanceVoteVisualizer(param: TransactionImportanceBlockParams
 					{ simulateGovernanceContractExecutionReply !== undefined ? <>
 						<button
 							class = { `button is-primary is-small` }
-							onClick = { simulateGovernanceVote }>
+							onClick = { simulateGovernanceVote }
+						>
 							Refresh
 						</button>
```
MicahZoltu, line 81: Style nitpick: This feels unnecessary noise in the vertical direction.  Consider one-lining it to encourage visual scanning of the code.
```suggestion
						<button class = { `button is-primary is-small` } onClick = { simulateGovernanceVote }>Refresh</button>
```

app/ts/components/simulationExplaining/customExplainers/GovernanceVoteVisualizer.tsx
```diff
@@ -57,7 +75,8 @@ export function GovernanceVoteVisualizer(param: TransactionImportanceBlockParams
...
 					</> : <></>}
```
MicahZoltu, line 82: Consider creating locally scoped components to improve readability:
Earlier in this function (after you set `simulateGovernanceVote` and `simulateGovernanceContractExecutionReply`) add something like this:
```ts
const [MaybeRefreshButton_] = useState(() => () => simulateGovernanceContractExecutionReply !== undefined ? <button class = { `button is-primary is-small` } onClick = { simulateGovernanceVote }>Refresh</button> : <>)
```
Then here you would put:
```suggestion
					<MaybeRefreshButton_/>
```
The nice thing about function-local components is that you can reference captured variables so you don't need to have properties on the component and pass a bunch of stuff around.

Note: The `[]` and the doubly-nested lambda are both necessary.  The `_` suffix on the variable name is a style I use in my personal projects to help add clarity when a component is function-local vs file-local or imported, as well as to avoid naming collisions since for these local functions I often find myself wanting to name them things like `Button` or `Refresh` which often collide with existing names of things.

It may be worth multi-lining the button definition above if the one-liner is too noisy in this case.

app/ts/components/simulationExplaining/customExplainers/GovernanceVoteVisualizer.tsx
```diff
@@ -67,7 +86,13 @@ export function GovernanceVoteVisualizer(param: TransactionImportanceBlockParams
...
 				{ simulateGovernanceContractExecutionReply !== undefined ? <>
 					<div style = 'display: grid; grid-template-rows: max-content' >
-						{ 'error' in simulateGovernanceContractExecutionReply.data ? <ErrorComponent warning = { false } text = { simulateGovernanceContractExecutionReply.data.error.message }/> : <>
+						{ 'error' in simulateGovernanceContractExecutionReply.data ? <>
+							{ simulateGovernanceContractExecutionReply.data.error.type === 'MissingAbi' ? <MissingAbi
+								errorMessage = { simulateGovernanceContractExecutionReply.data.error.message }
+								addressBookEntry = { simulateGovernanceContractExecutionReply.data.error.addressBookEntry }
+								renameAddressCallBack = { param.renameAddressCallBack }
+							/> : <ErrorComponent warning = { false } text = { simulateGovernanceContractExecutionReply.data.error.message }/> }
+						</> : <>
```
MicahZoltu, line 95: Similar comment as above.  Anytime I find myself doing ternaries inside of my TSX blocks I find it valuable to extract the code out into a function-local component and put the ternary in there.  I find this makes the TSX much more readable as you just have a bunch of named components in a hierarchy rather than a bunch of code.  You can also have multiple layers of named function-local components if you have a situation like this where you have a conditional inside a conditional.

app/ts/components/simulationExplaining/customExplainers/GovernanceVoteVisualizer.tsx
```diff
@@ -101,6 +128,14 @@ export function GovernanceVoteVisualizer(param: TransactionImportanceBlockParams
 							}>
 							Simulate execution on a passing vote
 						</button>
+					: <> { param.simTx.transaction.to !== undefined ? 
+							<MissingAbi
+								errorMessage = { 'The governance contract is missing an ABI. Add ABI to simulate with the contract.' }
```
MicahZoltu, line 133:
```suggestion
								errorMessage = { 'The governance contract is missing an ABI. Add an ABI to simulate execution of this proposal.' }
```

MicahZoltu (approved):
Why did you decide to remove the built-in ABI for `0x6f3e6272a167e8accb32072d08e0957f9c79223d`?  Seems like there is some value in us baking in some common/popular governance contracts rather than requiring the user add them manually or go to Etherscan.

KillariDev:
> Why did you decide to remove the built-in ABI for `0x6f3e6272a167e8accb32072d08e0957f9c79223d`? Seems like there is some value in us baking in some common/popular governance contracts rather than requiring the user add them manually or go to Etherscan.

I feel this should be in address metadata or somewhere else than in Interceptor. And I don't feel we should add that many ABI's that are not really widely used. That's abi for Nouns DAO, I am not expecting that many Nouns DAO voters to use interceptor that it makes sense to include it

app/ts/background/background.ts
KillariDev: > I wonder if, at this point, we could just remove support for pre-London blocks entirely? Is there anything we do in Interceptor that actually needs to support them?

a DAPP can Query for an old block header/block I think. Some dapp was querying for block 1 for example (cannot remember which)

app/ts/background/background.ts
KillariDev: If something goes wrong, I feel its good that there's some place where you can read more details on what went wrong, even thought its expected that errors can occur. I am not sure how else handle that

PR #725, opened 2023-11-01
parse events with ABI and display the results

KillariDev:
- Some refactoring on solidity type parsing
- When ABI is available, parse events and display them
image


app/ts/utils/abi.ts
```diff
@@ -81,3 +81,18 @@ export const CompoundGovernanceAbi: InterfaceAbi = [
 export const CompoundTimeLock: InterfaceAbi = [
 	'function executeTransactions(address[] memory targets, uint[] memory values, string[] memory signatures, bytes[] memory datas, uint eta) public payable'
 ]
+
+function getStringBetweenParentheses(inputString: string): string | undefined {
+	const regex = /\(([^)]+)\)/
```
MicahZoltu, line 86: I believe this will do the trick, and also work with `()` (returning empty string).  Note that the `?` tells the regex to not be greedy, which will make it stop as soon as it runs into the closing `)` since that can be matched by the next item.
```suggestion
	const regex = /\((.*?)\)/
```

app/ts/utils/abi.ts
```diff
@@ -81,3 +81,18 @@ export const CompoundGovernanceAbi: InterfaceAbi = [
...
+	const match = inputString.match(regex)
+	if (match) return match[1]
+	return undefined
+}
+
+// Transfer(address,address,uint256) -> ['address', 'address', 'uint256']
+export const extractFunctionArgumentTypes = (signature: string) => {
+	const args = getStringBetweenParentheses(signature)
+	return args === undefined ? undefined : args.split(',')
+}
```
MicahZoltu, line 96: Are you confident you will always have the canonical function signature?  `function Transfer(address sender, address receiver, uint256 amount)` is valid (but not canonical) in many places.

MicahZoltu (approved):

app/ts/components/simulationExplaining/Transactions.tsx
```diff
@@ -305,29 +305,50 @@ export function TokenLogAnalysis(param: LogAnalysisParams) {
...
-	nonTokenLog: MulticallResponseEventLog
+	nonTokenLog: MaybeParsedEvent
+}
+
+function insertBetweenElements<T>(array: readonly T[], elementToInsert: T): readonly T[] {
+	return array.reduce((acc, currentElement, index, arr) => {
+		acc.push(currentElement)
+		if (index < arr.length - 1) acc.push(elementToInsert)
+		return acc
+	}, [] as T[])
```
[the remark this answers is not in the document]
KillariDev: hmm I think the for loop is the clearest for me to read in this case

app/ts/components/subcomponents/solidityType.tsx
```diff
@@ -0,0 +1,34 @@
+import { EnrichedGroupedSolidityType, PureGroupedSolidityType } from '../../types/solidityType.js'
+import { RenameAddressCallBack } from '../../types/user-interface-types.js'
+import { checksummedAddress, dataStringWith0xStart } from '../../utils/bigint.js'
+import { assertNever } from '../../utils/typescript.js'
+import { SmallAddress } from './address.js'
+
+const textStyle = 'text-overflow: ellipsis; overflow: hidden;'
+const StringElement = ({ text }: { text: string }) => <p class = 'paragraph' style = { textStyle }>{ `${ text }` }</p>
```
jubalm, line 8: text is already a string here

```suggestion
const StringElement = ({ text }: { text: string }) => <p class = 'paragraph' style = { textStyle }>{ text }</p>
```

jubalm (changes requested):

jubalm (approved):

Issue #727, opened 2023-11-01
parse logs with ethers first, and only then figure out if there's ERC20/ERC721's etc there


KillariDev:
this PR https://github.com/DarkFlorist/TheInterceptor/pull/725 parses logs, but only for non-token events and in too late stage. Move the parsing earlier and refactor token parsing to use output of this

PR #730, opened 2023-11-01
unify non-token events and events

KillariDev:
- Now all logs are being processed at the same place

fix https://github.com/DarkFlorist/TheInterceptor/issues/727


app/ts/background/metadataUtils.ts
```diff
@@ -139,25 +140,21 @@ export async function identifyAddress(ethereumClientService: EthereumClientServi
...
 		if (vis === undefined) continue
 		const ethBalanceAddresses = vis.ethBalanceChanges.map((x) => x.address)
-		const from = vis.tokenResults.map((x) => x.from)
-		const to = vis.tokenResults.map((x) => x.to)
-		tokenAddresses = tokenAddresses.concat(vis.tokenResults.map((x) => x.tokenAddress))
-		addressesToFetchMetadata = addressesToFetchMetadata.concat(ethBalanceAddresses, from, to)
+		const eventArguments = vis.events.map((x) => x.type !== 'NonParsed' ? x.args : []).flat()
+		const addressesInEvents = eventArguments.filter((a): a is { typeValue: { type: 'address', value: EthereumAddress }, paramName: string } => a.typeValue.type === 'address').map((a) => a.typeValue.value)
+		addressesToFetchMetadata = addressesToFetchMetadata.concat(ethBalanceAddresses, addressesInEvents, vis.events.map((a) => a.loggersAddress))
 	}
```
MicahZoltu, line 150:
```suggestion
	for (const visualization of visualizerResult) {
		const eventArguments = visualization.events.map((x) => x.type !== 'NonParsed' ? x.args : []).flat()
		const addressesInEvents = eventArguments.filter((a): a is { typeValue: { type: 'address', value: EthereumAddress }, paramName: string } => a.typeValue.type === 'address').map((a) => a.typeValue.value)
		addressesToFetchMetadata = addressesToFetchMetadata.concat(ethBalanceAddresses, addressesInEvents, visualization.events.map((a) => a.loggersAddress))
	}
```

app/ts/background/metadataUtils.ts
```diff
@@ -139,25 +140,21 @@ export async function identifyAddress(ethereumClientService: EthereumClientServi
...
 
 	for (const vis of visualizerResult) {
 		if (vis === undefined) continue
 		const ethBalanceAddresses = vis.ethBalanceChanges.map((x) => x.address)
-		const from = vis.tokenResults.map((x) => x.from)
-		const to = vis.tokenResults.map((x) => x.to)
-		tokenAddresses = tokenAddresses.concat(vis.tokenResults.map((x) => x.tokenAddress))
-		addressesToFetchMetadata = addressesToFetchMetadata.concat(ethBalanceAddresses, from, to)
+		const eventArguments = vis.events.map((x) => x.type !== 'NonParsed' ? x.args : []).flat()
+		const addressesInEvents = eventArguments.filter((a): a is { typeValue: { type: 'address', value: EthereumAddress }, paramName: string } => a.typeValue.type === 'address').map((a) => a.typeValue.value)
```
MicahZoltu, line 148: Why `x` on the first line and `a` on the second?

app/ts/background/metadataUtils.ts
```diff
@@ -170,9 +167,9 @@ export async function nameTokenIds(ethereumClientService: EthereumClientService,
 	}
 	let tokenAddresses: TokenAddressTokenIdPair[] = visualizerResult.map((vis) => {
 		if (vis === undefined) return undefined
-		return vis.tokenResults.map((x) => { 
-			if (x.type !== 'ERC1155') return undefined
-			return { tokenAddress: x.tokenAddress, tokenId: x.tokenId }
+		return vis.events.map((x) => { 
+			if (x.type !== 'TokenEvent' || x.tokenInformation.type !== 'ERC1155') return undefined
+			return { tokenAddress: x.tokenInformation.tokenAddress, tokenId: x.tokenInformation.tokenId }
 		})
```
MicahZoltu, line 173: Same commente as above, recommend `visualization` instead of `vis`.

app/ts/simulation/simulator.ts
```diff
@@ -36,19 +43,53 @@ const logHandler = new Map<string, Loghandler>([
...
-export const visualizeTransaction = (blockNumber: bigint, singleMulticallResponse: SingleMulticallResponse) => {
-	if (singleMulticallResponse.statusCode !== 'success') return { ethBalanceChanges: [], tokenResults: [], blockNumber }
-	let tokenResults: TokenVisualizerResult[] = []
-	for (const eventLog of singleMulticallResponse.events) {
-		const logSignature = eventLog.topics[0]
-		if (logSignature === undefined) continue
+const parseEvents = async (singleMulticallResponse: SingleMulticallResponse, ethereumClientService: EthereumClientService, userAddressBook: UserAddressBook): Promise<MaybeParsedEvents> => {
+	if (singleMulticallResponse.statusCode !== 'success' ) return []
+	return await Promise.all(singleMulticallResponse.events.map(async (event) => {
+		// todo, we should do this parsing earlier, to be able to add possible addresses to addressMetaData set
```
MicahZoltu, line 49: Do we have a ticket for this so it doesn't get lost/forgotten?

app/ts/types/visualizer-types.ts
```diff
@@ -22,43 +22,85 @@ export const NetworkPrice = funtypes.ReadonlyObject({
...
+	typeValue: PureGroupedSolidityType,
+	paramName: funtypes.String
+})
+
+export type ParsedEvent = funtypes.Static<typeof ParsedEvent>
+export const ParsedEvent = funtypes.ReadonlyObject({
+	isParsed: funtypes.Literal('Parsed'),
+	name: funtypes.String, // eg. 'Transfer'
+	signature: funtypes.String, // eg. 'Transfer(address,address,uint256)'
+	args: funtypes.ReadonlyArray(SolidityVariable), // TODO: add support for structs (abiV2)
```
MicahZoltu, line 36: Do we have an issue for this?

app/ts/utils/abi.ts
```diff
@@ -82,6 +84,17 @@ export const CompoundTimeLock: InterfaceAbi = [
...
+	if (loggerAddress === 0x0n) return weth9Abi
```
MicahZoltu, line 89: Why does address 0 return an ABI for weth9?

If this is for `eth_multicallV1`'s ETH events, I think we should just hand-craft an ABI that only has that one event defined on it and use that instead of using weth9 ABI.

MicahZoltu (approved):
Would like to see comments addressed or discussed, but no blockers for merging.

jubalm (approved):

app/ts/simulation/simulator.ts
KillariDev: its done in later PR

app/ts/types/visualizer-types.ts
KillariDev: added issue

app/ts/utils/abi.ts
KillariDev: this is to get withdraw / deposit ABI's as well. Other option is that we hardcode weth9 contract to its address for each chain and then here just return standard erc20 here. I was thinking we don't make a big error if we assume a random ERC20 would have the same ABI than weth9

app/ts/background/metadataUtils.ts
KillariDev: hmm I am not understanding your changes. You are leaving `from`, `to` and `tokenAddress` fields out from metadata retrieval

app/ts/background/metadataUtils.ts
KillariDev: no particular reason, I added better var names

app/ts/utils/abi.ts
MicahZoltu: Address `0` doesn't have a withdraw/deposit function.

MicahZoltu: For clarity, my concern here isn't about using a generic ERC20 token contract.  It is its assignment to address 0 which has no contract code at it.

app/ts/background/metadataUtils.ts
MicahZoltu: Bleh.  This was a mistake on my end.  I just meant to rename `vis` to `visualization` and GitHub UI did the wrong thing.

PR #740, opened 2023-11-06
Add abi to metadata addr

KillariDev:
- Now user is able to add ABI to metadata entries

fix https://github.com/DarkFlorist/TheInterceptor/issues/729

image

### DIFF
--- a/app/ts/background/medataSearch.ts
+++ b/app/ts/background/medataSearch.ts
@@ -73,6 +73,12 @@ const convertContractDefinitionToAddressBookEntry = ([address, def]: [string, Co
 	entrySource: 'DarkFloristMetadata' as const,
 })
 
+function concatArraysUniqueByAddress<T>(addTo: readonly (T & { address: bigint })[], addFrom: readonly (T & { address: bigint })[]) {
+	const existingValues = new Set(addTo.map(item => addressString(item.address)))
+	const uniqueItems = addFrom.filter(item => !existingValues.has(addressString(item.address)))
+	return [...addTo, ...uniqueItems]
+}
+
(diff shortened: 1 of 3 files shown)


app/ts/background/medataSearch.ts
```diff
@@ -73,14 +73,20 @@ const convertContractDefinitionToAddressBookEntry = ([address, def]: [string, Co
 	entrySource: 'DarkFloristMetadata' as const,
 })
 
+function concatArraysUniqueByAddress<T>(addTo: readonly (T & { address: bigint })[], addFrom: readonly (T & { address: bigint })[]) {
```
jubalm, line 76: not sure if `T` here should also be cast within the function but can alternatively use ReadonlyArray as well

```suggestion
function concatArraysUniqueByAddress<T>(addTo: ReadonlyArray<T & { address: bigint }>, addFrom: ReadonlyArray<T & { address: bigint }>) {
```

app/ts/components/pages/AddNewAddress.tsx
```diff
@@ -415,8 +412,6 @@ export function AddNewAddress(param: AddAddressParam) {
 
 	function isSubmitButtonDisabled() {
 		return !areInputValid()
```
jubalm, line 414: not actually change related but could use some verb grammar for `areInputValid` to either `isInputValid` or `areInputsValid`. it would require updating multiple lines.

jubalm (approved):

app/ts/background/medataSearch.ts
KillariDev: these are the same thigns

PR #747, opened 2023-11-07
cache Rpc Requests, invalidate on new block

KillariDev:
- fix https://github.com/DarkFlorist/TheInterceptor/issues/681


app/ts/simulation/services/EthereumClientService.ts
```diff
@@ -77,11 +77,13 @@ export class EthereumClientService {
 		if (this.retrievingBlock) return
 		try {
 			this.retrievingBlock = true
-			const response = await this.requestHandler.jsonRpcRequest({ method: 'eth_getBlockByNumber', params: ['latest', true] })
+			const response = await this.requestHandler.jsonRpcRequest({ method: 'eth_getBlockByNumber', params: ['latest', true] }, true)
 			if (this.cacheRefreshTimer === undefined) return
 			const newBlock = EthereumBlockHeader.parse(response)
 			console.log(`Current block number: ${ newBlock.number }`)
```
jubalm, line 83: these would be a good use case for `console.info`

KillariDev: hmm I don't really see difference between this and console.log

PR #748, opened 2023-11-07
allow removing all user added entries and not others

KillariDev:
Allow user to remove user added entries (and not just active addresses and contacts)


app/css/interceptor.css
```diff
@@ -343,6 +343,13 @@ a.dropdown-item.is-active, button.dropdown-item.is-active {
 	padding: 0;
 	padding-left: 0.75rem;
 	padding-right: 0.75rem;
+	background-color: transparent;
+}
+
+.card-header-icon[disabled] {
```
jubalm, line 349: you might be targeting "disabled state" here and not presence of "disabled" attribute on the button

```suggestion
.card-header-icon:disabled {
```

app/ts/AddressBook.tsx
```diff
@@ -145,13 +145,8 @@ export function ListElement(entry: ListElementParam) {
...
 
 					<div class = 'content' style = 'color: var(--text-color); display: flex; height: 100%; flex-direction: column; justify-content: space-between;'>
-						<button class = 'card-header-icon' style = 'padding: 0px; margin-left: auto;' aria-label = 'delete'>
-							<span
-								class = 'icon'
-								style = 'color: var(--text-color);'
-								onClick = { entry.type != 'empty' && (entry.category === 'My Active Addresses' || entry.category === 'My Contacts')  ? () => entry.removeEntry(entry) : () => {} }>
-								X
-							</span>
+						<button class = 'card-header-icon' style = 'padding: 0px; margin-left: auto;' aria-label = 'delete' disabled = { entry.type === 'empty' || entry.entrySource !== 'User' } onClick = { entry.type !== 'empty' ? () => entry.removeEntry(entry) : () => {} }>
```
jubalm, line 148: there should be no significant performance difference, just more common to see a single function declaration for click assignment and handle the condition within

```suggestion
						<button class = 'card-header-icon' style = 'padding: 0px; margin-left: auto;' aria-label = 'delete' disabled = { entry.type === 'empty' || entry.entrySource !== 'User' } onClick = { () => entry.type !== 'empty' ? entry.removeEntry(entry) : undefined }>
```

jubalm (approved):

PR #754, opened 2023-11-09
Fix export settings

KillariDev:
Fixes export settings. Does not fix the import settings problem on firefox

### DIFF
--- a/app/ts/background/settings.ts
+++ b/app/ts/background/settings.ts
@@ -7,7 +7,7 @@ import { ActiveAddressArray, ContactEntries } from '../types/addressBookTypes.js
 import { WebsiteAccessArray } from '../types/websiteAccessTypes.js'
 import { RpcEntries, RpcNetwork } from '../types/rpc.js'
 import { NetworkPrice } from '../types/visualizer-types.js'
-import { browserStorageLocalGet, browserStorageLocalGetUnParsed, browserStorageLocalSet } from '../utils/storageUtils.js'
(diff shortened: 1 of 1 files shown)


app/ts/background/settings.ts
```diff
@@ -182,26 +182,27 @@ export const setUseTabsInsteadOfPopup = async(useTabsInsteadOfPopup: boolean) =>
 export const getMetamaskCompatibilityMode = async() => (await browserStorageLocalGet('metamaskCompatibilityMode'))?.['metamaskCompatibilityMode'] ?? false
 export const setMetamaskCompatibilityMode = async(metamaskCompatibilityMode: boolean) => await browserStorageLocalSet({ metamaskCompatibilityMode })
 
-
-export async function exportSettingsAndAddressBook() {
-	const results = {
+export async function exportSettingsAndAddressBook(): Promise<ExportedSettings> {
+	const exportDate = (new Date).toISOString().split('T')[0]
+	if (exportDate === undefined) throw new Error('Datestring did not contain Date')
```
jubalm, line 187: if the intent is to compress storage size, this may also work tho the output will differ slightly (with slashes, `11/11/2023`) and should still be a valid Date string. can also get away without checks

```suggestion
	const exportDate = (new Date).toLocaleDateString('en-US')
```

jubalm (approved):

app/ts/background/settings.ts
KillariDev: The point is to just have the date, not datetime. And I would rather use ISO date than US dates as they have weirdly month and day reversed :(

PR #761, opened 2023-11-12
Fix visualization for unidentified EIP712 message

KillariDev:
Seems like using `key` as variable name is not good idea and produces `undefined`, I am not really sure why, but this fixes the issue


jubalm (approved):
LGTM

PR #769, opened 2023-11-18
Cleans up styling

KillariDev:
align NFT sending rows


jubalm (approved):
nothing particular within the changeset, tho i'd generally move styling to css (paddings and margins) instead of using `&nbsp;` unless it actually involves within the flow of content for adding spaces between elements. nevertheless i'm no opposed to the implementation here either

Issue #771, opened 2023-11-19
connecting to aave is broken


KillariDev:
https://app.aave.com/

PR #774, opened 2023-11-21
Access management

KillariDev:
- fix https://github.com/DarkFlorist/TheInterceptor/issues/771
- fixes connecting to ENS

fixes bugs in access management:
- on race condition that multiple requests are made that requires access, they are all queued and replied at once
- handle long chains of access requests:
1) dapp asks access for account on signing mode
-> we prompt access to site
-> provider prompt to get address from provider 
-> interceptor prompt access to address that were gotten from provider
-> reply to dapp that we have access

### DIFF
--- a/app/inpage/ts/inpage.ts
+++ b/app/inpage/ts/inpage.ts
@@ -207,7 +207,7 @@ class InterceptorMessageListener {
 	}
 
 	private readonly WindowEthereumSend = async (payload: { readonly id: string | number | null, readonly method: string, readonly params: readonly unknown[], _params: readonly unknown[] } | string, maybeCallBack: undefined | ((error: IJsonRpcError | null, response: IJsonRpcSuccess<unknown> | null) => void)) => {
-		const fullPayload = typeof payload === 'string' ? { method: payload, id: 0, params: [] } : payload
+		const fullPayload = typeof payload === 'string' ? { method: payload, id: 1, params: [] } : payload
 		if (maybeCallBack !== undefined && typeof maybeCallBack === 'function') return this.WindowEthereumSendAsync(fullPayload, maybeCallBack)
 		if (this.metamaskCompatibilityMode) {
 			if (window.ethereum === undefined) throw new Error('window.ethereum is missing')
--- a/app/inpage/ts/inpage.ts
+++ b/app/inpage/ts/inpage.ts
@@ -206,16 +206,17 @@ class InterceptorMessageListener {
 		}
 	}
 
-	private readonly WindowEthereumSend = async (payload: { readonly id: string | number | null, readonly method: string, readonly params: readonly unknown[], _params: readonly unknown[] }, maybeCallBack: undefined | ((error: IJsonRpcError | null, response: IJsonRpcSuccess<unknown> | null) => void)) => {
-		if (maybeCallBack !== undefined && typeof maybeCallBack === 'function') return this.WindowEthereumSendAsync(payload, maybeCallBack)
+	private readonly WindowEthereumSend = async (payload: { readonly id: string | number | null, readonly method: string, readonly params: readonly unknown[], _params: readonly unknown[] } | string, maybeCallBack: undefined | ((error: IJsonRpcError | null, response: IJsonRpcSuccess<unknown> | null) => void)) => {
+		const fullPayload = typeof payload === 'string' ? { method: payload, id: 0, params: [] } : payload
+		if (maybeCallBack !== undefined && typeof maybeCallBack === 'function') return this.WindowEthereumSendAsync(fullPayload, maybeCallBack)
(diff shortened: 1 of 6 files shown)


app/ts/background/providerMessageHandlers.ts
```diff
@@ -17,7 +17,11 @@ export async function ethAccountsReply(simulator: Simulator, websiteTabConnectio
 	if (port.sender?.tab?.id === undefined) return returnValue
 
 	const signerAccountsReply = EthereumAccountsReply.parse(request.params)
-	const signerAccounts = signerAccountsReply[0]
+	if (signerAccountsReply[0].type === 'error') {
+		console.warn('Wallet returned an error!') // TODO, propagate this to the UI for user to see
```
KillariDev, line 21: propagate wallet errors to UI

app/inpage/ts/inpage.ts
```diff
@@ -284,12 +287,33 @@ class InterceptorMessageListener {
...
+			await this.sendMessageToBackgroundPage({ method: 'eth_accounts_reply', params: [{ type: 'success', accounts: this.signerAccounts, requestAccounts: ask_eth_requestAccounts }] })
+			if (this.waitForAccountsFromWallet !== undefined) {
+				this.waitForAccountsFromWallet.resolve(true)
+				this.waitForAccountsFromWallet = undefined
+			}
+			return
+		}
+		if (this.pendingSignerAddressRequest !== undefined) {
+			await this.pendingSignerAddressRequest
+			await this.sendMessageToBackgroundPage({ method: 'eth_accounts_reply', params: [{ type: 'success', accounts: this.signerAccounts, requestAccounts: ask_eth_requestAccounts }] })
```
MicahZoltu, line 302: Does this need to be in a try/catch block like the one below?  Same question for the one above.

app/ts/background/providerMessageHandlers.ts
MicahZoltu: Do we not have a generalized solution for this?  Feels risky to merge a PR that has errors only going to the console.

app/ts/background/windows/interceptorAccess.ts
```diff
@@ -211,12 +217,12 @@ async function resolve(simulator: Simulator, websiteTabConnections: WebsiteTabCo
...
-		if (request !== undefined) await handleContentScriptMessage(simulator, websiteTabConnections, request, website, activeAddress)
 	}
 	
-	const pendingRequests = await updatePendingAccessRequests(async (previousPendingAccessRequests) => {
-		return previousPendingAccessRequests.filter((x) => !(x.website.websiteOrigin === website.websiteOrigin && (x.requestAccessToAddress?.address === accessReply.requestAccessToAddress || x.requestAccessToAddress?.address === accessReply.originalRequestAccessToAddress)))
-	})
+	const isAffectedEntry = (pending: PendingAccessRequest) => pending.website.websiteOrigin === website.websiteOrigin && (pending.requestAccessToAddress?.address === accessReply.requestAccessToAddress || pending.requestAccessToAddress?.address === accessReply.originalRequestAccessToAddress) 
+
+	const pendingRequests = await updatePendingAccessRequests(async (previousPendingAccessRequests) => previousPendingAccessRequests.filter((pending) => !isAffectedEntry(pending)))
+	const affected = pendingRequests.previous.filter((pending) => isAffectedEntry(pending))
```
MicahZoltu, line 225: Can this be moved down to line 234, or does it need to happen prior to the two conditionals?

app/ts/background/windows/interceptorAccess.ts
```diff
@@ -225,6 +231,10 @@ async function resolve(simulator: Simulator, websiteTabConnections: WebsiteTabCo
 		await closePopupOrTab(openedDialog.popupOrTab)
 		openedDialog = undefined
 	}
+
+	await Promise.all(affected.map(async (r) => {
+		if (r.request) await handleInterceptedRequest(undefined, r.website.websiteOrigin, r.website, simulator, r.socket, r.request, websiteTabConnections)
```
MicahZoltu, line 236: It feels like this conditional should be extracted out into a `.filter`  That would allow you to remove the `async` and `await` and curly braces as you can just turn this lambda into an expression.

MicahZoltu (approved):

app/ts/background/providerMessageHandlers.ts
KillariDev: well, before it was just in the pages console, I moved it to interceptors console where I plan to display it to user

app/ts/background/windows/interceptorAccess.ts
KillariDev: I don't think that's cleaner as filter doesn't manage the typing right and then we need to use `is` and construct a type that has that variable as non-undefined

Issue #710, opened 2023-10-24
store add new account details along with the page we are currently on


KillariDev:
![image](https://github.com/DarkFlorist/TheInterceptor/assets/13102010/9e7fec47-85aa-4a33-aa08-c1dca7945681)

folllowing bug occurs if you start to create new account, close interceptor popup and then open it again

PR #802, opened 2023-12-01
Refactor modify address modal

KillariDev:
- Refactor address modal so that the validation logic is happening in background
- Refactor page variable to also contain page's state when the state is important
fix https://github.com/DarkFlorist/TheInterceptor/issues/710


KillariDev:
@coderabbitai review

app/ts/background/settings.ts
```diff
@@ -131,7 +131,8 @@ export async function getSettings() : Promise<Settings> {
 	}
 }
 
-export const setPage = async (page: Page) => await browserStorageLocalSet({page})
+export const setPage = async (openedPage: Page) => await browserStorageLocalSet({ openedPage })
+export const getPage = async() => (await browserStorageLocalGet('openedPage'))?.['openedPage'] ?? { page: 'Home' }
```
jubalm, line 135: doesn't the first condition already coalesce as null?

```suggestion
export const getPage = async() => (await browserStorageLocalGet('openedPage')) ?? { page: 'Home' }
```

jubalm (approved):

app/ts/background/settings.ts
KillariDev: yeah, but then it returns `{ openedPage: { page: 'home' } }` on success, instead of the wanted `{ page: 'home' }` field

PR #804, opened 2023-12-04
Tighten up the address tags.

MicahZoltu:
Before:
image

After:
image

The `::after` element needs to have the *exact* same font details as the text inside the `p` element deep inside.  The `p` element had the `paragraph` class on it, which set the font-size to 14px causing it to be different width than the invisible text in the `::after` element.

By removing the `paragraph` class on the inner element, it will inherit its font from its parent(s).  By adding the `14px` directly onto the `small-address-container` class it will apply to both the `p` text and the `::after` text equally, thus making it so these things line up right.


MicahZoltu:
Actually fixed the issue for real, rather than hacks.  Updated PR description to reflect this.

KillariDev (approved):

Issue #800, opened 2023-12-01
Wrapping an ENS name shows 1 log, but none are present.


MicahZoltu:
Attempting to wrap an ENS name results in this:

![image](https://github.com/DarkFlorist/TheInterceptor/assets/886059/5071f567-f4a3-47e2-9b60-4d6e12834407)

PR #805, opened 2023-12-04
Log parsing

KillariDev:
Parse only contracts identified as token contracts with token events. This make it so that contracts that are not token contracts can no longer throw events that just look like tokens.

I also had to do some refactoring as `export const ExternalPopupMessage = funtypes.Union(MessageToPopup, PopupMessage)` got too big for funtypes.

This bug occurred as the ENS contract sent an event that matched with 1155, while the contract does not obey 1155 standard otherwise:
fix https://github.com/DarkFlorist/TheInterceptor/issues/800


jubalm (approved):
LGTM not sure what the rabbit would say :)

KillariDev:
> LGTM not sure what the rabbit would say :)

miss that fluffy guy!

PR #808, opened 2023-12-05
forgot that 0x0 is special case for erc20 event handling

KillariDev:
Handle logs from 0x0 address as erc20 logs, as they are the new ETH logs from multicall


jubalm (approved):
LGTM

Issue #775, opened 2023-11-21
propagate wallet errors to UI


KillariDev:
propagate wallet errors to UI

_Originally posted by @KillariDev in https://github.com/DarkFlorist/TheInterceptor/pull/774#discussion_r1400960460_

Issue #460, opened 2023-05-23
handle `MetaMask - RPC Error: Already processing eth_requestAccounts. Please wait`


KillariDev:
{code: -32002, message: 'Already processing eth_requestAccounts. Please wait.'}

PR #816, opened 2023-12-11
Signer errors

KillariDev:
fix https://github.com/DarkFlorist/TheInterceptor/issues/460
fix https://github.com/DarkFlorist/TheInterceptor/issues/775

image


app/ts/components/pages/ChangeChain.tsx
```diff
@@ -72,9 +72,7 @@ export function ChangeChain() {
 								<b> { chainChangeData.rpcNetwork.name } </b>
 							</p>
 							{ chainChangeData.rpcNetwork.httpsRpc === undefined && chainChangeData.simulationMode ?
-								<ErrorContainer
-									text = { 'This chain is not supported by The Interceptor. If you want to use this chain anyway. Select Signing mode instead of Simulation mode and attempt to change the chain again. You will then be able to disable The Interceptor and send transactions without its protection.' }
-								/>
+								<ErrorComponent text = { 'This chain is not supported by The Interceptor. If you want to use this chain anyway. Select Signing mode instead of Simulation mode and attempt to change the chain again. You will then be able to disable The Interceptor and send transactions without its protection.' }/>
```
jubalm, line 75: should we use "network" instead of "chain" here since the resulting change prompt from the user's wallet (like MetaMask) use that term?

jubalm (approved):

Issue #814, opened 2023-12-09
Bug: Erroneous error about attempting to send to something in address book.


MicahZoltu:
![image](https://github.com/DarkFlorist/TheInterceptor/assets/886059/69d5b59b-1537-49de-81ed-c0125c10a7b1)

Contract wallet recipient is in my address book, but Interceptor complains that it is not.

KillariDev:
this is 0.76? The recipient is EOA and as "contact" in interceptor? I thought I fixed this issue

PR #823, opened 2023-12-15
get rid of addressinfo/contacts, don't show onchain addressbook entries

KillariDev:
fix https://github.com/DarkFlorist/TheInterceptor/issues/814

We used to have contacts and addressinfo fields in settings. These used to contain users contacts and active addresess. However, we also had addressbook that also had this information. This PR refactors contacts and addressinfo away and we'll just use addressbook.

This should also fix issue where editing address did not change it in addressbook (because the entry was in addressbook and in contacts/addressinfo)

I also hide onchain entries from addressbook. These are information that i retrieve onchain and cache to addressbook so that we don't keep retrieving them all the time again and again.


jubalm (approved):
LGTM

Was a huge PR, lot's of refactored names anyways so just had to make sure still works

PR #828, opened 2023-12-22
add a simple search to gas estimation

KillariDev:
after estimate is gotten, add 25% and try it, if it succeeds, go with it.
If it fails, add 35% to original estimate, if it succeeds, go with it
If it fails, add 45% to original estimate, if it succeeds, go with it
if it fails, return error

### DIFF
--- a/app/ts/simulation/services/SimulationModeEthereumClientService.ts
+++ b/app/ts/simulation/services/SimulationModeEthereumClientService.ts
@@ -102,7 +102,7 @@ export const simulateEstimateGas = async (ethereumClientService: EthereumClientS
 		return { maxFeePerGas: 0n, maxPriorityFeePerGas: 0n }
 	}
 
-	const tmp = {
+	const transactionToSimulate = {
 		type: '1559' as const,
 		from: sendAddress,
 		chainId: ethereumClientService.getChainId(),
@@ -114,7 +114,7 @@ export const simulateEstimateGas = async (ethereumClientService: EthereumClientS
 		input: getInputFieldFromDataOrInput(data),
 		accessList: []
 	}
-	const multiCall = await simulatedMulticall(ethereumClientService, simulationState, [tmp], block.number + 1n)
+	const multiCall = await simulatedMulticall(ethereumClientService, simulationState, [transactionToSimulate], block.number + 1n)
 	const lastResult = multiCall[multiCall.length - 1]
 	if (lastResult === undefined) {
 		return {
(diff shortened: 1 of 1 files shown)


app/ts/simulation/services/SimulationModeEthereumClientService.ts
```diff
@@ -134,8 +134,31 @@ export const simulateEstimateGas = async (ethereumClientService: EthereumClientS
...
+		return increasedEstimate < maxGas ? increasedEstimate : maxGas
+	}
+	const tryGasAmount = async (gas: bigint) => {
+		const multicallWithGasSpentAttempt = await simulatedMulticall(ethereumClientService, simulationState, [{ ...transactionToSimulate, gas }], block.number + 1n)
+		const simulationResultForTheTransaction = multicallWithGasSpentAttempt[multicallWithGasSpentAttempt.length - 1]
+		if (simulationResultForTheTransaction === undefined) return false
+		if (simulationResultForTheTransaction.statusCode === 'failure') return false
+		return true
+	}
+	const MAX_ATTEMPTS = 3
```
jubalm, line 148: This could sit in the `constants` file or maybe a user setting in the future?

app/ts/simulation/services/SimulationModeEthereumClientService.ts
```diff
@@ -134,8 +134,31 @@ export const simulateEstimateGas = async (ethereumClientService: EthereumClientS
...
+	for (let attempt = 0; attempt < MAX_ATTEMPTS; attempt++) {
+		// add 25% extra to account for possible gas refunds <https://eips.ethereum.org/EIPS/eip-3529>
+		// increase estimate by 10 percentage units every search attempt
+		const gasSpentEstimate = getGasEstimateWithIncrease(lastResult.gasSpent, 125n + BigInt(attempt) * 10n)
+		if (await tryGasAmount(gasSpentEstimate)) return { gas: gasSpentEstimate }
+	}
+	return {
+		error: {
+			code: ERROR_INTERCEPTOR_GAS_ESTIMATION_FAILED,
+			message: `failed to estimate gas. The transaction itself depends on the gas limit."`,
```
jubalm, line 158:
```suggestion
			message: `Failed to estimate gas. The transaction itself depends on the gas limit."`,
```

jubalm (approved):
The approach for this is something that @MicahZoltu would be keen on looking at before merging. Left minor comments

MicahZoltu:
I'm not convinced that the benefits here are worth the added pressure on the Ethereum Client.  This will make it so every single gas estimation costs 2 multicalls instead of 1, which is likely to add around 33% to total multicall burden on the node.  I would rather just add gas limit override ability to the UI than hammer the node with multiple attempts like this.

PR #830, opened 2023-12-22
with signed messages, split integer to unsigned and signed integers. Allow these numbers to be strings as well.

KillariDev:


jubalm (approved):
LGTM

PR #834, opened 2024-01-25
add eth_feeHistory

KillariDev:


app/ts/simulation/services/SimulationModeEthereumClientService.ts
```diff
@@ -918,3 +918,27 @@ export const appendSignedMessage = async (ethereumClientService: EthereumClientS
...
+	//const numberOfBlocks = request.params[0]
+	const blockTag = request.params[1]
+	const rewardPercentiles = request.params[2]
+	const newestBlock = await getSimulatedBlock(ethereumClientService, simulationState, blockTag, true)
+	const blocksToConsider = [newestBlock]
+	const oldestBlock = blocksToConsider[0]?.number
+	if (oldestBlock === undefined) throw new Error('oldest block was undefined')
+	return {
+		baseFeePerGas: blocksToConsider.map((block) => block.baseFeePerGas).filter((baseFeePerGas): baseFeePerGas is bigint => !!baseFeePerGas),
+		gasUsedRatio: blocksToConsider.map((block) => Number(block.gasUsed) / Number(block.gasLimit)).filter((gasUsedRatio): gasUsedRatio is number => !!gasUsedRatio),
```
MicahZoltu, line 933: These maps seem unnecessary when the `blocksToConsider` is always a single item tuple.  Why not just do something like
```suggestion
		baseFeePerGas: !!(newestBlock.baseFeePerGas) ? [newestBlock.baseFeePerGas] : [],
		gasUsedRatio: !!(newestBlock.gasUsed) ? [Number(newestBlock.gasUsed) / Number(block.gasLimit)] : [],
```

MicahZoltu: Same question for the more complicated `rewardPercentiles`.

app/ts/simulation/services/SimulationModeEthereumClientService.ts
```diff
@@ -918,3 +918,27 @@ export const appendSignedMessage = async (ethereumClientService: EthereumClientS
...
+
+// takes the most recent block that the application is querying and does the calculation based on that
+export const getSimulatedFeeHistory = async (ethereumClientService: EthereumClientService, simulationState: SimulationState | undefined, request: FeeHistory): Promise<EthGetFeeHistoryResponse> => {
+	//const numberOfBlocks = request.params[0]
+	const blockTag = request.params[1]
+	const rewardPercentiles = request.params[2]
+	const newestBlock = await getSimulatedBlock(ethereumClientService, simulationState, blockTag, true)
+	const blocksToConsider = [newestBlock]
+	const oldestBlock = blocksToConsider[0]?.number
+	if (oldestBlock === undefined) throw new Error('oldest block was undefined')
```
MicahZoltu, line 930: `blocksToConsider[0]` will always be `newestBlock`, why not just use that directly?

app/ts/simulation/services/SimulationModeEthereumClientService.ts
```diff
@@ -918,3 +918,27 @@ export const appendSignedMessage = async (ethereumClientService: EthereumClientS
...
+	return {
+		baseFeePerGas: blocksToConsider.map((block) => block.baseFeePerGas).filter((baseFeePerGas): baseFeePerGas is bigint => !!baseFeePerGas),
+		gasUsedRatio: blocksToConsider.map((block) => Number(block.gasUsed) / Number(block.gasLimit)).filter((gasUsedRatio): gasUsedRatio is number => !!gasUsedRatio),
+		oldestBlock,
+		...rewardPercentiles === undefined ? {} : {
+			reward: blocksToConsider.map((block) => rewardPercentiles.map((percentile) => {
+				const effectivePriority = block.transactions.map((tx) => tx.type === '1559' ? min(tx.maxPriorityFeePerGas, tx.maxFeePerGas - (block.baseFeePerGas ?? 0n)) : tx.gasPrice - (block.baseFeePerGas ?? 0n))
+				// we are using transaction.gas as a weighting factor while this should be gas used. Getting gas used requires getting receipts which is difficult
+				const gasWeights = block.transactions.map((tx) => tx.gas)
+				return calculateWeightedPercentile(effectivePriority, gasWeights, BigInt(percentile))
```
MicahZoltu, line 940: Recommend pluralizing.  I spent a while confused because I thought this was a single value, rather than a collection.
```suggestion
				const effectivePriorityFees = block.transactions.map((tx) => tx.type === '1559' ? min(tx.maxPriorityFeePerGas, tx.maxFeePerGas - (block.baseFeePerGas ?? 0n)) : tx.gasPrice - (block.baseFeePerGas ?? 0n))
				// we are using transaction.gas as a weighting factor while this should be gas used. Getting gas used requires getting receipts which is difficult
				const gasWeights = block.transactions.map((tx) => tx.gas)
				return calculateWeightedPercentile(effectivePriorityFees, gasWeights, BigInt(percentile))
```

MicahZoltu: It feels like `calculateWeightedPercentile` should take an array of tuples or objects like:
```ts
[
	[effectivePriorityFee1, gasLimit1],
	[effectivePriorityFee2, gasLimit2],
]
// or
[
	{ effectivePriorityFee: effectivePriorityFee1, gasLimit: gasLimit1 },
	{ effectivePriorityFee: effectivePriorityFee2, gasLimit: gasLimit2 },
]
```
Passing in two separate arrays that *should* line up is generally a bad pattern as it can lead to easy mistakes the compiler cannot protect against.

app/ts/utils/bigint.ts
```diff
@@ -176,3 +176,32 @@ export function isHexEncodedNumber(input: string): boolean {
 	const hexNumberRegex = /^(0x)?[0-9a-fA-F]+$/
 	return hexNumberRegex.test(input)
 }
+
+export function calculateWeightedPercentile(data: bigint[], weights: bigint[], percentile: bigint): bigint {
```
MicahZoltu, line 180: As mentioned above, this feels like it should take in a single array of type `readonly [bigint, bigint][]` or `readonly {effectivePriorityFee, gasUsed}[]` (`readonly` is less important, but good practice in general).

MicahZoltu (approved):
No blockers, but I'm not a huge fan of some of the coding style decisions made here.

app/ts/simulation/services/SimulationModeEthereumClientService.ts
KillariDev: I wanted to add a support for multiple blocks, in future

PR #835, opened 2024-01-25
Support for eth_simulateV1 (rename / use 0xeeee for eth / fix basefee -> undefined )

KillariDev:


app/ts/simulation/services/EthereumClientService.ts
```diff
@@ -198,7 +198,16 @@ export class EthereumClientService {
...
+			
+			const transactionsWithRemoveZeroPricedOnes = transactions.map((transaction) => {
+				if (transaction.type !== '1559') return transaction
+				const { maxFeePerGas, ...transactionWithoutMaxFee } = transaction
+				return {
+					...transactionWithoutMaxFee,
+					...maxFeePerGas === 0n ? {} : { maxFeePerGas }
+				}
+			})
+			return this.executionSpec383MultiCallOnlyTransactionsAndSignatures(transactionsWithRemoveZeroPricedOnes, spoofedSignatures, blockNumber, extraAccountOverrides)
```
MicahZoltu, line 210: At some point we should do a find and replace of `multicall` with `simulate` (with various casing adjustments as needed).

app/ts/utils/ethereum.ts
```diff
@@ -49,6 +49,23 @@ export interface IUnsignedTransaction1559 {
...
+	readonly maxPriorityFeePerGas: bigint
+	readonly gasLimit: bigint
+	readonly to: bigint | null
+	readonly value: bigint
+	readonly input: Uint8Array
+	readonly accessList: readonly {
+		readonly address: bigint
+		readonly storageKeys: readonly bigint[]
+	}[]
+}
```
MicahZoltu, line 67: Consider changing this to a type and just doing something like:
```ts
export type OptionalIUnsignedTransaction1559 = Omit<IUnsignedTransaction1559, 'maxFeePerGas'> & Optional<IUnsignedTransaction1559>
```
Note, above code is pseudocode to give you an idea, I'm not sure what the best way to achieve this is though.  Happy to workshop it in Discord.

MicahZoltu (approved):

app/ts/utils/ethereum.ts
KillariDev: its not actually used anywhere :D. I removed it

PR #843, opened 2024-02-12
improve RPC request caching by caching already pending values, make request limiter global and double the amount

KillariDev:


app/ts/simulation/services/EthereumJSONRpcRequestHandler.ts
```diff
@@ -1,53 +1,65 @@
 import { RpcEntry } from '../../types/rpc.js'
-import { assertIsObject } from '../../utils/typescript.js'
 import { EthereumJsonRpcRequest, JsonRpcResponse } from '../../types/JsonRpc-types.js'
 import { FetchResponseError, JsonRpcResponseError } from '../../utils/errors.js'
 import { serialize } from '../../types/wire-types.js'
 import { keccak256, toUtf8Bytes } from 'ethers'
 import { fetchWithTimeout } from '../../utils/requests.js'
+import { Future } from '../../utils/future.js'
+
+type ResolvedResponse = { responseState: 'failed', response: Response } | { responseState: 'success', responseString: string }
```
jubalm, line 9: It seems `responseString` is being passed the promise result of `response.json()` which may be misleading

jubalm (approved):

app/ts/simulation/services/EthereumJSONRpcRequestHandler.ts
KillariDev: hmm, I don't understand what you mean by this. What do you suggest?

PR #847, opened 2024-02-14
change active address/chain immediately when requested and then update simulation later

KillariDev:
- also make unexpected error error box smaller


jubalm (approved):
LGTM

PR #848, opened 2024-02-14
the calculated number is the delta that is added/subtracted

KillariDev:
I was calculating basefee of the new block wrong. That is because I just implemented the comments code instead of seeing its adding/subtracting


jubalm (approved):
Code wise looks good but might want to verify with @MicahZoltu for calculation strategy

PR #852, opened 2024-02-16
fix error display

KillariDev:
- There's a bug that if yellow error message appears on confirm transaction, it will hide send and reject buttons. Now it should be visible:
image


jubalm (approved):
Confirm fix

PR #864, opened 2024-02-20
Fix issue that results chain ID to be different on network selector and simulator

KillariDev:
- also change goerlis rich address to a new one that is rich

### DIFF
--- a/app/ts/background/background-startup.ts
+++ b/app/ts/background/background-startup.ts
@@ -116,6 +116,7 @@ export async function onContentScriptConnected(simulator: Simulator, port: brows
 }
 
 async function newBlockAttemptCallback(blockheader: EthereumBlockHeader, ethereumClientService: EthereumClientService, isNewBlock: boolean, simulator: Simulator) {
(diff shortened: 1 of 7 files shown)


app/ts/background/background-startup.ts
```diff
@@ -116,6 +116,7 @@ export async function onContentScriptConnected(simulator: Simulator, port: brows
 }
 
 async function newBlockAttemptCallback(blockheader: EthereumBlockHeader, ethereumClientService: EthereumClientService, isNewBlock: boolean, simulator: Simulator) {
+	if (ethereumClientService.getChainId() !== simulator.ethereum.getChainId()) throw new Error(`Chain Id Mismatch, node is on ${ ethereumClientService.getChainId() } while simulator is on ${ simulator.ethereum.getChainId() }`)
```
jubalm, line 119: wouldn't be noticeable, but `getChainId` function returns could be stored in a variable to reduce calls
```suggestion
  const clientChainId = ethereumClientService.getChainId()
  const simulatorChainId = simulator.ethereum.getChainId()
	if (clientChainId !== simulatorChainId) throw new Error(`Chain Id Mismatch, node is on ${ clientChainId } while simulator is on ${ simulatorChainId }`)
```

jubalm (approved):

app/ts/background/background-startup.ts
KillariDev: they are just functions that return a single variable right away so I don't think its worth to modify the oneliner to 3 lines

Issue #853, opened 2024-02-17
Optimize interceptors load on RPC when idle


KillariDev:
- see if dapps are querying interceptor and if not, go to slee

PR #867, opened 2024-02-21
Make interceptor sleep if not used

KillariDev:
fix https://github.com/DarkFlorist/TheInterceptor/issues/853


jubalm (approved):
LGTM

MicahZoltu (approved):
Code looks generally fine, but this is one of those things that I feel like is very easy to create a problem where it just stops updating and won't restart so we should be sure to test thoroughly before we release.

PR #871, opened 2024-02-23
Move Interceptor sleeping logic away from node to more higher level

KillariDev:
Now when dapp makes rpc request, we will start a timer that interceptor goes to sleep in 24 secs, unless again a new rpc query comes and again restart this timer.

Also keeping home page open refreshes the timer as well as confirm transaction view (if focused)

### DIFF
--- a/app/ts/background/background-startup.ts
+++ b/app/ts/background/background-startup.ts
@@ -2,7 +2,7 @@ import 'webextension-polyfill'
 import { defaultRpcs, getSettings } from './settings.js'
 import { handleInterceptedRequest, popupMessageHandler } from './background.js'
 import { retrieveWebsiteDetails, updateExtensionBadge, updateExtensionIcon } from './iconHandler.js'
-import { clearTabStates, getEthereumSubscriptionsAndFilters, getPrimaryRpcForChain, removeTabState, setRpcConnectionStatus, updateTabState, updateUserAddressBookEntries } from './storageVariables.js'
+import { clearTabStates, getPrimaryRpcForChain, removeTabState, setRpcConnectionStatus, updateTabState, updateUserAddressBookEntries } from './storageVariables.js'
 import { Simulator } from '../simulation/simulator.js'
 import { TabConnection, TabState, WebsiteTabConnections } from '../types/user-interface-types.js'
 import { EthereumBlockHeader } from '../types/wire-types.js'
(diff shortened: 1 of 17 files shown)


app/ts/background/background-startup.ts
```diff
@@ -186,6 +186,8 @@ async function startup() {
 		}
 	})
 
+	setInterval(async () => checkIfInterceptorShouldSleep(simulator.ethereum), 1000 )
```
jubalm, line 189: an async function in setInterval looks like a race condition, did you mean to run this synchronously?

```suggestion
	setInterval(() => checkIfInterceptorShouldSleep(simulator.ethereum), 1000 )
```

app/ts/background/windows/confirmTransaction.ts
```diff
@@ -25,25 +25,36 @@ let openedDialog: PopupOrTab | undefined = undefined
...
+	if (await updateConfirmTransactionViewWithPendingTransaction(ethereumClientService) === true) return
+	if (openedDialog) await closePopupOrTabById(openedDialog)
 	openedDialog = undefined
 }
 
+export const isConfirmTransactionFocused = async () => {
+	if (openedDialog === undefined) return false
+	const pendingTransactions = await getPendingTransactions()
+	if (pendingTransactions[0] === undefined) return false
+	const popup = await getPopupOrTabOnlyById(pendingTransactions[0].popupOrTabId)
```
jubalm, line 48: Convenience alternative

```suggestion
	const [pendingTransaction] = await getPendingTransactions()
	if (pendingTransaction === undefined) return false
	const popup = await getPopupOrTabOnlyById(pendingTransaction.popupOrTabId)
```

app/ts/components/pages/ConfirmTransaction.tsx
```diff
@@ -283,8 +283,8 @@ export function ConfirmTransaction() {
 	const [unexpectedError, setUnexpectedError] = useState<string | undefined>(undefined)
 
 	const updatePendingTransactions = (message: ConfirmTransactionDialogPendingChanged | UpdateConfirmTransactionDialog) => {
-		setPendingTransactions(message.data)
-		const firstMessage = message.data[0]
+		setPendingTransactions(message.data.pendingTransactions)
+		const firstMessage = message.data.pendingTransactions[0]
```
jubalm, line 287: Alternatively
```suggestion
		const [firstMessage] = message.data.pendingTransactions
```

jubalm (approved):
Other than a possible race condition, tested to work nicely

Issue #810, opened 2023-12-07
When MetaMask locks while Interceptor is reviewing a transaction, sending to signer just sends to /dev/null.


MicahZoltu:
1. Install Interceptor.
2. Install MetaMask.
3. Configure MetaMask to have a short idle timeout.
4. Initiate a transaction on some dapp so Interceptor pops up the signing window.
5. Wait for MetaMask to idle timeout.
6. Click button in Interceptor to send transaction to signer.
7. Notice that the Interceptor window closes and the MetaMask doesn't pop-up.

KillariDev:
fixed in https://github.com/DarkFlorist/TheInterceptor/pull/876

PR #874, opened 2024-02-27
Refactor confirm transaction

KillariDev:
- Refactor confirm transaction so that we do not wait for a promise anymore, but resolve the transactions from the local storage always
- When forward to signer request is sent to a signer, get the signers reply back to the interceptor and then return it only after that (currently we just log that this happened, but we don't do any modifications to it)
- Add `type` variable to forwarding messages to make it more clear what kind of message it is. Before there was quite confusing way to recognize it

work towards https://github.com/DarkFlorist/TheInterceptor/issues/810


app/ts/background/windows/confirmTransaction.ts
```diff
@@ -166,106 +168,71 @@ export async function openConfirmTransactionDialog(
...
+				if (alreadyOpenWindow) return alreadyOpenWindow
+				await resolvePendingTransactionsAsNoResponse(pendingTransactions, simulator, websiteTabConnections)
 			}
-			await appendPendingTransaction(pendingTransaction)
-			if (justAddToPending) await sendPopupMessageToOpenWindows({ method: 'popup_confirm_transaction_dialog_pending_changed', data: { pendingTransactions: await getPendingTransactions() } })
-				
+			return await openPopupOrTab({ url: getHtmlFile('confirmTransaction'), type: 'popup', height: 800, width: 600 })
+		}
+		const openedDialog = await getPendingTransactionWindow()
+		if (openedDialog === undefined) return false //todo add better error here, somehow we failed to create window
```
jubalm, line 192: maybe switch the comment to github issue if this is already tracked

jubalm (approved):

app/ts/background/windows/confirmTransaction.ts
KillariDev: its already handled in the most recent pr

app/ts/background/popupMessageHandlers.ts
```diff
@@ -44,12 +44,8 @@ export async function confirmRequestAccess(simulator: Simulator, websiteTabConne
...
 	const tabId = getCurrentTabId()
 	const tabs = await browser.tabs.query({ active: true, lastFocusedWindow: true })
-	if (tabs[0]?.id === undefined) {
-		return await tabId
-	}
-	if (await tabId !== tabs[0].id) {
-		saveCurrentTabId(tabs[0].id)
-	}
+	if (tabs[0]?.id === undefined) return await tabId
+	if (await tabId !== tabs[0].id) saveCurrentTabId(tabs[0].id)
```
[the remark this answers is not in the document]
KillariDev: good catch, fixed

PR #885, opened 2024-03-06
update message sign metadata when user changes metadata

KillariDev:
fixes an issue caused by https://github.com/DarkFlorist/TheInterceptor/pull/882 . Where if you updated metadata (egg accounts name) in personal sign dialog, it would not update.

### DIFF
--- a/app/ts/background/popupMessageHandlers.ts
+++ b/app/ts/background/popupMessageHandlers.ts
@@ -173,7 +173,7 @@ export async function refreshPopupConfirmTransactionMetadata(ethereumClientServi
 			}
 		})
 	}
-	const visualizerResults = await Promise.all(data.visualizerResults.map(async (x) => ({ ...x, events: await parseEvents(x.events.map((e) => ({ loggersAddress: e.loggersAddress, topics: e.topics, data: e.data })), ethereumClientService) })))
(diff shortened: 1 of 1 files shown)


app/ts/background/popupMessageHandlers.ts
```diff
@@ -153,13 +154,27 @@ export async function refreshSimulation(simulator: Simulator, settings: Settings
...
+				visualizedSimulatorState: await visualizedSimulatorStatePromise,
+				currentBlockNumber: await currentBlockNumberPromise,
+				pendingTransactionAndSignableMessages: [{
+					...first,
+					visualizedPersonalSignRequest: await visualizedPersonalSignRequestPromise, transactionOrMessageCreationStatus: 'Simulated' as const 
+				}, ...promises.slice(1)]
+			}
+		})
+	}
+	const visualizerResults = await Promise.all(data.visualizerResults.map(async (x) => ({ ...x, events: await parseEvents(x.events.map((e) => ({ loggersAddress: e.loggersAddress, topics: e.topics, data: e.data })), ethereumClientService) })))
```
jubalm, line 175: just some descriptor as x is very vague
```suggestion
	const visualizerResults = await Promise.all(data.visualizerResults.map(async (result) => ({ ...result, events: await parseEvents(result.events.map((e) => ({ loggersAddress: e.loggersAddress, topics: e.topics, data: e.data })), ethereumClientService) })))
```

jubalm (approved):

PR #887, opened 2024-03-08
Add cleanup script

jubalm:
This PR only includes the configuration and script to run [knip.dev](https://knip.dev) and does not apply the changes or fix yet.

May need some tweaking on entrypoint, looks very useful already
Screenshot 2024-03-08 at 6 18 28 PM


package.json
```diff
@@ -13,7 +13,8 @@
 		"firefox": "node -e \"const fs = require('fs'); fs.copyFileSync('app/manifestV2.json', 'app/manifest.json');\"",
 		"chrome": "node -e \"const fs = require('fs'); fs.copyFileSync('app/manifestV3.json', 'app/manifest.json');\"",
 		"prepare-inpage": "tsc --project tsconfig-inpage-create-injection-script.json && node --enable-source-maps app/inpage/js/create_injection_script.mjs && node --input-type=module -e \"import { promises as fs } from 'fs'; await fs.rm('app/inpage/js/create_injection_script.mjs')\"",
-		"inpage": "tsc --project tsconfig-inpage.json && npm run prepare-inpage && tsc --project tsconfig-inpage-output.json"
+		"inpage": "tsc --project tsconfig-inpage.json && npm run prepare-inpage && tsc --project tsconfig-inpage-output.json",
+		"cleanup": "npx knip"
```
KillariDev, line 17: could be called something else. I guess this just checks stuff but does not actually do any cleaning?

jubalm: should we do `cleanup:dryrun` and `cleanup` to straight up apply fixes?

jubalm:
Seems it doesn't recognize entrypoint from html files; would assume because we reference the built `.js` file on them so I used the raw `.ts` files of those instead

knip.json
```diff
@@ -0,0 +1,15 @@
...
+	"entry": [
+		"app/ts/addressBookRender.ts",
+		"app/ts/background-startup.ts",
+		"app/ts/backgroundServiceWorker.ts",
+		"app/ts/changeChain.ts",
+		"app/ts/confirmTransaction.ts",
+		"app/ts/interceptorAccess.ts",
+		"app/ts/popup.ts",
+		"app/ts/settingsView.ts",
+		"build/vendor.ts",
```
KillariDev, line 11:
```suggestion
		"build/vendor.mts",
```

knip.json
```diff
@@ -0,0 +1,15 @@
...
+		"build/bundler.ts"
```
KillariDev, line 12:
```suggestion
		"build/bundler.mts"
```

KillariDev (changes requested):
it is saying:
```
Unused devDependencies (1)
@types/firefox-webext-browser  package.json
```
which is wrong. If you run the cleanup you'll see that vscode will complain about `browser`

```
Unlisted dependencies (1)
@zoltu/file-copier  build/vendor.mts
```
this is also wrong? its listed in `build/package.json` ?

```
Unused dependencies (2)
@noble/curves  package.json
@noble/hashes  package.json
```
this is also wrong, they are used in the html files (these are only used in the browser)

KillariDev (approved):

PR #895, opened 2024-03-20
Remove old multicall

KillariDev:
fix https://github.com/DarkFlorist/TheInterceptor/issues/891
- also fixes issue that when depositing or withrawing WETH, no token balances were shown for weth in the swap
- also identifies if simulation state is corrupted, and recreates it


app/ts/simulation/services/EthereumClientService.ts
```diff
@@ -210,51 +187,22 @@ export class EthereumClientService {
 		] } as const
 		const unvalidatedResult = await this.requestHandler.jsonRpcRequest(call)
 		/*
-		console.log('executionSpec383MultiCall')
+		console.log('ethSimulateV1')
```
jubalm, line 190: These look like debug codes that never fires anyway, let's rid if no longer used

jubalm (approved):

app/ts/simulation/services/EthereumClientService.ts
KillariDev: its code that I need to add there everytime there's issues with eth_simulateV1, so I rather keep it in.

jubalm:
there's a type issue here `test/tests/nethermindComparison.ts`

PR #896, opened 2024-03-20
fix no tab id

KillariDev:
fix https://github.com/DarkFlorist/TheInterceptor/issues/892

### DIFF
--- a/app/ts/utils/requests.ts
+++ b/app/ts/utils/requests.ts
@@ -68,12 +68,45 @@ export async function fetchWithTimeout(resource: RequestInfo | URL, init?: Reque
 }
 
 export const safeGetTab = async (tabId: number) => {
-    const tab = await browser.tabs.get(tabId)
-    try {
+	const tab = await browser.tabs.get(tabId)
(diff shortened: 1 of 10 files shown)


app/ts/utils/requests.ts
```diff
@@ -68,12 +68,46 @@ export async function fetchWithTimeout(resource: RequestInfo | URL, init?: Reque
...
+	try {
+		const window = await browser.windows.update(tabId, updateProperties)
+		checkAndThrowRuntimeLastError()
+		return window
+	} catch (e: unknown){
+		return undefined
+	}
+}
+
+export const checkAndThrowRuntimeLastError = () => {
```
jubalm, line 110: not really an issue but was wondering about the decision to throw error here as opposed to safe parsing?

jubalm (approved):

app/ts/utils/requests.ts
KillariDev: what do you mean by safe parsing? What is should be parsed into?

Issue #584, opened 2023-07-20
interceptor_getSimulationStack wrong gas pricing returned


0xjimmy:
When comparing the net balance changes of an account, I found that the Ether spent for gas matched the price of "maxPriorityFeePerGas" and not the realized price or maxFee price.

Steps to reproduce
1. Simulation /w make me rich
2. Add another transaction, I unwrapped 5 WETH and compared the balance changes - gas * gasPrice.
```ts
window.ethereum.request({
	method: 'interceptor_getSimulationStack',
	params: ['1.0.0']
}).then(({ payload }) => {
	const netBalanceChange = BigInt(payload[1].value) - payload[1].balanceChanges.filter(x => x.address === payload[1].from).reduce((sum, bal) => sum + BigInt(bal.before) - BigInt(bal.after), 0n)
	const minusGasOnlyMaxPriority = netBalanceChange + BigInt(payload[1].maxPriorityFeePerGas) * BigInt(payload[1].gasSpent)
	const minusGasOnlyMaxFee = netBalanceChange + BigInt(payload[1].maxFeePerGas) * BigInt(payload[1].gasSpent)
	const minusGasRealized = netBalanceChange + BigInt(payload[1].realizedGasPrice) * BigInt(payload[1].gasSpent)
	console.table({ netBalanceChange, minusGasOnlyMaxPriority, minusGasOnlyMaxFee, minusGasRealized })
})
```
![image](https://github.com/DarkFlorist/TheInterceptor/assets/52896585/be52012b-5d2a-492f-9f9a-69fffcac0e59)

Here I would expect my balance change to be slightly less than 5 ETH because of gas fees, as shown.
When I try to add back the ETH spent on gas it only matches 5 ETH increase based on the maxPriorityFeePerGas price.

PR #902, opened 2024-03-22
fix getSimulatedStack

KillariDev:
fix https://github.com/DarkFlorist/TheInterceptor/issues/584
returns now (with 0 spending transaction):
image

### DIFF
--- a/app/ts/simulation/services/SimulationModeEthereumClientService.ts
+++ b/app/ts/simulation/services/SimulationModeEthereumClientService.ts
@@ -773,10 +773,11 @@ const getSimulatedTokenBalances = async (ethereumClientService: EthereumClientSe
 	if (balanceQueries.length === 0) return []
 	function removeDuplicates(queries: BalanceQuery[]): BalanceQuery[] {
 		const unique: Map<string, BalanceQuery> = new Map()
-		queries.forEach(query => {
-			const key = query.type === 'ERC20' ? `ERC20-${ query.token }-${ query.owner }` : `ERC1155-${ query.token }-${ query.owner }-${ query.tokenId }`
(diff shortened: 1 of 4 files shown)


app/ts/simulation/services/SimulationModeEthereumClientService.ts
```diff
@@ -750,10 +771,20 @@ type BalanceQuery = {
 
 const getSimulatedTokenBalances = async (ethereumClientService: EthereumClientService, transactionQueue: EthereumUnsignedTransaction[], signedMessages: readonly SignatureWithFakeSignerAddress[], balanceQueries: BalanceQuery[], blockNumber: bigint): Promise<TokenBalancesAfter> => {
 	if (balanceQueries.length === 0) return []
+	function removeDuplicates(queries: BalanceQuery[]): BalanceQuery[] {
+		const unique: Map<string, BalanceQuery> = new Map()
+		queries.forEach(query => {
+			const key = query.type === 'ERC20' ? `ERC20-${ query.token }-${ query.owner }` : `ERC1155-${ query.token }-${ query.owner }-${ query.tokenId }`
+			if (!unique.has(key)) unique.set(key, query)
+		})
+		return Array.from(unique.values())
```
jubalm, line 780: We can start using for..of for loops
```suggestion
		const unique: Map<string, BalanceQuery> = new Map()
		for (const query of queries) {
			const key = `${query.type}-${query.token}-${query.owner}${ query.type === 'ERC1155' ? `${query.tokenId}` : ''}`
			if (unique.has(key)) continue
			unique.set(key, query)
		}
		return Array.from(unique.values())
```

jubalm (approved):

PR #920, opened 2024-03-27
Ensure async functions contains promise

jubalm:
There are technical modifications here that could be an area of discussion


jubalm:
Closing as commits may break tests and would warrant it's own PR

PR #921, opened 2024-03-27
Remove unused exports

jubalm:
Removed exports from functions and variables that aren't used anywhere else in the codebase. In effect, some functions that aren't used in the same files were also removed through detecting type errors

Exceptions were made for constants file where exports were commented out instead of removing them entirely as they may be used in the future or for debugging purposes


KillariDev (changes requested):
this cannot be removed

PR #927, opened 2024-04-02
Add 2.0.0 simulation stack and fix old stack retrieval

KillariDev:


app/inpage/ts/inpage.ts
```diff
@@ -584,7 +584,7 @@ class InterceptorMessageListener {
 				&& connectSignerReply.activeAddress !== undefined && typeof connectSignerReply.activeAddress === 'string') {
 					this.currentAddress = connectSignerReply.activeAddress
 					if (connectSignerReply.metamaskCompatibilityMode && window.ethereum !== undefined) {
-						try { window.ethereum.selectedAddress = this.currentAddress } catch(error) {console.log(error)}
+						try { window.ethereum.selectedAddress = this.currentAddress } catch(error) { }
```
jubalm, line 587: why aren't we interested in this error? or did you just want to bypass lint?

jubalm (approved):

app/inpage/ts/inpage.ts
KillariDev: because it cannot be set if metamask is enabled, and that's not too information

PR #929, opened 2024-04-02
indexed fields that are too long are returned as hashes in logs

KillariDev:


app/ts/simulation/simulator.ts
```diff
@@ -87,7 +86,11 @@ export const parseEvents = async (events: readonly EthereumEvent[], ethereumClie
 			const isArray = solidityType.includes('[')
 			const verifiedSolidityType = SolidityType.safeParse(removeTextBetweenBrackets(solidityType))
 			if (verifiedSolidityType.success === false) throw new Error(`unknown solidity type: ${ solidityType }`)
-			return { paramName: paramName, typeValue: parseSolidityValueByTypePure(verifiedSolidityType.value, value, isArray) }
+			if (typeof value === 'object' && 'hash' in value) {
```
jubalm, line 89: just to be certain that value isn't null as `null` is also of type object
```suggestion
			if (typeof value === 'object' && value !== null && 'hash' in value) {
```

jubalm (approved):

PR #939, opened 2024-04-03
Consolidated linter fixes

jubalm:
The commits should make more sense on the file relationships


app/ts/simulation/services/SimulationModeEthereumClientService.ts
```diff
@@ -184,7 +184,7 @@ export const simulateEstimateGas = async (ethereumClientService: EthereumClientS
 		} as const
 	}
 	if (lastResult.status === 'failure') {
-		console.log(lastResult)
+		console.warn(`Gas estimate failure result: ${lastResult}`)
```
KillariDev, line 187: I don't think this should be a waning, as its not really warning. Application just asked for gas estimate and it couldn't be estimated

KillariDev (changes requested):

KillariDev (approved):

PR #948, opened 2024-04-07
Adds holesky, and changes URL for mainnet.

MicahZoltu:
I don't know what WETH address is on Holesky, we should figure that out before merging.


jubalm (approved):

KillariDev (changes requested):
find out Holesky Weth address

KillariDev (approved):

Issue #701, opened 2023-10-18
simulation update can read in race conditions


KillariDev:
currently when simulation is requested to be updated (eg transaction is removed), we only allow one update to be run at once (semaphore), this can end up in multiple updates being queued up at once, while user is only generally interested in the last.

make simulation state two step:
1) make very easy to update state that contains the simulation state that user wants, but nothing that requires a connection to a node
2) make slow process that always check if the fast-state has updated, if so, start a process. This requires node connection

KillariDev:
this should result in fast active address updates as well

PR #951, opened 2024-04-09
adds abort controller

KillariDev:
fixes https://github.com/DarkFlorist/TheInterceptor/issues/701
- When new simulation request comes in, cancel the pending simulation and start a new one
- also fixes if interceptor makes request and that request fails (eg to timeout), the next exactly the same request will wait for that to resolve forever ( `finally { this.pendingCache.delete(hash) }` )


app/ts/background/metadataUtils.ts
```diff
@@ -103,11 +103,11 @@ async function identifyAddressWithoutNode(address: bigint, rpcEntry: RpcNetwork
 	return undefined
 }
 
-export async function identifyAddress(ethereumClientService: EthereumClientService, address: bigint, useLocalStorage = true) : Promise<AddressBookEntry> {
+export async function identifyAddress(ethereumClientService: EthereumClientService, requestAbortController: AbortController | undefined, address: bigint, useLocalStorage = true) : Promise<AddressBookEntry> {
```
jubalm, line 106: `AbortController` can be moved as last parameter since it's optional anyway. it would required a lot of changes outside this file tho so can't do suggestions

jubalm (approved):

app/ts/background/metadataUtils.ts
KillariDev: I feel its clear that its always after EthereumClientService in every function that has it

PR #961, opened 2024-04-11
Run a formatter test

jubalm:
This is to compare the results of the formatter. Results are from running the cli directly against `app/ts` folder 

```bash
npx jubalm/bloom app/ts
```


KillariDev:
lets close this, this has been open for so long that it doesn't seem to be worked on anymore

Issue #715, opened 2023-10-26
modify make me rich to use new multicall balance override instead of tx 


KillariDev:
we should wait for deprecation of old multicall

PR #968, opened 2024-04-15
get rid of make me rich transaction and perform the same thing with state transition instead

KillariDev:
- For old simulation stack, return the make me rich transaction as the first transaction
- fix https://github.com/DarkFlorist/TheInterceptor/issues/715


jubalm (approved):
[`includesWithComparator`](https://github.com/search?q=repo:DarkFlorist/TheInterceptor%20includesWithComparator&type=code) in `app/ts/utils/typed-arrays.ts` seem to no longer be in use and can be removed as well.

Issue #969, opened 2024-04-16
add chain id -> weth mapping


KillariDev:

KillariDev:
remove weth from rpc list

PR #971, opened 2024-04-17
weth mapping

KillariDev:
fix https://github.com/DarkFlorist/TheInterceptor/issues/969


jubalm (approved):
nice! should be very useful for rpc settings

PR #972, opened 2024-04-17
use modifyobject

KillariDev:


app/ts/utils/typescript.ts
```diff
@@ -29,3 +29,8 @@ export function getWithDefault<Key, Value>(map: Map<Key, Value>, key: Key, defau
 	if (previousValue === undefined) return defaultValue
 	return previousValue
 }
+
+type Split<T> = { [K in keyof T]: { [P in K]: T[P] } }[keyof T] | Record<PropertyKey, never>
+export function modifyObject<T extends object>(original: T, subObject: NoInfer<Split<T>>): T {
```
jubalm, line 34: I may be missing some test cases for this but looks to me like `Partial<T>` would already be sufficient for type `subObject`?

```ts
export function modifyObject<T extends object>(original: T, subObject: Partial<T>) {
	return {...original, subObject }
}
```

jubalm (approved):
FWIW, typescript on this version has already shipped `satisfies` which can also be used to ensure type match 

```ts
const base = { a: 'A', b: 'B' }
// adding c will cause a type mismatch
const modified = { ...base, c: 'c' } satisfies typeof base
```

KillariDev:
> FWIW, typescript on this version has already shipped `satisfies` which can also be used to ensure type match
> 
> ```ts
> const base = { a: 'A', b: 'B' }
> // adding c will cause a type mismatch
> const modified = { ...base, c: 'c' } satisfies typeof base
> ```

how we could use it here?

jubalm:
> > FWIW, typescript on this version has already shipped `satisfies` which can also be used to ensure type match
> > ```ts
> > const base = { a: 'A', b: 'B' }
> > // adding c will cause a type mismatch
> > const modified = { ...base, c: 'c' } satisfies typeof base
> > ```
> 
> how we could use it here?

it's actually something we can use to clone the object and modify it's *existing* property without the `modifyObject` function

PR #990, opened 2024-05-07
resolve ens name hashes

KillariDev:
- resolves ens name hashes that are available in the ens name wrapper interface
- merge token events and normal events to the same array
- clean up a code bit (eg remove db file as its not used anywhere)
image


app/ts/background/metadataUtils.ts
```diff
@@ -192,3 +193,22 @@ export async function nameTokenIds(ethereumClientService: EthereumClientService,
...
 }
+
+export const extractTokenEvents = (events: readonly EnrichedEthereumEventWithMetadata[]): readonly TokenVisualizerResultWithMetadata[] =>{
+	return events.filter((tokenEvent): tokenEvent is TokenEvent => tokenEvent.type === 'TokenEvent').map((token) => token.logInformation)
+}
+
+export async function retrieveEnsNodeHashes(ethereumClientService: EthereumClientService, events: EnrichedEthereumEvents) {
+	const hashes = events.map((event) => event.type === 'ENSAddrChanged' || event.type === 'ENSAddressChanged' ? event.logInformation.node : undefined).filter((maybeNodeHash): maybeNodeHash is bigint => maybeNodeHash !== undefined)
+	const deduplicated = Array.from(new Set(hashes))
+	return Promise.all(deduplicated.map((hash) => getAndCacheEnsNodeHash(ethereumClientService, hash)))
```
jubalm, line 204: The map and filter may be more efficiently executed using `for..of`

```suggestion
	const hashes = new Set<bigint>()
	for (const event of events) {
		if (event.type !== 'ENSAddrChanged' && event.type !== 'ENSAddressChanged') continue
		hashes.add(event.logInformation.node)
	}
	return Promise.all([...hashes].map((hash) => getAndCacheEnsNodeHash(ethereumClientService, hash)))
```

jubalm (approved):

app/ts/background/metadataUtils.ts
KillariDev: created issue about this to avoid conflict

PR #996, opened 2024-05-10
resolve ens label hashes

KillariDev:
show label names in logs as well (the id=..., is originally just labelhash):
image

### DIFF
--- a/app/ts/background/metadataUtils.ts
+++ b/app/ts/background/metadataUtils.ts
@@ -7,7 +7,7 @@ import { ENS_TOKEN_WRAPPER, ETHEREUM_COIN_ICON, ETHEREUM_LOGS_LOGGER_ADDRESS, MO
 import { EthereumClientService } from '../simulation/services/EthereumClientService.js'
 import { IdentifiedAddress, itentifyAddressViaOnChainInformation } from '../utils/tokenIdentification.js'
 import { assertNever } from '../utils/typescript.js'
-import { addEnsNodeHash, addUserAddressBookEntryIfItDoesNotExist, getEnsNodeHashes, getUserAddressBookEntries } from './storageVariables.js'
+import { addEnsLabelHash, addEnsNodeHash, addUserAddressBookEntryIfItDoesNotExist, getEnsLabelHashes, getEnsNodeHashes, getUserAddressBookEntries } from './storageVariables.js'
 import { getUniqueItemsByProperties } from '../utils/typed-arrays.js'
 import { getEthereumNameServiceNameFromTokenId } from '../utils/ethereumNameService.js'
 import { defaultActiveAddresses } from './settings.js'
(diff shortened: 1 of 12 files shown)


app/ts/background/metadataUtils.ts
```diff
@@ -200,15 +200,35 @@ export const extractTokenEvents = (events: readonly EnrichedEthereumEventWithMet
...
+export const getAndCacheEnsNodeHash = async (ethereumClientService: EthereumClientService, ensNameHash: EthereumBytes32) => {
 	const currentHashes = await getEnsNodeHashes()
-	const entry = currentHashes.find((entry) => entry.nameHash === nameHash)
+	const entry = currentHashes.find((entry) => entry.nameHash === ensNameHash)
 	if (entry !== undefined) return entry
-	const name = await getEthereumNameServiceNameFromTokenId(ethereumClientService, undefined, nameHash)
-	if (name !== undefined) await addEnsNodeHash({ nameHash, name })
-	return { nameHash, name }
+	const name = await getEthereumNameServiceNameFromTokenId(ethereumClientService, undefined, ensNameHash)
+	if (name !== undefined) { //
```
jubalm, line 228:
```suggestion
	if (name !== undefined) {
```

app/ts/background/metadataUtils.ts
```diff
@@ -200,15 +200,35 @@ export const extractTokenEvents = (events: readonly EnrichedEthereumEventWithMet
...
+	// update the maping if we have new labels
+	const deduplicatedLabels = Array.from(new Set(newLabels))
+	await Promise.all(deduplicatedLabels.map(async (label) => await addEnsLabelHash(label)))
+	
+	// return the label hashes that we have now available
+	const currentLabelHashes = await getEnsLabelHashes()
+	return Array.from(new Set(labelHashesToRetrieve)).map((labelHash) => {
+		const found = currentLabelHashes.find((entry) => entry.labelHash === labelHash)
+		return { labelHash, label: found?.label }
+	})
```
jubalm, line 220: Looping the event log should be performant in getting info from it, also by initiating the reference as a `Set`, would by nature deduplicate the result.

```suggestion
	const labelHashesToRetrieve = new Set<bigint>()
	const newLabels = new Set<string>()

	for (const event of events) {
		if (event.type === 'ENSNameRenewed' || event.type === 'ENSRegistrarNameRenewed') {
			labelHashesToRetrieve.add(event.logInformation.labelHash)
		}
		if (event.type === 'ENSRegistrarNameRenewed') {
			newLabels.add(event.logInformation.name)
		}
	}

	await Promise.all([...newLabels].map(async (label) => await addEnsLabelHash(label)))

	const currentLabelHashes = await getEnsLabelHashes()
	return [...labelHashesToRetrieve].map((labelHash) => {
		const found = currentLabelHashes.find((entry) => entry.labelHash === labelHash)
		return { labelHash, label: found?.label }
	})
```

jubalm (approved):

app/ts/background/metadataUtils.ts
KillariDev: I don't think we should program these things with efficiency in mind as these are quite fast operations. I think adding to set one by one is anyway slower than sorting it once.

PR #1006, opened 2024-05-16
reverse resolution hashes

KillariDev:
- when doing ENS name resolutions, add all addresses in the stack and calculate reverse hashes and check if any nodes matches to those. If a math is found cache is for future use

image

- name changed event
- reverse claimed event


app/ts/background/metadataUtils.ts
```diff
@@ -220,15 +223,24 @@ export async function retrieveEnsLabelHashes(events: EnrichedEthereumEvents) {
 	})
 }
 
-export const getAndCacheEnsNodeHash = async (ethereumClientService: EthereumClientService, ensNameHash: EthereumBytes32) => {
-	const currentHashes = await getEnsNodeHashes()
+const addNewEnsNameEntry = async (name: string) => {
+	const [label, _] = name.split('.')
```
jubalm, line 227: Omit unused variable (underscore)
```suggestion
	const [label] = name.split('.')
```

jubalm (approved):

PR #1029, opened 2024-06-04
Fix UI issues after updating data

jubalm:
Found some UI issues, created a separate PR for clarity in review

- rpc url set to read-only on edit mode as it's an identifier
- fixed issue where the form is not updated the first time it is re-opened after saving
- used signals for rpc listing
- switched form focus to rpc name instead of url when editing

- [ ] catch error on console instead


app/ts/components/pages/SettingsView.tsx
```diff
@@ -190,39 +189,39 @@ export function SettingsView() {
...
 		</ul>
 	)
 }
 
-const RpcSummary = ({ entries, info }: { entries: RpcEntries, info: RpcEntry }) => {
+const RpcSummary = ({ info }: { info: RpcEntry }) => {
 	const networkName = getChainName(info.chainId)
+
+	// rerender form if entry is updated in the background by specifying a unique key
+	const infoHash = Object.values(info).join('')
```
KillariDev, line 223:
```suggestion
	const infoKey = Object.values(info).join('|')
```
as this is not a hash, I would not call it a hash. Its also often better to use some value to join with so values of other don't conflict with each other

app/ts/components/pages/SettingsView.tsx
```diff
@@ -231,8 +230,29 @@ const RpcSummary = ({ entries, info }: { entries: RpcEntries, info: RpcEntry })
...
 		</li>
 	)
 }
+
+export function useRpcConnectionsList() {
+	const entries = useSignal<RpcEntries>([])
+
+	const trackRpcListChanges = (message: unknown) => {
+		const maybeParsed = MessageToPopup.safeParse(message)
+		if (!maybeParsed.success) return
```
KillariDev, line 244: I think we want to throw error here as the message should not be anything else?

app/ts/components/subcomponents/ConfigureRpcConnection.tsx
```diff
@@ -34,24 +35,31 @@ function useQueryRpc() {
...
+		const { currentRpcNetwork } = await getSettings()
+
+		await sendPopupMessageToBackgroundPage({
 			method: 'popup_set_rpc_list',
-			data: [rpcEntry].concat(rpcEntries.filter(entry => entry.httpsRpc !== rpcEntry.httpsRpc))
+			data: [rpcEntry].concat(rpcEntries.value.filter(entry => entry.httpsRpc !== rpcEntry.httpsRpc))
 		})
+
+		if (currentRpcNetwork.httpsRpc !== rpcEntry.httpsRpc) return
+		console.warn(`Automatically switched to recently added or modified RPC (${rpcEntry.httpsRpc})`)
```
KillariDev, line 55:
```suggestion
		console.warn(`Automatically switched to recently added or modified RPC (${ rpcEntry.httpsRpc })`)
```

KillariDev (approved):

app/ts/components/pages/SettingsView.tsx
jubalm: now throws 'popup message is malformed' if unable to parse

app/ts/components/pages/SettingsView.tsx
jubalm: added in recent commit [92daa24](https://github.com/DarkFlorist/TheInterceptor/pull/1029/commits/92daa249f367de1ff75dedc7dea55bae91fc12a4)

app/ts/components/pages/SettingsView.tsx
KillariDev: you can make this cleaner by just using `.parse` instead of using safeParse, then you don't have to check for success as it throws with a good error right away

PR #1042, opened 2024-06-14
don't squish address icons to be smaller

KillariDev:
prevents address icon into squishing when hovered over:

image


app/ts/components/subcomponents/address.tsx
```diff
@@ -38,7 +38,7 @@ export function AddressIcon(param: AddressIconParams) {
 	if (param.logoUri !== undefined) {
 		return (
 			<AddressIconFrame isBig = { param.isBig }>
-				<img src = { param.logoUri } style = { { display: 'block', width: '1em', height: '1em' } }/>
+				<img src = { param.logoUri } style = { { display: 'block', width: '1em', 'min-width': '1em', height: '1em' } }/>
```
jubalm, line 41:
```suggestion
				<img src = { param.logoUri } style = { { display: 'block', width: '1em', minWidth: '1em', height: '1em' } }/>
```

jubalm (approved):

PR #1054, opened 2024-06-21
Re-layout Raw Transaction Information

jubalm:
Updated the layout of raw transaction for data value to be closer to it's label and with more prominent colors. Pre-formatted data (raw) will also start using the new legible font

| Screenshot 2024-06-21 at 10 40 50 AM | Screenshot 2024-06-21 at 10 46 02 AM |
|:--:|:--:|
|_new_|_existing_|

### DIFF
--- a/app/css/interceptor.css
+++ b/app/css/interceptor.css
@@ -1165,3 +1165,45 @@ summary:where(details[open] > *) {
 	text-overflow: ellipsis;
 	overflow: hidden;
 }
+
+.columnar {
+	--grid-cols: minmax(min-content,max-content) minmax(50%,1fr);
+	--gap-x: 1rem;
+	--gap-y: 0.5rem;
+	font-size: 0.875rem;
+	line-height: 1.2;
+	width: fit-content;
+	margin: 0 auto;
+}
+
+.columnar > * {
+	max-width: 100%;
+	min-width: 0;
+}
+
+.columnar dt {
+	justify-self: end;
+	text-align: right;
+	color: var(--subtitle-text-color);
+}
+
+.columnar dd {
+	justify-self: start;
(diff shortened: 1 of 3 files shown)


app/css/interceptor.css
```diff
@@ -1165,3 +1165,45 @@ summary:where(details[open] > *) {
 	text-overflow: ellipsis;
 	overflow: hidden;
 }
+
+.columnar {
+	--grid-cols: minmax(min-content,max-content) minmax(50%,1fr);
```
KillariDev, line 1170:
```suggestion
	--grid-cols: minmax(min-content, max-content) minmax(50%, 1fr);
```

app/css/interceptor.css
```diff
@@ -1165,3 +1165,45 @@ summary:where(details[open] > *) {
...
+	width: fit-content;
+	margin: 0 auto;
+}
+
+.columnar > * {
+	max-width: 100%;
+	min-width: 0;
+}
+
+.columnar dt {
```
KillariDev, line 1184: can we come up with some better name for this?

app/css/interceptor.css
```diff
@@ -1165,3 +1165,45 @@ summary:where(details[open] > *) {
...
+	justify-self: end;
+	text-align: right;
+	color: var(--subtitle-text-color);
+}
+
+.columnar dd {
```
KillariDev, line 1190: can we come up with some better name for this?

KillariDev:
Very good changes! Looks a lot better

app/css/interceptor.css
jubalm: how about `key-value-pair`?

KillariDev (approved):

PR #1059, opened 2024-06-26
Firefox font fix

jubalm:
Screenshot 2024-06-27 at 7 59 53 AM

Firefox doesn't seem to correctly load the TTF version of Inter font and seem to complain about font options being unavailable in the file. Quick fix is to replace it with WOFF2 (ironically is smaller file size)


KillariDev (approved):

jubalm:
> Screenshot 2024-06-27 at 7 59 53 AM
> Firefox doesn't seem to correctly load the TTF version of Inter font and seem to complain about font options being unavailable in the file. Quick fix is to replace it with WOFF2 (ironically is smaller file size)

I was actually wrong about this, and the browser is actually trying to load the locally installed `Inter` font (which isn't a variable font type) and Firefox complains it can't load some weights or slant values but Chrome seem to just fallback to whatever is available for that font type.

To mitigate this, I renamed the font-family name to `InterVariable` so it would, less likely, match an OS installed `Inter` font. Also, changed to use the `woff2` version for smaller file size.

Issue #1041, opened 2024-06-14
renaming address (eg, verifying contract) in eip712 message does not rename it immediately


KillariDev:
<img width="618" alt="image" src="https://github.com/DarkFlorist/TheInterceptor/assets/13102010/0eee75cb-756a-4b05-9abc-c2e7318616be">

PR #1068, opened 2024-07-08
fix pending metadata updating for signed messages

KillariDev:
fix https://github.com/DarkFlorist/TheInterceptor/issues/1041


jubalm (approved):
LGTM

Issue #328, opened 2023-04-13
show price of the token in swap view


KillariDev:
![image](https://user-images.githubusercontent.com/13102010/231696175-82f6028f-7d71-4b28-95f9-b8ceff98d5ae.png)

PR #1072, opened 2024-07-09
Add pricing information to swaps and fix bug with previous balance (+ -> -)

KillariDev:
fix https://github.com/DarkFlorist/TheInterceptor/issues/328
image


jubalm (approved):
Just noticed that we still have font size as constants in JS. Maybe we can also start to move font sizes to CSS variables like `--text-sm`, `text-lg` etc. and if needed just add or subtract to it `font-size: calc(2px + var(--text-sm));`

KillariDev:
> Just noticed that we still have font size as constants in JS. Maybe we can also start to move font sizes to CSS variables like `--text-sm`, `text-lg` etc. and if needed just add or subtract to it `font-size: calc(2px + var(--text-sm));`

added issue for it. Good catch

PR #1075, opened 2024-07-10
empty parentheses for price

KillariDev:
if there's no price estimate, this shows `( )`. Nowadays we just leave it out


jubalm (approved):
LGTM

PR #1081, opened 2024-07-19
ethSimulateV1Result  -> EthSimulateV1Result

KillariDev:


jubalm (approved):
Additionally, there's also `ethSimulateV1ParamObject` type that we may want to capitalize

PR #1082, opened 2024-07-19
add space between number and token symbol in token price

KillariDev:
- add space between number and token symbol in token price
- in raw transaction view, don't show some values in gray


app/ts/components/subcomponents/coins.tsx
```diff
@@ -218,6 +218,7 @@ type TokenWithAmountParams = TokenSymbolParams & {
 export function TokenWithAmount(param: TokenWithAmountParams) {
 	return <div style = 'width: fit-content; display: flex'>
 		<TokenAmount { ...param } />
+		<p>&nbsp;</p>
```
jubalm, line 221: Is this effectively the same as `<br/>`?

jubalm (approved):

app/ts/components/subcomponents/coins.tsx
KillariDev: no, its space (" "), <br> is line break

PR #1083, opened 2024-07-19
BFcache

KillariDev:
- Ignore BF cache error on interceptor. BF cache sends disconnect to socket anyway so we will handle disconnect there
- Inpage now connects back on bf cache return
- fix https://github.com/DarkFlorist/TheInterceptor/issues/1078


jubalm (approved):
Tested OK with Lunaria and devtools BF cache (https)

PR #1086, opened 2024-07-23
inform UI bettter on if we are retrying the connection or not, do not block poll rpc on unsupported network

KillariDev:


jubalm (approved):
LGTM

PR #1087, opened 2024-07-23
refactor error parsing in inpage

KillariDev:


app/inpage/ts/inpage.ts
```diff
@@ -496,6 +496,17 @@ class InterceptorMessageListener {
 		return
 	}
 
+	private parseRpcError = (maybeErrorObject: unknown) => {
+		if (typeof maybeErrorObject !== 'object' || maybeErrorObject === null) return new EthereumJsonRpcError(METAMASK_ERROR_BLANKET_ERROR, 'Unexpected thrown value.', { maybeErrorObject } )
```
jubalm, line 500: Is building the object with named property `maybeErrorObject` intentional? This will return an object as `{ maybeErrorObject: null }` or something like `{ maybeErrorObject: 'An error message' }`

```suggestion
		if (typeof maybeErrorObject !== 'object' || maybeErrorObject === null) return new EthereumJsonRpcError(METAMASK_ERROR_BLANKET_ERROR, 'Unexpected thrown value.', { rawError: maybeErrorObject } )
```

jubalm (approved):
The error data can use a better name for it's property in case something else is returned other than an object

PR #1089, opened 2024-07-25
primary rpc fix

KillariDev:
- when rpc is selected, promote it as primary
- when searching for a primary rpc, first search among primary's and then among non-primary's


jubalm (approved):
Thanks for this! 👍

PR #1104, opened 2024-08-05
Fix Addressbook search error on special character input

### LINKED ISSUE #1103: Interceptor throws when search input contains special characters
Image

jubalm:
Fixes #1103

- [x] remove `.toLowerCase()` transforms in regex match replacing it with `i` flag in `RegExp` constructor

### DIFF
--- a/app/ts/background/medataSearch.ts
+++ b/app/ts/background/medataSearch.ts
@@ -73,7 +73,7 @@ function concatArraysUniqueByAddress<T>(addTo: readonly (T & { address: bigint }
 }
 
 async function filterAddressBookDataByCategoryAndSearchString(addressBookCategory: AddressBookCategory, searchString: string | undefined): Promise<AddressBookEntries> {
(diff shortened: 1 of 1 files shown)


KillariDev (commented):
can we find a function that does the sanitasion for regexp instead of trying to list all cases?

app/ts/background/medataSearch.ts
```diff
@@ -73,8 +73,9 @@ function concatArraysUniqueByAddress<T>(addTo: readonly (T & { address: bigint }
 }
 
 async function filterAddressBookDataByCategoryAndSearchString(addressBookCategory: AddressBookCategory, searchString: string | undefined): Promise<AddressBookEntries> {
+	const unicodeEscapeString = (input: string) => `\\u{${input.charCodeAt(0).toString(16)}}`
```
KillariDev, line 76:
```suggestion
	const unicodeEscapeString = (input: string) => `\\u{ ${ input.charCodeAt(0).toString(16) } }`
```

KillariDev (approved):

KillariDev (approved):

PR #1129, opened 2024-09-18
add square brackets to fuses logs

KillariDev:


app/ts/components/simulationExplaining/Transactions.tsx
```diff
@@ -345,13 +345,14 @@ function NonTokenLogEvent(params: NonTokenLogEventParams) {
...
 					return <>
 						<p style = { textStyle } class = 'paragraph'> { `${ arg.paramName } =` }&nbsp;</p>
-						<EnsNamedHashComponent type = 'labelHash' nameHash = { params.nonTokenLog.logInformation.labelHash.labelHash } name = { params.nonTokenLog.logInformation.labelHash.label } editEnsNamedHashCallBack = { params.editEnsNamedHashCallBack }/>	
+						<EnsNamedHashComponent type = 'labelHash' nameHash = { params.nonTokenLog.logInformation.labelHash.labelHash } name = { params.nonTokenLog.logInformation.labelHash.label } editEnsNamedHashCallBack = { params.editEnsNamedHashCallBack }/>
 					</>
 				}
 				if (arg.paramName === 'fuses' && 'logInformation' in params.nonTokenLog && 'fuses' in params.nonTokenLog.logInformation) {
 					return <>
-						<p style = { textStyle } class = 'paragraph'> { `${ arg.paramName } =` }&nbsp;</p>
+						<p style = { textStyle } class = 'paragraph'> { `${ arg.paramName } = [` }</p>
```
jubalm, line 353: this is one of those areas where were inlining multiple `<p>` tags (where they typically default to break lines) and can be clunky to work with at times. 

i can't really add suggest lines here as these would be multi-line changes

jubalm (approved):

app/ts/components/simulationExplaining/Transactions.tsx
KillariDev: hmm, you can copy the lines and write a comment, it doesn't need to be a code suggestion supported by github

jubalm: ok confirm this wasn't the case here

Issue #1099, opened 2024-07-31
if transaction is first failing, but then updates to be succesfull, the button is missing


KillariDev:
<img width="666" alt="image" src="https://github.com/user-attachments/assets/8de4e2b7-efc2-4d9d-9d6d-cbc9818098b8">

PR #1135, opened 2024-09-23
fix transaction simulation state update

KillariDev:
fix https://github.com/DarkFlorist/TheInterceptor/issues/1099

### DIFF
--- a/app/ts/components/pages/ConfirmTransaction.tsx
+++ b/app/ts/components/pages/ConfirmTransaction.tsx
@@ -325,6 +325,17 @@ type ModalState =
 	{ page: 'editEns', state: EditEnsNamedHashWindowState } |
 	{ page: 'noModal' }
 
+type RejectButtonParams = {
+	onClick: () => void
+}
+const RejectButton = ({ onClick }: RejectButtonParams) => {
+	return <div>
+		<button className = 'button is-primary is-danger button-overflow dialog-button-left' onClick = { onClick } >
+			{ 'Reject' }
+		</button>
+	</div>
+}
+
 type ButtonsParams = {
 	currentPendingTransactionOrSignableMessage: PendingTransactionOrSignableMessage | undefined
 	reject: () => void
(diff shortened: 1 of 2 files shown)


app/ts/components/pages/ConfirmTransaction.tsx
```diff
@@ -320,11 +320,54 @@ const WebsiteErrors = ({ website, websiteSocket, simulationMode }: NetworkErrorP
...
+	{ page: 'editEns', state: EditEnsNamedHashWindowState } |
 	{ page: 'noModal' }
 
+type ButtonsParams = {
+	currentPendingTransactionOrSignableMessage: PendingTransactionOrSignableMessage | undefined
+	reject: () => void
+	approve: () => void
+	confirmDisabled: boolean
+}
+function Buttons({ currentPendingTransactionOrSignableMessage, reject, approve, confirmDisabled }: ButtonsParams) {
```
jubalm, line 334: actually quite curious why `onlyReject`'s `<button>` element here is wrapped with a "flexed" `div` but it's the only child of that div?

Anyways, the `rejectOnly` component can also be moved outside this component

```ts
// 	const onlyReject = <div style = 'display: flex; flex-direction: row;'> ...
const RejectButton = ({ onClick }: { onClick: () => void }) => <div style = 'display: flex; flex-direction: row;'>
		<button className = 'button is-primary is-danger button-overflow dialog-button-left' onClick = { onClick } >
			{ 'Reject' }
		</button>
	</div>

function Buttons({ currentPendingTransactionOrSignableMessage, reject, approve, confirmDisabled }: ButtonsParams) {
//... then later
	if (identified === undefined) return <RejectButton onClick = { reject } />
```

jubalm (approved):

app/ts/components/pages/ConfirmTransaction.tsx
KillariDev: the flex style is just remainer when there used to be two buttons. I removed the styling here and also moved the reject out

PR #1141, opened 2024-09-25
Invalidate block cache if its too old

KillariDev:
Because interceptor goes sleep, it's possible that we query really old block and think its the newest block, this results us to make eth_simulate queries against blocks that are older than 256 blocks, resulting into a query that fails


app/ts/simulation/services/EthereumClientService.ts
```diff
@@ -35,7 +35,12 @@ export class EthereumClientService {
 	public readonly getNewBlockAttemptCallback = () => this.newBlockAttemptCallback
 	public readonly getOnErrorBlockCallback = () => this.onErrorBlockCallback
 
-	public getCachedBlock = () => this.cachedBlock
+	public getCachedBlock = () => {
+		if (this.cachedBlock === undefined) return undefined
+		// if the block is older than 100 block intervals, invalidate cache
+		if ((Date.now() - this.cachedBlock.timestamp.getTime() * 1000) > TIME_BETWEEN_BLOCKS * 100) return undefined
```
jubalm, line 41: the block cache size could be a constant. ex. `MAX_BLOCK_CACHE`

jubalm (approved):

PR #1159, opened 2024-10-22
Replace ETH UI with Inline Card

jubalm:
Replaced Eth symbol to use UI same with address

| Screenshot 2024-10-22 at 1 53 08 PM | Screenshot 2024-10-22 at 1 52 56 PM |
|--|--|

New Feature:
- [x] edit button will only appear if `onEditClicked` callback is provided as props


MicahZoltu (approved):
We can merge this as-is, but we should also add (in a separate PR) the ability to disable copy button for ETH.

PR #1165, opened 2024-10-27
handle null getblock return

KillariDev:
- fix bug that we are querying block number +1 while we should query the current block number.
- Handle case where the RPC returns NULL as the block


test/tests/nethermindComparison.ts
```diff
@@ -92,6 +92,7 @@ export async function main() {
 
 		should('getBlock with true aligns with Nethermind', async () => {
 			const block = await getSimulatedBlock(ethereum, undefined, simulationState, blockNumber, true)
+			if (block === null) throw new Error('block was null')
```
jubalm, line 95:
```suggestion
			if (block === null) throw new Error('Block was null')
```

jubalm (approved):
Funtypes issue fixed with this change 👍

PR #1177, opened 2024-10-30
add support for optimism

KillariDev:
add https://optimism.blockpi.network/v1/rpc/public as rpc to test


jubalm (approved):
Looking good, did some tests with `https://optimism.blockpi.network/v1/rpc/public` RPC

PR #1179, opened 2024-10-30
make micah poor again

KillariDev:
if we are in signing mode, don't make any address rich


app/ts/simulation/services/SimulationModeEthereumClientService.ts
```diff
@@ -224,14 +224,19 @@ export const appendTransaction = async (ethereumClientService: EthereumClientSer
 		const signed = mockSignTransaction(transaction.transaction)
 		return simulationState === undefined ? [signed] : simulationState.simulatedTransactions.map((x) => x.preSimulationTransaction.signedTransaction).concat([signed])
 	}
+	const getMakeMeRichAddress = async () => {
+		if (typeof browser === 'undefined') return simulationState?.addressToMakeRich // if we are not running in browser (tests)
+		if (simulationState === undefined) return undefined // we are not simulation, don't make anyone rich
```
jubalm, line 229: can swap the order of execution so that `simulationState` will not need to optionally chain `addressToMakeRich`

jubalm (approved):

PR #1180, opened 2024-11-01
Big address component update

jubalm:
Replace big address component

| Screenshot 2024-11-05 at 3 25 53 PM | Screenshot 2024-11-05 at 3 26 07 PM |
|--|--|
| *preserve functions* | *interchangeable actions (copy/edit)* |


app/ts/components/subcomponents/address.tsx
```diff
@@ -53,73 +54,38 @@ type BigAddressParams = {
...
+	const renameAddressAction:TextAction = {
+		label: 'Edit',
+		icon: () => <EditIcon />,
+		onClick: () => params.addressBookEntry && params.renameAddressCallBack(params.addressBookEntry)
+	}
 
-const AddressTitle = ({ content, useLegibleFont  }: { content: string, useLegibleFont?: boolean }) => {
-	return <p class = {  `title is-5 is-spaced address-text noselect nopointer${ useLegibleFont ? ' text-legible' : '' }` }>{ content }</p>
-}
+	const labelConfig:ActionableTextProps = {
```
KillariDev, line 71:
```suggestion
	const labelConfig: ActionableTextProps = {
```

app/ts/components/subcomponents/address.tsx
```diff
@@ -53,73 +54,38 @@ type BigAddressParams = {
...
+		displayText: title,
+		action: !params.noEditAddress && title !== addrString ? renameAddressAction : undefined
+	}
+
+	const noteConfig:ActionableTextProps = {
```
KillariDev, line 76:
```suggestion
	const noteConfig: ActionableTextProps = {
```

KillariDev (commented):
image

when changing active address it doesn't highlight the whole button as blue anymore. I guess this is because there's a background color that was not there before?

KillariDev (changes requested):
looks good but this active address selector looks ugly as the whole button doesn't highlight, there's also copy buttons there that do not work (it just selects the address and copy doesnt actually copy)
image

PR #1181, opened 2024-11-02
separate addressbook for each chain

KillariDev:
- separate addressbook for each chain
- still missing: universal addresses
- still missing: address metadata for different chains
image


jubalm (approved):
This breaks current addressbook categories but should be amended in #1182

PR #1182, opened 2024-11-05
add universal addresses

KillariDev:
- now addresses can be universal.
- still need to conduct some more testing with this


jubalm:
Looks like chain selector highlights by Chain ID and matches 2 selections
Screenshot 2024-11-14 at 10 07 50 AM

jubalm:
Adding an address in "All Chains" first would allow you to add the same address on one of the other chains.
| Screenshot 2024-11-14 at 10 23 21 AM | Screenshot 2024-11-14 at 10 23 17 AM |
|--|--|
| _All Chains_ | _Mainnet_ |

app/ts/AddressBook.tsx
```diff
@@ -167,57 +167,70 @@ function AddressBookEntryCard({ removeEntry, renameAddressCallBack, ...entry }:
...
-				const chainId = currentRpcNetwork.peek()?.chainId
+				const chainId = currentChain.peek()?.chainId
 				if (chainId !== undefined) sendQuery()
 				return
 			}
 			if (parsed.method === 'popup_settingsUpdated') return sendPopupMessageToBackgroundPage({ method: 'popup_requestSettings' })
 			if (parsed.method === 'popup_requestSettingsReply') {
 				rpcEntries.value = parsed.data.rpcEntries
-				const prevCurrentNetwork = currentRpcNetwork.peek()
+				const prevCurrentNetwork = currentChain.peek()
```
jubalm, line 211: `prevCurrentNetwork` ("Previous" "Current") sound opposing. Something like `currentChainSnapshot`, `previousChain` or `previousNetwork` may work

app/ts/background/background-startup.ts
```diff
@@ -58,11 +58,29 @@ async function migrateAddressInfoAndContacts() {
...
-		await updateUserAddressBookEntries((previousEntries) => getUniqueItemsByProperties(updated.concat(previousEntries), ['address']))
+		await updateUserAddressBookEntriesV2Old((previousEntries) => getUniqueItemsByProperties(updated.concat(previousEntries), ['address']))
 		await browserStorageLocalRemove(['userAddressBookEntries'])
 	}
 }
-migrateAddressInfoAndContacts()
+async function migrateAddressInfoAndContactsFromV2ToV3() {
+	const userAddressBookEntries = (await browserStorageLocalGet(['userAddressBookEntriesV2'])).userAddressBookEntriesV2
+	const convertOldActiveAddressToAddressBookEntry = (entry: AddressBookEntry): AddressBookEntry => {
+		if ('chainId' in entry && entry.chainId !== undefined) return entry
```
jubalm, line 68: `entry` is already an `AddressBookEntry` type which should already guarantee the property `chainId`
```suggestion
		if (entry.chainId !== undefined) return entry
```

app/ts/components/subcomponents/ChainSelector.tsx
```diff
@@ -27,17 +28,49 @@ export function ChainSelector(params: ChainSelectorParams) {
...
+}
+
+interface ChainSelectorParams {
+	chainId: ReadonlySignal<ChainIdWithUniversal>
+	rpcEntries: Signal<RpcEntries>
+	changeChain: (entry: ChainEntry) => void
+}
+
+export function ChainSelector(params: ChainSelectorParams) {
+	const [isOpen, setIsOpen] = useState(false)
```
jubalm, line 49: Wouldn't be much to switch to signal as the entire component already is utilizing it already anyway

```ts
// const [isOpen, setIsOpen] = useState(false)
const dropdownOpen = useSignal(false)
```

jubalm (approved):

app/ts/AddressBook.tsx
KillariDev: the variable being referred is "currentNetwork" so snapshot and chain, chain and network are wrong. I guess `activeNetwork` would be better as we use `activeAddress` as well

KillariDev: I added issue for this

PR #1187, opened 2024-11-08
don't check duplicates from dark florist metadata unless we are on chain 1 or allchains

KillariDev:
there's a bug that interceptor complains about duplicate address when we are on different chain than mainnet


lattaai4:
Hello, I tried to fix the issue

This is what I did:

Modified the addUserAddressBookEntryIfItDoesNotExist function to consider chain ID when checking for duplicate addresses. This allows users to add the same address for different chains without triggering a duplicate address complaint.

I also created pull request: https://github.com/DarkFlorist/TheInterceptor/pull/1188


> [!CAUTION]
> Disclaimer: This fix was created by Latta AI and you should never merge before you check the correctness of generated code!
>
> The fix provided by Latta AI might not be complete and it can serve as an inspiration.

---
This bug was fixed for free by Latta AI - https://latta.ai/ourmission

If you no longer want Latta AI to attempt fixing issues on your repository, you can block this account.

jubalm (approved):

PR #1191, opened 2024-11-14
Fix duplicate entry in chain list

jubalm:
Proposing this change to only have a single dropdown option for each chain id in Chain selector. This should fix the issue mentioned on #1182 

| Screenshot 2024-11-14 at 10 07 50 AM |
|--|
| _duplicate entry is highlighted as well_ |


app/ts/components/ui-utils.tsx
```diff
@@ -177,5 +178,11 @@ export const getAddressBookEntryOrAFiller = (addressMetaData: readonly AddressBo
 }
 
 export const rpcEntriesToChainEntriesWithAllChainsEntry = (rpcEntries: RpcEntries): readonly ChainEntry[] => {
-	return [ ...rpcEntries.map((rpcEntry) => ({ name: rpcEntry.name, chainId: rpcEntry.chainId })), { name: 'All Chains', chainId: 'AllChains' as const }]
+	const entries = rpcEntries.map(({ chainId }): [string, ChainEntry] => {
+		const chainIdString = String(Number(chainId))
```
KillariDev, line 182:
```suggestion
		const chainIdString = chainId.toString()
```

app/ts/components/ui-utils.tsx
```diff
@@ -177,5 +178,11 @@ export const getAddressBookEntryOrAFiller = (addressMetaData: readonly AddressBo
...
+		return [chainIdString, { chainId, name: CHAIN_NAMES.get(chainIdString) || chainIdString }]
```
KillariDev, line 183:
```suggestion
		return [chainIdString, { chainId, name: CHAIN_NAMES.get(chainIdString) || `Chain ID: ${ chainIdString }` }]
```

KillariDev (approved):

PR #1197, opened 2024-11-22
Fixed build error when app/js folder does not exist

jubalm:
On a fresh checkout, `app/js` does not exist and build script can error without creating it first. Additionally added `node:` prefix to node built-in imports for future proofing. 

FWIW: Tried running bundler script (`bundler.mts`) in Node experimental and Deno with the `node:` prefix and they work without the need for temporary file outputs.

### DIFF
--- a/build/bundler.mts
+++ b/build/bundler.mts
@@ -1,6 +1,6 @@
-import * as path from 'path'
-import * as url from 'url'
-import { promises as fs } from 'fs'
+import * as path from 'node:path'
+import * as url from 'node:url'
+import { promises as fs } from 'node:fs'
 
 const directoryOfThisFile = path.dirname(url.fileURLToPath(import.meta.url))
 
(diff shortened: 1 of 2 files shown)


build/bundler.mts
```diff
@@ -62,12 +62,21 @@ async function* getFiles(topDir: string): AsyncGenerator<string, any, undefined>
 	}
 }
 
+async function ensureDirectoryExists(dir: string) {
+	try {
+		await fs.access(dir)
+	} catch {
+		await fs.mkdir(dir, { recursive: true })
```
KillariDev, line 69: do we need this to be recursive?

jubalm: in the current state of the build process it's not indeed required so i removed it in [a79243b](https://github.com/DarkFlorist/TheInterceptor/pull/1197/commits/a79243ba097b91367f5a4dfb8fc3b1acd9ee81fe)

PR #1202, opened 2024-11-25
when removing address, remove it only for that chain id

KillariDev:


app/ts/background/popupMessageHandlers.ts
```diff
@@ -83,7 +83,10 @@ export async function changeMakeMeRich(ethereumClientService: EthereumClientServ
 }
 
 export async function removeAddressBookEntry(simulator: Simulator, websiteTabConnections: WebsiteTabConnections, removeAddressBookEntry: RemoveAddressBookEntry) {
-	await updateUserAddressBookEntries((previousContacts) => previousContacts.filter((contact) => contact.address !== removeAddressBookEntry.data.address))
+	await updateUserAddressBookEntries((previousContacts) => previousContacts.filter((contact) =>
+		!(contact.address === removeAddressBookEntry.data.address
+		&& (contact.chainId === removeAddressBookEntry.data.chainId || (contact.chainId === undefined && removeAddressBookEntry.data.chainId === 1n))))
+	)
```
jubalm, line 89: while code works perfectly fine, it may be more readable with the conditions broken down a bit
```suggestion
	await updateUserAddressBookEntries((previousContacts) => {
		const shouldKeepContact = (contact: typeof previousContacts[number]) => {
			const isNotTheAddressToRemove = contact.address !== removeAddressBookEntry.data.address
			const isNotWithinTheSameChain = contact.chainId !== removeAddressBookEntry.data.chainId
			const isChainIdNotUniversal = !(contact.chainId === undefined && removeAddressBookEntry.data.chainId === 1n)

			if (isNotTheAddressToRemove) return true
			return isNotWithinTheSameChain && isChainIdNotUniversal
		}

		return previousContacts.filter(shouldKeepContact)
	})
```

jubalm (approved):

app/ts/background/popupMessageHandlers.ts
KillariDev: hmm I don't think this makes it cleaner as those variable names just say directly what the statement is

PR #1207, opened 2024-11-28
remove use states from confirmTransaction

KillariDev:
- there's a bug that when as a user you try to force send transaction that is failing, interceptor doesn't let you as `useComputed` doesn't update the buttons disabled state


MicahZoltu (approved):
I'm a fan of finally moving to Signals!

PR #1208, opened 2024-11-28
merge margins together (instead of mixing paddings and margings), allow importance box to scale smaller

KillariDev:
- there's mixed uses of paddings and margins for errors. This PR separates them by margin only so the margins get merged together
- allow importance box to be smaller than 400px on small window

### DIFF
--- a/app/css/interceptor.css
+++ b/app/css/interceptor.css
@@ -743,8 +743,7 @@ svg.spinner > circle {
 	padding: 10px;
 	margin: auto;
 	margin-bottom: 10px;
-	min-width: 400px;
-	max-width: fit-content;
+	max-width: min(400px, 100%);
 	color: var(--text-color)
 }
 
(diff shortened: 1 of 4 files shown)


app/css/interceptor.css
```diff
@@ -743,8 +743,7 @@ svg.spinner > circle {
 	padding: 10px;
 	margin: auto;
 	margin-bottom: 10px;
-	min-width: 400px;
-	max-width: fit-content;
+	max-width: min(400px, 100%);
```
MicahZoltu, line 746: I'm not a fan of fixed pixel count max width.  Why is this necessary?  What does it look like on a very small screen?  What about a very big screen?  What happens in both cases if you just set it to 100%?

MicahZoltu (approved):
Approved because we are trying to close out issues/PRs, but I really wolud prefer to see the max-width removed.

app/css/interceptor.css
KillariDev: yeah, `fit-content` is actually what we want. It also fix the issue I had yesterday with this

KillariDev: and get rid of the min-width that we used to have

PR #1214, opened 2024-11-29
seems like mint can also be null on Optimism based chains

KillariDev:


MicahZoltu (approved):

MicahZoltu:
I'm surprised that you don't need to update code anywhere else.

PR #1216, opened 2024-12-04
don't add async message listeners

KillariDev:
apparently you shouldn't have async message listeners according to this: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/onMessage

### DIFF
--- a/app/ts/components/App.tsx
+++ b/app/ts/components/App.tsx
@@ -184,7 +184,7 @@ export function App() {
 			setWebsiteAccess(settings.websiteAccess)
 		}
 
-		const popupMessageListener = async (msg: unknown) => {
+		const popupMessageListener = (msg: unknown) => {
 			const maybeParsed = MessageToPopup.safeParse(msg)
 			if (!maybeParsed.success) return // not a message we are interested in
 			const parsed = maybeParsed.value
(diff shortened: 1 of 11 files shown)


app/ts/components/App.tsx
```diff
@@ -197,7 +197,7 @@ export function App() {
 				}
 				case 'popup_websiteIconChanged': return setTabConnection(parsed.data)
 				case 'popup_new_block_arrived': {
-					await sendPopupMessageToBackgroundPage({ method: 'popup_refreshHomeData' })
+					sendPopupMessageToBackgroundPage({ method: 'popup_refreshHomeData' })
```
MicahZoltu, line 200: Every call to an async function from a synchronous function should wrap the async call in `catchAllErrorsAndCall` or some similar function with a form like:
```ts
async function handleAllAsyncExceptions(func: () => void): void {
	try {
		await func()
	} catch (error) {
		// TODO: handle all possible errors, don't let anything bubble
	}
}
```
```suggestion
					handleAllAsyncExceptions(() => sendPopupMessageToBackgroundPage({ method: 'popup_refreshHomeData' }))
```

MicahZoltu (approved):

PR #1218, opened 2024-12-09
Update extension tooltip with correct icon info

KillariDev:


app/ts/background/backgroundUtils.ts
```diff
@@ -99,6 +99,21 @@ export async function setExtensionIcon(details: browser.action._SetIconDetails)
...
+		if (manifest.manifest_version === 2) {
+			await browser.browserAction.setTitle(details)
+		} else {
+			await browser.action.setTitle(details)
+		}
+		checkAndThrowRuntimeLastError()
+	} catch {
+		console.warn('failed to set extension title')
+		console.warn(details)
+	}
```
MicahZoltu, line 114: Can you move this try/catch block to `setInterceptorIcon` or whatever function calls that?  Generally speaking, I find it preferable to put generalized error handling as high up the stack as possible, usually where there is an async function given as a sync callback.

MicahZoltu (approved):

PR #1220, opened 2024-12-09
fix website access search type error

KillariDev:


app/ts/background/websiteAccessSearch.ts
```diff
@@ -44,9 +44,7 @@ function calculateWebsiteAccessScore(entry: WebsiteAccess, query: string): Searc
 	const titleMatch = entry.website.title ? computeSearchMatch(query, entry.website.title.toLowerCase()) : undefined
 	const addressMatches = entry.addressAccess?.map((addr: WebsiteAddressAccess) => computeSearchMatch(query, addressString(addr.address).toLowerCase())) || []
 
-	const bestAddressMatch = addressMatches.length > 0 ? addressMatches.reduce(selectLongerMatch) : undefined
-
-	const bestResult = [urlMatch, titleMatch, bestAddressMatch]
+	const bestResult = [urlMatch, titleMatch, ...addressMatches]
```
MicahZoltu, line 47: Why is there only one url match, one title match, but many address matches?  We don't want many of each?

jubalm: the function this piece of code runs on calculates scores on a single entry in the access list, with the data structure having only a single title, a single url and multiple addresses. the prior implementation (`bestAddressMatch`) took only a single address to be matched against the rest of the list of entries

jubalm:
Updated the test [here](https://github.com/DarkFlorist/TheInterceptor/pull/1222) to match these changes @KillariDev

jubalm (approved):

app/ts/background/websiteAccessSearch.ts
KillariDev: the older version also had wrong type for "bestAddressMatch", so it worked wrong

PR #1224, opened 2024-12-10
estimate gnosis safe message with all gas left in the block

KillariDev:
this will now make eth_simulateV1 call with two blocks where the second block will contain all the calls not part of the simulation stack.

This fixes the gnosis safe issue partly but not completely (still the proxy call trickery is needed


app/ts/simulation/services/EthereumClientService.ts
```diff
@@ -233,16 +233,22 @@ export class EthereumClientService {
...
+			time: new Date(parentBlock.timestamp.getTime() + (index + 1) * 12 * 1000),
+			gasLimit: parentBlock.gasLimit,
+			feeRecipient: parentBlock.miner,
+			baseFeePerGas: parentBlock.baseFeePerGas === undefined ? 15000000n : parentBlock.baseFeePerGas
+		})
+
+		const [firstBlocksTransactions, ...restofTheBlocks] = transactionsWithRemoveZeroPricedOnes
+		if (firstBlocksTransactions === undefined) throw new Error('No blocks specified for simulateTransactionsAndSignatures')
+
+		// add stateOverrides only to the first block (ecrecover overrides, make me rich and such)
```
MicahZoltu, line 248: State overrides carry over to subsequent blocks?

MicahZoltu (approved):

app/ts/simulation/services/EthereumClientService.ts
KillariDev: yeah, they modify state and state stays modified

PR #1225, opened 2024-12-11
add gnosis safe delegate calling

KillariDev:


app/ts/background/background.ts
```diff
@@ -134,17 +135,37 @@ export const simulateGovernanceContractExecution = async (pendingTransaction: Pe
...
+
+		// Call: 0x0, DelegateCall: 0x1
+		// https://github.com/safe-global/safe-smart-account/blob/main/contracts/libraries/Enum.sol
+		const isDelegateCall = gnosisSafeMessage.message.message.operation === 0x1n
+		const ORIGINAL_GNOSIS_SAFE = 0x0000000000000000000000000000000000920515n
+		/*
+		If we are doing a normal call, we send a transaction from gnosis safe to the callable address
+		If we are doing a delegate call, we do a following operation:
+			1) move safe (gnosisSafeMessage.verifyingContract.address) -> ORIGINAL_GNOSIS_SAFE
+			2) replace safe with GnosisSafeProxyProxy (a contract that delegates everything to ORIGINAL_GNOSIS_SAFE, except calls to `delegateCallExecute`)
```
MicahZoltu, line 148: I think you can get rid of step 1.  Just replace the existing proxy with our proxy, which does everything the original proxy did *except* has a handler for `delegateCallExecute(...)`.

app/ts/background/background.ts
```diff
@@ -134,17 +135,37 @@ export const simulateGovernanceContractExecution = async (pendingTransaction: Pe
 export const simulateGnosisSafeMetaTransaction = async (gnosisSafeMessage: VisualizedPersonalSignRequestSafeTx, simulationState: SimulationState | undefined, ethereumClientService: EthereumClientService, tokenPriceService: TokenPriceService): Promise<DistributiveOmit<SimulateExecutionReplyData, 'transactionOrMessageIdentifier'>> => {
 	const returnError = (errorMessage: string) => ({ success: false as const, errorType: 'Other' as const, errorMessage })
 	try {
-		const transactionWithoutGas = {
+		const delegateCallExecuteInterface = new ethers.Interface(['function delegateCallExecute(address, bytes memory) payable external returns (bytes memory)'])
+
+		// Call: 0x0, DelegateCall: 0x1
+		// https://github.com/safe-global/safe-smart-account/blob/main/contracts/libraries/Enum.sol
+		const isDelegateCall = gnosisSafeMessage.message.message.operation === 0x1n
+		const ORIGINAL_GNOSIS_SAFE = 0x0000000000000000000000000000000000920515n
```
MicahZoltu, line 143: Does this number mean anything?  If so we should include a comment as to what.  If not, we should include a comment saying it is just an arbitrary chosen value.

MicahZoltu (approved):

app/ts/background/background.ts
KillariDev: its "gnosis"
9=G
2=N
0=O
5=S
1=I

MicahZoltu: Yeah, you definitely need a comment for that.  Why not UTF-8 GNOSIS at least instead of leet speak?

KillariDev: just what came to first to my mind. The criteria for the number is not very restrictive:
1) it should not be the same we use for ecrecover spoofing
2) it should not conflict with precompiles
3) it should be something its unlikely someone has private key for or something you can deploy contract

KillariDev: I feel also leet speak is more human readable when you look at the number vs UTF-8 encoding. But maybe thats just me :D

PR #1228, opened 2024-12-12
lint

KillariDev:


app/ts/background/popupMessageHandlers.ts
```diff
@@ -272,9 +272,8 @@ export async function refreshPopupConfirmTransactionSimulation(simulator: Simula
 			case 'Transaction': {
 				if (transactionToSimulate.success) {
 					return { ...transactionOrMessage, transactionToSimulate, simulationResults: refreshMessage, transactionOrMessageCreationStatus: 'Simulated' }
-				} else {
-					return { ...transactionOrMessage, transactionToSimulate, simulationResults: refreshMessage, transactionOrMessageCreationStatus: 'FailedToSimulate' }
 				}
+				return { ...transactionOrMessage, transactionToSimulate, simulationResults: refreshMessage, transactionOrMessageCreationStatus: 'FailedToSimulate' }
```
MicahZoltu, line 276: Why not instead do this?
```ts
return { ...transactionOrMessage, transactionToSimulate, simulationResults: refreshMessage, transactionOrMessageCreationStatus: transactionToSimulate.success ? 'Simulated' : 'FailedToSimulate' }
```

app/ts/components/subcomponents/AutosizingInput.tsx
```diff
@@ -55,15 +55,12 @@ function ParsedInput<T>(model: ParsedInputModel<T>) {
...
 }
 
 function Input<T>(model: UnparsedInputModel | ParsedInputModel<T>) {
-	if ('tryParse' in model && model.tryParse) {
-		return <ParsedInput { ...model }/>
-	} else {
-		return <ParsedInput { ...model } value = { new OptionalSignal(model.value)} sanitize = { model.sanitize || (x => x) } tryParse = { value => ({ ok: true, value }) } serialize = { x => x || '' }/>
-	}
+	if ('tryParse' in model && model.tryParse) return <ParsedInput { ...model }/>
+	return <ParsedInput { ...model } value = { new OptionalSignal(model.value)} sanitize = { model.sanitize || (x => x) } tryParse = { value => ({ ok: true, value }) } serialize = { x => x || '' }/>
```
MicahZoltu, line 63: I like the if/else better, since neither code path is an "early exit", they are just two equally valid code paths.

I like one-line returns when it is a guard clause, not when it is a logical branch.

MicahZoltu (approved):

app/ts/background/popupMessageHandlers.ts
KillariDev: I think the typescript cannot handle it as the types depend on each other

MicahZoltu: Hmm, OK.  In that case I prefer if/else for the same reason as below.  Doesn't really feel like a guard clause, but just a branch.

PR #1235, opened 2024-12-30
tell user if we cannot parse input data and don't crash

KillariDev:


MicahZoltu (approved):
This will address the problem with both events and methods?

KillariDev:
> This will address the problem with both events and methods?

yep

PR #1238, opened 2025-01-02
if eth call fails because of eth simulate error, forward that error to dapp

KillariDev:


MicahZoltu (approved):
I assume this is for situations like eth_simulateV1 is missing from the RPC, or the RPC itself is down entirely?

KillariDev:
> I assume this is for situations like eth_simulateV1 is missing from the RPC, or the RPC itself is down entirely?

or eth call requires eth simulate to spend too much gas (which was the case in your bug report). This also happens if eth call sets too low base fee etc

PR #1248, opened 2025-04-01
add time manipulation to transaction stack

KillariDev:


app/ts/simulation/services/EthereumClientService.ts
```diff
@@ -212,7 +212,25 @@ export class EthereumClientService {
...
+			time: parentBlock.timestamp,
+			gasLimit: parentBlock.gasLimit,
+			feeRecipient: parentBlock.miner,
+			baseFeePerGas: parentBlock.baseFeePerGas === undefined ? 15000000n : parentBlock.baseFeePerGas
+		}
+		for (const inputBlock of simulationStateInput.blocks) {
+			const getNewBlockTimeStampOverride = (previousBlockTimeStamp: Date, blockTimeManipulation: BlockTimeManipulation) => {
+				const prevTime = BigInt(previousBlockTimeStamp.getTime()) / 1000n
+				if (blockTimeManipulation.type === 'AddToTimestamp') return new Date(Number(prevTime + blockTimeManipulation.deltaToAdd) * 1000)
+				return new Date(Number(max(prevTime, blockTimeManipulation.timeToSet)) * 1000)
```
MicahZoltu, line 226: I wonder if we should add a utility method for `bigintSecondsToDate` and `dateToBigintSeconds`?

MicahZoltu: Looks like there are other places where this would be beneficial.

MicahZoltu (approved):

app/ts/background/simulationUpdating.ts
```diff
@@ -7,35 +7,65 @@ import { EnrichedEthereumEvents, EnrichedEthereumInputData } from '../types/Enri
...
+				break
+			}
+			default: assertNever(operation)
+		}
+	}
+	if (currentBlockTransactions.length > 0 || currentBlockSignedMessages.length > 0 || Object.values(currentBlockStateOverrides).length > 0) {
+		inputBlocks.push({
+			stateOverrides: currentBlockStateOverrides,
+			transactions: currentBlockTransactions,
+			signedMessages: currentBlockSignedMessages,
```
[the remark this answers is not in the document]
KillariDev: I feel they should be processed in the order they are done as that's logical. There's a guarantee for that if you do such simulation and send the transactions/messages in that order. I have always thought the stack as the order you do operations, not as blockchain blocks.

app/ts/simulation/services/EthereumClientService.ts
KillariDev: yeah I think that's good idea

Issue #1252, opened 2025-04-01
move getStack methods out of simulator so that we don't have to guess rich addresses


KillariDev:
Why do we have to guess?  Isn't it just the selected simulation address?

_Originally posted by @MicahZoltu in https://github.com/DarkFlorist/TheInterceptor/pull/1247#discussion_r2022448989_

PR #1255, opened 2025-04-01
don't guess address to make rich

KillariDev:
fix https://github.com/DarkFlorist/TheInterceptor/issues/1252

- now instead of guessing if we are rich for simulation stack 1.0.1 and 1.0.0,  just use variable for it
- move stack extraction methods outside simulator code


MicahZoltu:
Any chance you could split this into two PRs?  One that just extracts to separate file, and another that makes the modifications?  Right now it is just showing a wall of green and red in the diff, which makes reviewing it hard.  😖

MicahZoltu (approved):

Issue #1257, opened 2025-04-02
I wonder if we should add a utility method for `bigintSecondsToDate` and `dateToBigintSeconds`?


KillariDev:
I wonder if we should add a utility method for `bigintSecondsToDate` and `dateToBigintSeconds`?

_Originally posted by @MicahZoltu in https://github.com/DarkFlorist/TheInterceptor/pull/1248#discussion_r2022795593_

PR #1262, opened 2025-04-03
add bigintSecondsToDate and dateToBigintSeconds

KillariDev:
fix https://github.com/DarkFlorist/TheInterceptor/issues/1257

### DIFF
--- a/app/ts/background/simulationUpdating.ts
+++ b/app/ts/background/simulationUpdating.ts
@@ -1,6 +1,6 @@
 import { Interface, ethers } from 'ethers'
 import { EthereumClientService } from '../simulation/services/EthereumClientService.js'
-import { DEFAULT_BLOCK_MANIPULATION, appendTransactionToInputAndSimulate, calculateRealizedEffectiveGasPrice, createSimulationState, getAddressToMakeRich, getBaseFeeAdjustedTransactions, getNonceFixedSimulationStateInput, getSimulatedCode, getTokenBalancesAfterForTransaction, getWebsiteCreatedEthereumUnsignedTransactions, mockSignTransaction, simulationGasLeft, sliceSimulationState } from '../simulation/services/SimulationModeEthereumClientService.js'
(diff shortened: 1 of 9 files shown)


app/ts/background/simulationUpdating.ts
```diff
@@ -135,7 +135,7 @@ export const simulateGovernanceContractExecution = async (pendingTransaction: Pe
 			simulatedBlocks: [{
 				signedMessages: [],
 				stateOverrides: {},
-				blockTimestamp: new Date(parentBlock.timestamp.toTimeString() + 12 * 1000),
+				blockTimestamp: bigintSecondsToDate((dateToBigintSeconds(parentBlock.timestamp) + getBlockTimeManipulationSeconds(DEFAULT_BLOCK_MANIPULATION.deltaToAdd, DEFAULT_BLOCK_MANIPULATION.deltaUnit))),
```
MicahZoltu, line 138: I wonder if we should just store dates as seconds everywhere and only turn them into `Date` instances at the edges?  It seems silly to have code like `bigintSecondsToDate((dateToBigintSeconds(...`

app/ts/utils/bigint.ts
```diff
@@ -139,3 +139,11 @@ export function calculateWeightedPercentile(data: readonly { dataPoint: bigint,
 	const interpolation = (targetIndex - lowerWeight) / (upperWeight - lowerWeight)
 	return lowerValue.dataPoint + (upperValue.dataPoint - lowerValue.dataPoint) * interpolation
 }
+
+export const bigintSecondsToDate = (seconds: bigint) => {
+	if (seconds > Number.MAX_SAFE_INTEGER) throw new Error(`Too big seconds value: ${ seconds }`)
```
MicahZoltu, line 144: Suggest removing this restriction for a few reasons:
1. A very big number will just get rounded, and I don't think anyone will care that a date millions of years in the future is rounded by a second or two.
2. `new Date(Number.MAX_SAFE_INTEGER * 1000)` results in an error because the Date object can't handle dates that far in the future.
3. You are multiplying this by 1000 below, so this doesn't actually protect against rounding.
You could address each of these, but it doesn't seem worth it.  If you do want to address them, then I recommend figuring out what the max safe date seconds is and using that as a restriction instead.  https://stackoverflow.com/a/27093173 suggests max safe date **seconds** is `8,640,000,000,000` which is *way* below `MAX_SAFE_INTEGER`.

```suggestion
```
```suggestion
	if (seconds > 8640000000000n) throw new Error(`Too big seconds value: ${ seconds }`)
```

MicahZoltu (approved):

app/ts/background/simulationUpdating.ts
KillariDev: yeah, I added issue for this, it's better

app/ts/utils/bigint.ts
KillariDev: it protects cases where the seconds are just too big for some probably accidental reason, but yeah, we can make it more strict

PR #1269, opened 2025-04-08
serialize `Uint8Array`'s as hex strings instead of arrays

KillariDev:


app/ts/utils/bigint.ts
```diff
@@ -91,7 +91,11 @@ export function bigintToUint8Array(value: bigint, numberOfBytes: number) {
 
 // biome-ignore lint/suspicious/noExplicitAny: matches JSON.stringify signature
 export function stringifyJSONWithBigInts(value: any, space?: string | number | undefined): string {
-	return JSON.stringify(value, (_key, value) => { return typeof value === 'bigint' ? `0x${ value.toString(16) }` : value }, space)
+	return JSON.stringify(value, (_key, value) => {
+		if (typeof value === 'bigint') return `0x${ value.toString(16) }`
```
MicahZoltu, line 95: Why don't we serialize bigints with `bigintToDecimalString`?

MicahZoltu (approved):

app/ts/utils/bigint.ts
KillariDev: because it might be address and also we don't know the decimals of random bigints

PR #1273, opened 2025-04-16
Eip712 validator

KillariDev:
- adds eip712 validator
- code tags background is no longer white
- if verification fails, show user the message with error


app/ts/utils/eip712.ts
```diff
@@ -0,0 +1,398 @@
...
+
+	// Check for bytes types like bytes32
+	const fixedBytesRegex = /^bytes(\d+)$/
+	const fixedBytesMatch = typeStr.match(fixedBytesRegex)
+	if (fixedBytesMatch) {
+		if (!fixedBytesMatch[1]) return { valid: false, reason: 'not valid bytes' }
+		const expectedSize = parseInt(fixedBytesMatch[1])
+		// Value must be a hex string with 2 + 2*expectedSize characters (0x + hex digits)
+		const valid = typeof value === 'string' && /^0x[a-fA-F0-9]*$/.test(value) && value.length === 2 + expectedSize * 2
+		return valid ? { valid: true } : { valid: false, reason: `${ value } is invalid bytes string` }
```
MicahZoltu, line 146:
```suggestion
		const expectedSize = parseInt(fixedBytesMatch[1])
		if (expectedSize < 1 || expectedSize > 32) return { valid: false, reason: `Only 'bytes1' through 'bytes32' are valid, '${typeStr}' is not in that range` }
		// Value must be a hex string with 2 + 2*expectedSize characters (0x + hex digits)
		const valid = typeof value === 'string' && /^0x[a-fA-F0-9]*$/.test(value) && value.length === 2 + expectedSize * 2
		return valid ? { valid: true } : { valid: false, reason: `${ value } is invalid bytes string` }
```

app/ts/utils/eip712.ts
```diff
@@ -0,0 +1,398 @@
...
+		// For signed: range is -2^(width-1) to 2^(width-1) - 1
+		const min = isUnsigned ? 0n : -(2n ** BigInt(bitWidth - 1))
+		const max = isUnsigned ? (2n ** BigInt(bitWidth)) - 1n : (2n ** BigInt(bitWidth - 1)) - 1n
+
+		let bigValue: bigint
+		if (typeof value === 'number') {
+			if (!Number.isInteger(value)) return { valid: false, reason: `${ JSON.stringify(value) } wasn't integer` }
+			bigValue = BigInt(value)
+		} else if (typeof value === 'string') {
+			const parsed = Eip712Number.safeParse(value) // todo, does the quantity need to be in hex format?
```
MicahZoltu, line 128: Does this `TODO` need to be addressed?  Are there any others?

app/ts/utils/eip712.ts
```diff
@@ -0,0 +1,398 @@
...
+	if (!identifierPattern.test(typeStr)) return { valid: false, reason: `invalid type: ${ typeStr }` }
+
+	// Check for integer types, e.g. uint128 or int256
+	const intRegex = /^(u?)int(\d+)$/
+	const intMatch = typeStr.match(intRegex)
+	if (intMatch) {
+		// Use BigInt to check numeric ranges
+		const isUnsigned = intMatch[1] === 'u'
+		if (!intMatch[2]) return { valid: false, reason: `intMatch was undefined` }
+		const bitWidth = parseInt(intMatch[2])
```
MicahZoltu, line 117:
```suggestion
		const bitWidth = parseInt(intMatch[2])
		if (bitWidth < 8 || bitWidth > 256) return { valid: false, reason: `${typeStr} is not a valid type, number widths must be in range [8,256]` }
		if (!Number.isInteger(bitWidth / 8) return {valid: false, reason: `${typeStr} is not a valid type, number widths must be divisible by 8.`
```

MicahZoltu (approved):
Has tests, must be fine.  😬

app/ts/utils/eip712.ts
KillariDev: The spec doesn't specify this but in practical cases the numbers can be represented in many formats. So I am assuming its correc

PR #1275, opened 2025-04-26
if no direct path is found, do not consider it a proxy transfer

KillariDev:
- fixes a bug that initial report is considered as proxy transfer on Augur


app/ts/components/simulationExplaining/identifyTransaction.tsx
```diff
@@ -212,7 +212,8 @@ function isProxyTokenTransfer(transaction: SimulatedAndVisualizedTransaction): t
 	const edges = tokenResults.map((tokenResult) => ({ from: tokenResult.from.address, to: tokenResult.to.address, data: tokenResult.to, amount: !tokenResult.isApproval && tokenResult.type !== 'ERC721' ? tokenResult.amount : 1n }))
 	const deadEnds = findDeadEnds(edges, transaction.transaction.from.address)
 	if (deadEnds.size === 0) return false
-
+	const path = removeDuplicates(Array.from(deadEnds).flatMap(([_key, edges]) => edges.slice(0, -1).map((edge) => edge.data)))
+	if (path.length === 0) return false
```
MicahZoltu, line 216: It feels like you could just do `Array.from(deadEnds).every(([_key, edges]) => edges.slice(0, -1).map(edge => edge.data).length === 0)` (or something along those lines) instead, rather that saving everything along the way and checking at the end to see if you collected anything.  If the first item has edge data, `.every` will terminate early instead of running over the whole array.

MicahZoltu (approved):

PR #1279, opened 2025-05-08
allow signed types to actually be negative

KillariDev:


app/ts/types/solidityType.ts
```diff
@@ -1,18 +1,18 @@
 import * as funtypes from 'funtypes'
-import { EthereumAddress, EthereumData, EthereumQuantity } from './wire-types.js'
+import { EthereumAddress, EthereumData, EthereumQuantity, SignedEthereumQuantity } from './wire-types.js'
 import { AddressBookEntry } from './addressBookTypes.js'
 
 export type PureGroupedSolidityType = funtypes.Static<typeof PureGroupedSolidityType>
 export const PureGroupedSolidityType = funtypes.Union(
-	funtypes.ReadonlyObject({ type: funtypes.Literal('signedInteger'), value: EthereumQuantity }),
+	funtypes.ReadonlyObject({ type: funtypes.Literal('signedInteger'), value: SignedEthereumQuantity }),
```
MicahZoltu, line 7: Hmm, I feel like we shouldn't be using `EthereumQuantity` for this at all.  `EthereumQuantity` has a specific meaning in RLP encoding, and RLP encoding doesn't have a concept of signed types at all.  Separately there is Solidity ABI encoding, which is unrelated to Ethereum RLP encoding but I think we should have different variables for that rather than trying to use the same ones for both.

MicahZoltu (approved):

PR #1280, opened 2025-05-08
error handling fixes

KillariDev:
- Fixed: if you make duplicated rpc calls, and the call fails, the duplicated promises never get rejected
- Fixed: If a new block arrives when new transaction is being refreshed, undefined error is thrown
- Fixed: if get code call fails, return error to UI instead showing undefined error

### DIFF
--- a/app/ts/simulation/services/EthereumJSONRpcRequestHandler.ts
+++ b/app/ts/simulation/services/EthereumJSONRpcRequestHandler.ts
@@ -51,16 +51,15 @@ export class EthereumJSONRpcRequestHandler {
 			const responseObject = response.ok ? { responseState: 'success' as const, response: await response.json() } : { responseState: 'failed' as const, response }
 			this.cache.set(hash, responseObject)
 			future.resolve(responseObject)
-			return responseObject
 		} catch(error: unknown) {
 			if (error instanceof Error) {
 				future.reject(error)
 			} else {
 				future.reject(new Error('Unknown error'))
 			}
-			throw error
 		} finally {
 			this.pendingCache.delete(hash)
+			return await future
 		}
 	}
 
--- a/app/ts/simulation/services/EthereumJSONRpcRequestHandler.ts
+++ b/app/ts/simulation/services/EthereumJSONRpcRequestHandler.ts
@@ -52,6 +52,13 @@ export class EthereumJSONRpcRequestHandler {
 			this.cache.set(hash, responseObject)
 			future.resolve(responseObject)
 			return responseObject
+		} catch(error: unknown) {
+			if (error instanceof Error) {
+				future.reject(error)
+			} else {
+				future.reject(new Error('Unknown error'))
+			}
+			throw error
 		} finally {
 			this.pendingCache.delete(hash)
 		}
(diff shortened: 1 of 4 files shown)


app/ts/simulation/services/EthereumJSONRpcRequestHandler.ts
```diff
@@ -52,6 +52,13 @@ export class EthereumJSONRpcRequestHandler {
 			this.cache.set(hash, responseObject)
 			future.resolve(responseObject)
 			return responseObject
+		} catch(error: unknown) {
+			if (error instanceof Error) {
+				future.reject(error)
+			} else {
+				future.reject(new Error('Unknown error'))
+			}
+			throw error
```
MicahZoltu, line 61: Why throw here?  It feels like the future rejection is sufficient?

app/ts/simulation/services/SimulationModeEthereumClientService.ts
```diff
@@ -450,16 +450,17 @@ export const getSimulatedCode = async (ethereumClientService: EthereumClientServ
...
-	const parsed = atInterface.decodeFunctionResult('at', lastResult.returnData)
-	return {
-		statusCode: lastResult.status,
-		getCodeReturn: EthereumData.parse(parsed.toString())
-	} as const
+	try {
+		const simulatedTransactions = await simulateTransactionsOnTopOfSimulationState(ethereumClientService, requestAbortController, simulationState, [getCodeTransaction], { [addressString(GET_CODE_CONTRACT)]: { code: getCodeByteCode() } })
+		const lastResult = simulatedTransactions.at(-1)
+		if (lastResult === undefined) throw new Error('last result did not exist in multicall')
+		if (lastResult.status === 'failure') return { statusCode: 'failure' } as const
```
MicahZoltu, line 457: The lines that are throws are situations that are unexpected while the `statusCode: failure` are "expected" failure modes?

app/ts/simulation/services/SimulationModeEthereumClientService.ts
```diff
@@ -450,16 +450,17 @@ export const getSimulatedCode = async (ethereumClientService: EthereumClientServ
...
+		const parsed = atInterface.decodeFunctionResult('at', lastResult.returnData)
+		return { statusCode: lastResult.status, getCodeReturn: EthereumData.parse(parsed.toString()) } as const
+	} catch(error: unknown) {
+		if (error instanceof JsonRpcResponseError) return { statusCode: 'failure' } as const
+		throw error
+	}
```
MicahZoltu, line 463: As we have discussed previously, it feels like we should not be using try/catch to handle code flow.  Erorrs from a remote server should be handled as part of normal code paths, not as part of exceptional code paths.  Is there no way to detect an error response without throwing in our current codebase?

MicahZoltu (approved):

app/ts/simulation/services/SimulationModeEthereumClientService.ts
KillariDev: yeah, the thrown ones should never happen if the RPC behaves correctly. The other failure happen on normal cases

app/ts/simulation/services/EthereumJSONRpcRequestHandler.ts
KillariDev: it wasn't sufficient as we never waited for that promise here (it was indended for others to wait it), but I made it more straightforward that it also awaits that

PR #1287, opened 2025-05-16
don't make address rich on signing mode

KillariDev:


MicahZoltu (approved):
Surely, fixing it for the 42nd time will actually fix the issue.  😬

PR #1294, opened 2025-07-18
use hostname+port if port is available to identify sites

KillariDev:


app/ts/utils/requests.ts
```diff
@@ -120,3 +120,8 @@ export const checkAndPrintRuntimeLastError = () => {
 	// biome-ignore lint/suspicious/noConsoleLog: <Used for support debugging>
 	if (error !== null && error !== undefined && error.message !== undefined) console.log(error)
 }
+
+export const getHostWithPort = (urlString: string): string => {
+	const url = new URL(urlString)
+	return url.port ? `${ url.hostname }:${ url.port }` : url.hostname
```
MicahZoltu, line 126: If we wanted to be explicit we could treat `http` without a port as `80` and `https` without a port as `443`.  However, since this project is in maintenance mode I'm fine with just this as it is strictly equal to or greater safety than appending ports to schemes with known ports.

MicahZoltu (approved):

PR #1306, opened 2025-12-11
make popup requests replying data reply that data and those that do not reply anything should return void

KillariDev:
- There's a bug that if popup window method returns data, the listener expects it to reply something to the request. However, most popup methods do not return any data so this results in an error


MicahZoltu (approved):
Pending debug code removal.

Issue #1309, opened 2025-12-15
This was allegedly fixed 2 years ago.  Can we switch to doing the right thing?


KillariDev:
This was allegedly fixed 2 years ago.  Can we switch to doing the right thing?
```suggestion
		<!-- a workaround for bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1803984 -->
		<script async type = 'module' src = '../js/fetchSimulationStack.js'></script>
```

_Originally posted by @MicahZoltu in https://github.com/DarkFlorist/TheInterceptor/pull/1305#discussion_r2612900910_

PR #1312, opened 2025-12-15
Bug around

KillariDev:
fix https://github.com/DarkFlorist/TheInterceptor/issues/1309


MicahZoltu (approved):
Did you verify this does in fact work on Firefox now if you load the page from a local filesystem (~1ms response time to requests, since it is a race condition).

KillariDev:
> Did you verify this does in fact work on Firefox now if you load the page from a local filesystem (~1ms response time to requests, since it is a race condition).

yeah, I tested that it works

PR #1316, opened 2025-12-17
dropdown fixes, allow changing address type and decimals

KillariDev:
- allow changing address types
- allow overwriting decimals
- some styling changes and removing some non-used styles
image

image


MicahZoltu (approved):
Approved after type extraction.

PR #1328, opened 2025-12-29
small cleanups, add check that we are not trying store duplicated ids, require ethereum bigint typest to be non-negative

KillariDev:


app/ts/types/wire-types.ts
```diff
@@ -9,6 +9,7 @@ const BigIntParser: funtypes.ParsedValue<funtypes.String, bigint>['config'] = {
 	},
 	serialize: value => {
 		if (typeof value !== 'bigint') return { success: false, message: `${typeof value} is not a bigint.`}
+		if (value < 0n) return { success: false, message: `${typeof value} is not a non negative bigint.`}
```
MicahZoltu: I know that in the primary consensus protocol there are no negative numbers, but are there really no negative numbers anywhere in the Ethereum JSON-RPC?

PR #1331, opened 2026-01-07
silence chrome uncaught promise errors

KillariDev:


app/ts/background/background-startup.ts
```diff
@@ -101,6 +101,7 @@ async function onContentScriptConnected(simulator: Simulator, port: browser.runt
 	const websiteOrigin = getHostWithPort(port.sender.url)
 	const identifier = websiteSocketToString(socket)
 	const websitePromise = (async () => ({ websiteOrigin, ...await retrieveWebsiteDetails(socket.tabId) }))()
+	silenceChromeUnCaughtPromise(websitePromise)
```
MicahZoltu: Same question everywhere you pass a local variable to this function: why use the helper function when it is less code to just inline directly?
```suggestion
	websitePromise.catch(() => {})
```

In cases where you are receiving the promise in the same line you want to attach the catch it does save you one line, so I get those cases.

app/ts/utils/future.ts
```diff
@@ -10,6 +10,7 @@ export class Future<T> implements PromiseLike<T> {
 			resolveFunction = resolve
 			rejectFunction = reject
 		})
+		this.promise.catch(() => { })
```
MicahZoltu: See my top level comment asking if we really need all of these.  At the least, I think this one is wrong and is perhaps worthy of further discussion that may bleed over into the others.  We should not be attaching catch handlers to *every* promise everywhere.  Only promises that have a codepath that results in the promise never being awaited.  This is a low level internal utility class that returns a promise to the caller.  It would be up to the caller to analyze whether there exists an unawaited code path or not and attach a catch handler in that case.

If you just want to silence the browser entirely, then I think we should just attach an unhandled rejection handler once and not include this code everywhere.

For context, I have been fighting NodeJS and browsers on this since (checks notes) 2015, so I'm not opposed to silencing the error entirely if you think that is the right solution.  I'm also not opposed to painstakingly going through each code path and determining whether or not there is an unhandled potentiality and carefully attaching handlers in those locations only.  What I'm against is littering the code with these as a way to blanket silence everything.

https://github.com/promises-aplus/unhandled-rejections-spec/issues/6#issuecomment-105335603
https://github.com/promises-aplus/unhandled-rejections-spec/issues/9#issuecomment-266178067

```suggestion
```

MicahZoltu (commented):
Are all of these locations really leaky?  Feels like far more than I expected.  I was expecting maybe one or two places could leak.

MicahZoltu (approved):

PR #1346, opened 2026-01-16
propagate failing simulation to the UI

KillariDev:
if eth_simulate fails to running out of gas (either global, or per block) propagate that error to the UI. Currently user can only clear the stack and start over, the idea is to let user see the transactions still and possibly reduce their gas limits or remove them. I also plan to make the system to detect this case and try to avoid it, but its not always possible


app/ts/utils/errors.ts
```diff
@@ -10,15 +10,20 @@ export class ErrorWithData extends Error {
...
 	public readonly code: number
+	public readonly data: string | undefined
 	public constructor(jsonRpcResponse: JsonRpcErrorResponse) {
-		super(jsonRpcResponse.error.message, jsonRpcResponse.error.data)
+		super(jsonRpcResponse.error.message)
 		this.code = jsonRpcResponse.error.code
 		this.id = jsonRpcResponse.id
+		this.data = jsonRpcResponse.error.data
 		Object.setPrototypeOf(this, JsonRpcResponseError.prototype)
 	}
```
MicahZoltu: Isn't this the same?  `ErrorWithData` already has a `.data` property.  Is the issue that sometimes the `.data` field is missing/undefined in JSON-RPC errors?

MicahZoltu (approved):

PR #1349, opened 2026-01-19
gas fix transactions

KillariDev:
- if gas limit of a transaction is higher than blocks gas, clamp it to there
- if total gas limit of all transactions in a block is higher than block gas limit, pick the highest gas limit transaction and reduce it by 10%, repeat until the transactions are under


KillariDev:
> I feel like this is going to break things more often than not as it indiscriminately reduces gas limits. Can we instead just split transactions across multiple blocks?

Hmm yeah, thats one approach, but even then we need to limit the gas limits to below those blocks. Its also fine that transactions fail to out of gas as the simulation can still complete

MicahZoltu (approved):

PR #1365, opened 2026-01-27
cleanup copy to clipboard element

KillariDev:


MicahZoltu (approved):
I like simplification.

PR #1372, opened 2026-04-13
clear button fixing

KillariDev:
previously if you click clear button it might not clear the simulation stack, this fixes the issue


MicahZoltu (approved):
This has annoyed me for a while, but I was always too busy to report it.  😬

PR #1376, opened 2026-04-14
fix uncaught errors

KillariDev:
- fix uncaught error: Simulation results are required
- add error boundary that catches and shows uncaught UI errors to users


MicahZoltu (approved):
Approved pending conversion into a functional component.  I feel pretty strongly about not using class based components, both for stylistic consistency and as a generalized pattern that results in more maintainable code (less hidden state).

PR #1377, opened 2026-04-14
Refactor signer bridge sync in inpage provider

KillariDev:
## Summary
- Removed the implicit signer account sync during connect and now rely on explicit account reply handling.
- Updated inpage provider access to use a typed `window` reference consistently for ethereum/web3 interactions.
- Adjusted background handling so signer connection and access refresh flows no longer depend on hidden account fetches.
- Added coverage for the signer bridge to verify connect only requests chain ID and explicit account requests still propagate correctly.

## Testing
- Not run (PR content only).
- Existing test added: `test/tests/inpageSignerBridge.ts` covers connect flow and explicit `eth_accounts` / `eth_requestAccounts` replies.
- Repo instructions in `AGENTS.md` require running `npm test` and `npm run setup-chrome` after code changes.

### DIFF
--- a/app/inpage/ts/inpage.ts
+++ b/app/inpage/ts/inpage.ts
@@ -141,10 +141,8 @@ type WindowEthereum = InjectFunctions & {
 	chainId?: string,
 	networkVersion?: string,
 }
-interface Window {
-	dispatchEvent: (event: Event) => boolean
-	addEventListener: (type: string, listener: (event: unknown) => void) => void
-	postMessage: (message: unknown, targetOrigin: string) => void
+
+type InpageWindow = Window & {
 	ethereum?: WindowEthereum
 	web3?: {
 		currentProvider: WindowEthereum
@@ -152,7 +150,7 @@ interface Window {
 	}
 }
 
-declare const window: Window
+const inpageWindow = window as InpageWindow
 
 interface EIP6963ProviderInfo {
 	uuid: string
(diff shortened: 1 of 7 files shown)


MicahZoltu:
Your bot's commit message generation could use some work.

app/inpage/ts/inpage.ts
```diff
@@ -141,15 +141,17 @@ type WindowEthereum = InjectFunctions & {
...
+
+type InpageWindow = Window & {
 	ethereum?: WindowEthereum
 	web3?: {
 		currentProvider: WindowEthereum
 		accounts: readonly string[]
 	}
 }
 
+const inpageWindow: InpageWindow = window
```
MicahZoltu, line 153: This is *probably* fine for the `inpage.ts` specifically, but file scoped variables are generally a bad idea.  Also, might be worth a comment indicating why you are typcasting here, just as a best practice?

MicahZoltu: I feel like we should be verifying that `window` actually is an `InpageWindow` at runtime, in case injection partially fails or something.

MicahZoltu (approved):

KillariDev:
> Your bot's commit message generation could use some work.

what do you mean by this? How would you improve them?

app/inpage/ts/inpage.ts
KillariDev: it is checked for that type a runtime, but it can fail at any time as it can be overridden

MicahZoltu:
For example, it wrote this in the commit message:
> Repo instructions in AGENTS.md require running npm test and npm run setup-chrome after code changes.

Also, I don't think writing what steps one took is useful in a PR description.  The PR description should describe the changes to the code, not the steps taken to get there.  For example, this line seems unnecessary:
> Not run (PR content only).

PR #1378, opened 2026-04-14
Refactor popup state to signals

KillariDev:
## Summary
- Migrates popup state management from `useState` to Preact signals across the main app and confirmation flows.
- Updates page components and shared subcomponents to read from signals directly, reducing prop drilling and state synchronization overhead.
- Adjusts type definitions and visualizer components to accept signal-backed state where needed.
- Preserves existing popup behavior while simplifying state updates for address, chain, simulation, and access-list UI.


MicahZoltu:
GLM 5.1 generated.  Take with grain of salt, but worth at least reviewing I think.

### BUG 1 (High): `useComputed` called inside conditional JSX — rules of hooks violation

**File:** `app/ts/components/pages/Home.tsx:315,319,329,333`

In `PopupVisualisation`, `useComputed` is called inside mutually exclusive conditional JSX branches:

```tsx
{ simResults.fail ? <>
    <TransactionsAndSignedMessages
        activeAddress = { useComputed(() => definedSimulationResults.value.activeAddress) }   // HOOK 1
        addressMetaData = { useComputed(() => definedSimulationResults.value.addressBookEntries) } // HOOK 2
    />
</> : <>
    <TransactionsAndSignedMessages
        activeAddress = { useComputed(() => definedSimulationResults.value.activeAddress) }   // HOOK 3
        addressMetaData = { useComputed(() => definedSimulationResults.value.addressBookEntries) } // HOOK 4
    />
</> }
```

`useComputed` is implemented via `useMemo` under the hood and must follow the **rules of hooks** — it must be called in the same order on every render. When the simulation result transitions between success/failure, a different number of hooks execute, which will corrupt Preact's hook memoization state and cause **crashes or stale/incorrect signal values**.

**Fix:** Lift the `useComputed` calls above the conditional, e.g.:
```tsx
const computedActiveAddress = useComputed(() => definedSimulationResults.value.activeAddress)
const computedAddressMetaData = useComputed(() => definedSimulationResults.value.addressBookEntries)
```
Then pass `computedActiveAddress` / `computedAddressMetaData` to both branches.

---

### BUG 2 (High): `popup_refreshHomeData` and `popup_refreshSimulation` no longer sent on new block arrival

**File:** `app/ts/components/App.tsx` — `popup_new_block_arrived` handler

The old code sent two messages on every new block:
```ts
sendPopupMessageToBackgroundPage({ method: 'popup_refreshHomeData' })
sendPopupMessageToBackgroundPage({ method: 'popup_refreshSimulation' })
```

The new code **removes both** and only updates `currentBlockNumber` and `rpcConnectionStatus` directly from the message data. The background (`background-startup.ts`) now handles `updatePopupVisualisationIfNeeded` server-side in simulation mode, but:

- **Non-simulation mode:** The background does NOT call `updatePopupVisualisationIfNeeded` on new blocks. This means the popup's simulation results will go stale in signing mode unless the user manually triggers a refresh.
- **`popup_refreshHomeData`** is never sent on new blocks, so address book, settings, tab state, etc. are not refreshed between blocks. This was previously how the popup stayed in sync with background state changes.

**Impact:** The popup will not reflect background state changes that happen between blocks (e.g., signer account changes, interceptor status changes, updated website access lists).

---

### BUG 3 (Medium): `noReplyExpectingBrowserRuntimeOnMessageListener` drops `sendResponse` capability

**File:** `app/ts/utils/browser.ts`

The old implementation passed `callback`'s return value directly as the `onMessage` listener's return value, which Chrome uses to synchronously respond to `sendMessage`. The new implementation:
```ts
browser.runtime.onMessage.addListener((message: unknown) => {
    void Promise.resolve(callback(message)).catch(...)
    return undefined
})
```

This **always** returns `undefined`, preventing any synchronous response. For `popup_isMainPopupWindowOpen` in `ConfirmTransaction.tsx` and `InterceptorAccess.tsx` (which use this wrapper), the `PopupMessageReplyRequests` message will never get a reply through this path. This is fine only because `App.tsx` handles the reply in its own separate listener — but it creates a **fragile coupling** where the reply only works if `App.tsx` is also loaded and listening.

---

### BUG 4 (Medium): `InterceptorAccess.tsx` — `informationChangedRecently` may not trigger re-render at the right time

**File:** `app/ts/components/pages/InterceptorAccess.tsx`

The `useComputed` for `informationChangedRecently` depends on `timeTicker.value` (incremented every second) and `informationUpdatedTimestamp.value`. However, `useComputed` only re-runs when its dependencies change **and** it's being observed. If no component reads `informationChangedRecently.value` between tick changes, the computed may not re-evaluate, leaving the button state stale. In practice this should work because the component reads it in its render, but it's worth noting as a subtle difference from the old `useState`-based approach where React always re-rendered on `setTimeSinceInformationUpdate`.

---

### BUG 5 (Low): `Modal.tsx` was already broken (now fixed)

**File:** `app/ts/components/subcomponents/Modal.tsx`

The old code had `addEventListener` in the cleanup function instead of `removeEventListener`. The new code fixes this. This is a **positive** change, not a regression.

---

### CONCERN 1 (Medium): Simulation timestamp-based staleness may reject legitimate updates

**File:** `app/ts/background/popupMessageHandlers.ts:299-313`, `app/ts/background/windows/confirmTransaction.ts`

New logic compares `simulationConductedTimestamp` to decide whether a visualization update should replace the current one:
```ts
if (currentTimestamp !== undefined && nextTimestamp !== undefined && nextTimestamp.getTime() < currentTimestamp.getTime()) return transactionOrMessage
```

This could reject valid updates if two simulations complete out of order (fast one finishes after slow one). Since `simulationConductedTimestamp` is now set to `new Date()` at visualization time (not simulation start time), this should mostly be correct, but there's a narrow race window where two concurrent simulation threads could have their timestamps ordered opposite to their actual freshness.

---

### CONCERN 2 (Low): `SimulationStateParam.removedTransactionOrSignedMessages` type inconsistency

**File:** `app/ts/types/user-interface-types.ts`

The type definition was changed to `Signal<readonly TransactionOrMessageIdentifier[]>`, but in `Home.tsx` the value is unwrapped with `.value` before passing to `TransactionsAndSignedMessages`. The `TransactionsAndSignedMessagesParams` type still expects `readonly TransactionOrMessageIdentifier[]` (plain array). This works but creates a type-level inconsistency where the type says "Signal" but consumers expect a plain array.

---

### CONCERN 3 (Low): `ConfirmTransaction.tsx` — inconsistent indentation of `getLoadingText`/`loadingText`

**File:** `app/ts/components/pages/ConfirmTransaction.tsx:581-591`

The `getLoadingText` function and `loadingText` computed are at an extra level of indentation (inside the component function body, but indented as if inside a block). While functionally correct (they're at component scope), this suggests a merge artifact or accidental re-indentation.

---

### Positive Changes Worth Noting

1. **Bug fix in Modal.tsx** — `addEventListener` → `removeEventListener` in cleanup
2. **Simulation fingerprinting** (`popupSimulationFingerprint.ts`) — avoids unnecessary re-simulations when input hasn't changed
3. **Message routing via `role` field** — prevents `confirmTransaction` window from processing main popup messages and vice versa
4. **Proper `await` of `updatePopupVisualisationIfNeeded`** — several places that previously fired-and-forgot now properly await, preventing race conditions
5. **`simulationConductedTimestamp` refresh** — prevents stale timestamps from blocking fresh visualization updates

app/ts/components/pages/Home.tsx
```diff
@@ -308,30 +309,30 @@ function PopupVisualisation(param: SimulationStateParam) {
...
-			/>
+				<TransactionsAndSignedMessages
+					simulationAndVisualisationResults = { definedSimulationResults }
+					removeTransactionOrSignedMessage = { param.removeTransactionOrSignedMessage }
+					activeAddress = { useComputed(() => definedSimulationResults.value.activeAddress) }
+					renameAddressCallBack = { param.renameAddressCallBack }
+					editEnsNamedHashCallBack = { param.editEnsNamedHashCallBack }
+					removedTransactionOrSignedMessages = { param.removedTransactionOrSignedMessages.value }
+					addressMetaData = { useComputed(() => definedSimulationResults.value.addressBookEntries) }
+				/>
```
MicahZoltu, line 321: `useComputed` is called inside conditional JSX branches. `useComputed` is implemented with `useRef` + `useMemo` under the hood, so it must follow the rules of hooks — the same number must be called in the same order on every render. `PopupVisualisation` has two early returns (lines ~282 and ~287) that execute before the main render path, resulting in fewer hook calls when the simulation is empty or undefined. Additionally, the two `useComputed` calls inside the ternary at line ~315 and ~329 create hooks inside conditional JSX props.

Fix by lifting the `useComputed` calls above all early returns and out of the JSX:
```ts
const computedActiveAddress = useComputed(() => definedSimulationResults.value.activeAddress)
const computedAddressBookEntries = useComputed(() => definedSimulationResults.value.addressBookEntries)
```
Then pass `computedActiveAddress` / `computedAddressBookEntries` to both branches of the ternary.

app/ts/types/visualizer-types.ts
```diff
@@ -287,9 +287,9 @@ export type TransactionVisualizationParameters = {
 	simTx: MaybeSimulatedTransaction
 	simulationAndVisualisationResults: ReadonlySignal<SimulationAndVisualisationResults>
 	removeTransactionOrSignedMessage: ((transactionOrMessageIdentifier: TransactionOrMessageIdentifier) => void) | undefined
-	activeAddress: bigint
+	activeAddress: ReadonlySignal<bigint | undefined>
 	renameAddressCallBack: RenameAddressCallBack
-	addressMetaData: readonly AddressBookEntry[]
+	addressMetaData: ReadonlySignal<readonly AddressBookEntry[]>
 	editEnsNamedHashCallBack: EditEnsNamedHashCallBack
```
MicahZoltu, line 293: `removedTransactionOrSignedMessages` is typed as `Signal<readonly TransactionOrMessageIdentifier[]>` in `SimulationStateParam`, but downstream consumers like `TransactionsAndSignedMessagesParams` (in `Transactions.tsx`) expect `readonly TransactionOrMessageIdentifier[]` (a plain array). The mismatch is bridged at call sites by manually unwrapping with `.value`, but the type declaration doesn't reflect this — it claims the prop is a Signal while consumers treat it as a plain array. Either the type should be `readonly TransactionOrMessageIdentifier[]` (matching usage), or the downstream types should accept `ReadonlySignal<...>` and the unwrapping should happen inside those components.

app/ts/components/pages/ConfirmTransaction.tsx
```diff
@@ -537,13 +581,14 @@ export function ConfirmTransaction() {
...
-		return 'Loading...'
-	}
+		const getLoadingText = (current: PendingTransactionOrSignableMessage | undefined) => {
+			if (current === undefined) return 'Initializing...'
+			if (current.transactionOrMessageCreationStatus === 'Crafting') return 'Crafting Transaction...'
+			if (current.transactionOrMessageCreationStatus === 'Simulating') return 'Simulating Transaction...'
+			if ('popupVisualisation' in current && current.popupVisualisation?.statusCode === 'failed') return 'Failed to simulate. Retrying...'
+			return 'Loading...'
+		}
+		const loadingText = useComputed(() => getLoadingText(currentPendingTransactionOrSignableMessage.value))
```
MicahZoltu: `getLoadingText` and `loadingText` are indented one level too deep, as though they're inside a block scope, but they're actually at the component function's top level (same scope as `isConfirmDisabled` and `renameAddressCallBack` above). This is likely a merge artifact and should be de-indented to match the surrounding code.

app/ts/background/popupMessageHandlers.ts
```diff
@@ -289,16 +299,29 @@ export async function refreshPopupConfirmTransactionSimulation(simulator: Simula
 		switch (transactionOrMessage.type) {
 			case 'SignableMessage': throw new Error('Tried to refresh simulation of a message')
 			case 'Transaction': {
+				if (transactionOrMessage.transactionOrMessageCreationStatus !== 'Simulated' && transactionOrMessage.transactionOrMessageCreationStatus !== 'FailedToSimulate') return transactionOrMessage
+				const currentTimestamp = getSimulationConductedTimestamp(transactionOrMessage.popupVisualisation)
+				const nextTimestamp = getSimulationConductedTimestamp(refreshMessage)
+				if (currentTimestamp !== undefined && nextTimestamp !== undefined && nextTimestamp.getTime() < currentTimestamp.getTime()) return transactionOrMessage
```
MicahZoltu, line 305: The timestamp comparison `nextTimestamp.getTime() < currentTimestamp.getTime()` guards against replacing a visualization with an older one. If two simulations are kicked off near-simultaneously (e.g. block update + user action), they could complete in the opposite order of their start times, causing the fresher result to be discarded. This is unlikely in practice due to the abort controller (`NEW_BLOCK_ABORT`), but worth noting that the timestamp comparison is a heuristic, not a guarantee of freshness.

MicahZoltu (approved):

PR #1388, opened 2026-04-20
Speed up popup open and stabilize refresh handling

KillariDev:
## Summary
- Added popup bootstrap caching and lazy loading to reduce popup open latency.
- Refactored popup state flow so active address and simulation ordering are tracked through UI params and timestamps instead of visualisation state.
- Tightened background popup refresh handling for metadata and simulation updates, including role-scoped popup messaging.
- Cleaned up inpage signer bridge handling and improved error propagation/reporting paths.
- Expanded coverage around popup reset, timestamped refreshes, and inpage signer bridge behavior.

## Testing
- Not run (not executed in this environment).
- Expected validation from repo instructions: `npm test`.
- Expected validation from repo instructions: `npm run setup-chrome`.


MicahZoltu (commented):
## CI Agent Review

The most significant concern is the weakening of `PopupReplyOption` from a strictly validated funtypes union to `Codec<any>`, which effectively disables runtime validation of all responses from the background script to the popup. This removes a defense-in-depth layer against corrupted or malformed data. A secondary concern is that `requestBootstrapData` silently falls back to default `TabState` values on error, which could mask failures.

app/ts/types/interceptor-reply-messages.ts
```diff
@@ -156,7 +204,8 @@ export type PopupRequests = funtypes.Static<typeof PopupMessageReplyRequests>
 export type PopupRequestsReplyReturn<Request extends PopupRequests> = Request['method'] extends keyof PopupRequestsReplies ? PopupRequestsReplies[Request['method']] : undefined
 
 export type PopupReplyOption = funtypes.Static<typeof PopupReplyOption>
-export const PopupReplyOption = funtypes.Union(
+export const PopupReplyOption: Codec<any> = funtypes.Union(
```
MicahZoltu: Changing `PopupReplyOption` from a typed funtypes union to `Codec<any>` disables runtime validation of all background-to-popup responses. Previously, `PopupReplyOption.parse()` would reject unexpected or corrupted data; now it accepts anything. This removes a defense-in-depth layer. Consider keeping the strict union type and adding `PopupBootstrapData` to it instead of widening to `Codec<any>`.

app/ts/background/popupMessageHandlers.ts
```diff
@@ -388,6 +389,49 @@ export async function requestNewHomeData(simulator: Simulator, requestAbortContr
 	await refreshHomeData(simulator, false, requestAbortController)
 }
 
+export async function requestBootstrapData(simulator: Simulator): Promise<PopupBootstrapData> {
```
MicahZoltu: The `.catch()` on `getTabState(tabId)` silently falls back to a default `TabState` with `signerConnected: false` and empty accounts. If `getTabState` fails for a valid tab (e.g., transient storage error), this could cause the popup to display an incorrect state — hiding the presence of a connected signer. Consider logging or surfacing the error rather than silently replacing it with defaults.

app/ts/background/background.ts
```diff
@@ -456,11 +452,12 @@ export async function popupMessageHandler(
 				case 'popup_openAddressBook': return await openNewTab('addressBook')
 				case 'popup_changeChainReadyAndListening': return await updateChainChangeViewWithPendingRequest()
 				case 'popup_interceptorAccessReadyAndListening': return await updateInterceptorAccessViewWithPendingRequests()
-				case 'popup_confirmTransactionReadyAndListening': {
-					await updateConfirmTransactionView(simulator)
-					return
-				}
-				case 'popup_requestNewHomeData': return await requestNewHomeData(simulator, simulationAbortController)
+					case 'popup_confirmTransactionReadyAndListening': {
```
MicahZoltu: The indentation of the switch cases changed from single-tab to double-tab inconsistently — `popup_confirmTransactionReadyAndListening`, `popup_requestBootstrapData`, and `popup_requestNewHomeData` are now indented further than their sibling cases. This makes the switch structure harder to read and could suggest an accidental nesting issue. Normalize the indentation.

app/ts/background/settings.ts
```diff
@@ -98,7 +98,19 @@ export const getDefaultBlockExplorer = (): BlockExplorer => ({ apiUrl: `https://
 
 export const getWethForChainId = (chainId: bigint) => wethForChainId.get(chainId.toString())
```
MicahZoltu, line 100: The `cachedSettings` cache is susceptible to interleaved updates. Functions like `setPage` and `changeSimulationMode` read the full settings object, modify a field, and eagerly replace the entire cache entry. If two async setters run concurrently, the second cache write can overwrite the first, leaving the in-memory cache inconsistent with storage until the next full read.

app/ts/background/storageVariables.ts
```diff
@@ -193,16 +193,37 @@ export async function updateEthereumSubscriptionsAndFilters(updateFunc: (prevSta
 	})
 }
 
-export const setRpcList = async(rpcEntries: RpcEntries) => await browserStorageLocalSet({ rpcEntries })
+let cachedRpcEntries: RpcEntries | undefined
```
MicahZoltu: Similar read-modify-write race for `cachedRpcEntries`. Without explicit serialization, concurrent calls to `promoteRpcAsPrimary` or `setRpcList` may interleave and produce a stale cached view of the RPC list.

MicahZoltu (commented):
## CI Agent Review

In-memory caching for settings and RPC lists introduces a potential cache coherency race condition. Multiple concurrent async updates can interleave read-modify-write operations, allowing one update to stale the full cached object over another because setters cache complete state objects rather than merging partial updates.

MicahZoltu (approved):
Approving to clear the PR backlog, but we should really fix the issue in `app/ts/types/interceptor-reply-messages.ts:207` as it seems pretty bad.

MicahZoltu (commented):
## CI Agent Review

The new `applyBootstrapData` function in App.tsx duplicates the signal-setting logic already present in `updateHomePage`, creating a forked data flow for home page state. The `BootstrapSettings` type is a structurally incomplete copy of `Settings`, so any new field added to `Settings` must be manually propagated to `BootstrapSettings`. This parallel initialization path will likely diverge over time, increasing maintenance burden.

app/ts/background/storageVariables.ts
```diff
@@ -185,16 +185,37 @@ export async function updateEthereumSubscriptionsAndFilters(updateFunc: (prevSta
...
```
MicahZoltu: Incomplete caching pattern: module-level mutable caches (`cachedRpcEntries`, `cachedRpcEntriesPromise`) are introduced along with an exported `resetRpcListCache()` function, but `resetRpcListCache` is never called anywhere in the codebase. The cache can diverge from storage if writes bypass `setRpcList` or if storage is modified externally (e.g., browser sync). Either remove the unused reset function or complete the pattern with proper invalidation at all relevant call sites.

app/ts/types/interceptor-reply-messages.ts
```diff
@@ -18,6 +22,33 @@ export const EnrichedRichListElement = funtypes.ReadonlyObject({
 	type: funtypes.Union(funtypes.Literal('PreviousActiveAddress'), funtypes.Literal('UserAdded'), funtypes.Literal('CurrentActiveAddress')),
 })
 
+const BootstrapSettings = funtypes.ReadonlyObject({
```
MicahZoltu: Duplicated settings schema: `BootstrapSettings` duplicates field definitions from the canonical `Settings` runtype in `interceptor-messages.ts`. Due to circular dependency constraints the duplication may be unavoidable, but there is no mechanism (comment, shared sub-schema, or re-export) to signal the structural coupling. Adding a field to `Settings` will silently omit it from bootstrap data, creating a maintenance hazard. Consider extracting a shared sub-schema that both modules can reference.

app/ts/types/interceptor-reply-messages.ts
```diff
@@ -163,9 +211,25 @@ export const RequestAbiAndNameFromBlockExplorer = funtypes.ReadonlyObject({
 	data: funtypes.ReadonlyObject({ address: EthereumAddress, chainId: ChainIdWithUniversal })
 }).asReadonly()
 
-export const PopupMessageReplyRequests = funtypes.Union(
+export type PopupRequests =
```
MicahZoltu: Manual type definitions decoupled from runtypes: `PopupRequests` and `PopupReplyOption` are now manually maintained union types instead of being derived via `funtypes.Static<typeof ...>`. Adding a new message type now requires three coordinated changes (manual type, runtype entry, and reply mapping entry) with no mechanical guarantee of alignment beyond the `Codec<T>` backstop. This increases ceremony and risk of omission errors when extending the protocol.

MicahZoltu (commented):
## CI Agent Review

Architectural concerns identified: an incomplete caching pattern introduces global mutable state without proper lifecycle management, a duplicated settings schema has no compile-time link to the canonical definition, and manual type definitions are now separated from their runtype sources, increasing maintenance burden for future protocol changes.

MicahZoltu (commented):
## CI Agent Review

No issues found.

MicahZoltu (approved):

PR #1390, opened 2026-04-20
Simulate Safe transactions on the active stack

### LINKED ISSUE #1249: simulate gnosis safe transaction simulations on top of the stack and not on top of mainnet
the bug was already there. I'll create issue for it

_Originally posted by @KillariDev in https://github.com/DarkFlorist/TheInterceptor/pull/1247#discussion_r2022643932_

KillariDev:
## Summary
- Adds Safe transaction simulation against the current stack so the popup reflects the active execution context.
fix https://github.com/DarkFlorist/TheInterceptor/issues/1249


app/ts/background/simulationUpdating.ts
```diff
@@ -147,11 +144,12 @@ export const simulateGovernanceContractExecution = async (pendingTransaction: Pe
 		const parentBlock = await ethereum.getBlock(undefined)
 		if (parentBlock === null) throw new Error('The latest block is null')
 		if (parentBlock.baseFeePerGas === undefined) return returnError('cannot build simulation from legacy block')
+		const simulationInput = await getCurrentSimulationInput()
 		const signedExecutionTransaction = mockSignTransaction({ ...contractExecutionResult.executingTransaction, gas: contractExecutionResult.ethSimulateV1CallResult.gasUsed })
 		const tokenBalancesAfter = await getTokenBalancesAfterForTransaction(
 			ethereum,
 			undefined,
-			[], // we are simulating on top of mainnet, not top of our stack. Fix to simulate on right place of the stack
+			simulationInput,
```
MicahZoltu, line 224: I'm curious why this was previously not fixed?  Is the fix more than this one change?

test/tests/gnosisSafeStackSimulation.ts
```diff
@@ -0,0 +1,306 @@
+// @ts-nocheck
```
MicahZoltu: We should always be type checking.  If this nocheck is necessary it should come with a very clear comment why removing it is impossible or an unrealistic amount of effort.

```suggestion
```

test/tests/gnosisSafeStackSimulation.ts
```diff
@@ -0,0 +1,306 @@
...
+		assert.equal(reply.result.visualizedSimulationState.visualizedBlocks.length, 2)
+		assert.equal(reply.result.visualizedSimulationState.visualizedBlocks[0]?.simulatedAndVisualizedTransactions.length, 1)
+		assert.equal(reply.result.visualizedSimulationState.visualizedBlocks[1]?.simulatedAndVisualizedTransactions.length, 1)
+	})
+}
+
+await runIfRoot(async () => {
+	await main()
+	await run()
+}, import.meta)
```
MicahZoltu: Why is this runnable outside of the normal test suite?  Why not just have it be a normal test?

test/tests/gnosisSafeStackSimulation.ts
MicahZoltu, line 1: This change is, presumably, just fixing a very tiny bug.  Why does it require adding a weird/novel testing mechanism?  While SAFE transactions are a bit novel internally, they aren't novel at the level that this unique testing methodology is implying.  It feels like they should be testable using existing testing harness and mocks without needing to construct whole new testing constructs.

MicahZoltu (approved):
The bugfix is fine, but the new test feels quite inconsistent with existing testing strategies and feels like a mistake in multiple ways.

KillariDev:
fixed your comments. And you were right to call out that fixing it was not simple enough. That was improvement, but token balance updates were not being run correctly, now they should be

MicahZoltu (commented):
## CI Agent Review

No issues found across all agents.

MicahZoltu (approved):

MicahZoltu (commented):
## CI Agent Review

No security issues found.

PR #1394, opened 2026-04-23
Add browser simulator popup lifecycle benchmark suite

KillariDev:
## Summary
- add a real-Chrome browser simulator popup lifecycle benchmark suite
- add popup and background performance marks plus RPC timing collection used by the benchmark
- document the scenario boundaries and report both end-to-end stacked timings and main-popup-only timings

## Validation
- npm test
- npm run setup-chrome
- npm run benchmark:popup-lifecycle

### DIFF
--- a/app/ts/background/background-startup.ts
+++ b/app/ts/background/background-startup.ts
@@ -23,6 +23,7 @@ import { AddressBookEntries, AddressBookEntry } from '../types/addressBookTypes.
 import { getUniqueItemsByProperties } from '../utils/typed-arrays.js'
 import { updateDeclarativeNetRequestBlocks } from './accessManagement.js'
 import { updatePopupVisualisationIfNeeded } from './popupVisualisationUpdater.js'
+import { POPUP_PERFORMANCE_MARKS, markPerformance } from '../utils/popupPerformance.js'
 
 const websiteTabConnections = new Map<number, TabConnection>()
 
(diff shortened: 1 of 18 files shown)


MicahZoltu (changes requested):
Since it sounds like this benchmark can use a headless Chromium, we should package it up into a Docker container so it is portable across environments.  `scripts/install-chrome.sh` can be changed to a Dockerfile, and running the benchmark can be as simple as building the docker image and then running it.  This would also make it so we could reliably run the benchmarks in a GitHub action, as I am skeptical that the shell install script will work nicely in CI directly (definitely not portable).

Importantly, adding that script outside of a docker container is the first thing that makes this project not compatible with a Windows development environment.  That is a bright line that we haven't crossed so far, and I would like to avoid crossing if we can (and I think we can avoid crossing it in this case).

KillariDev:
> Since it sounds like this benchmark can use a headless Chromium, we should package it up into a Docker container so it is portable across environments. `scripts/install-chrome.sh` can be changed to a Dockerfile, and running the benchmark can be as simple as building the docker image and then running it. This would also make it so we could reliably run the benchmarks in a GitHub action, as I am skeptical that the shell install script will work nicely in CI directly (definitely not portable).
> 
> Importantly, adding that script outside of a docker container is the first thing that makes this project not compatible with a Windows development environment. That is a bright line that we haven't crossed so far, and I would like to avoid crossing if we can (and I think we can avoid crossing it in this case).

I am already running it in docker, not in a docker in a docker thought.

I would not add this stuff to CI, just for the agent to run and test its changes. The project is still comptible with windows development environment, it just cannot run this feature, which is far from mandatory

MicahZoltu (approved):
Still not a fan, but I understand the problem better now.  You want your agent (which currently operates inside a container) to be able to build/run docker stuff.  In theory, Docker Sandbox is the right solution for this but I looked into it briefly when it launched and I'm not impressed so I don't think we should move over to that.

PR #1400, opened 2026-04-25
Migrate tests and devcontainer to Bun

KillariDev:
## Summary
- switch the repo package-manager metadata and contributor guidance from npm to Bun
- remove the custom test runner helpers and run the active suite directly with `bun:test`
- add Bun provisioning for the devcontainer and update related benchmark/build messaging

## Validation
- bun test
- bun run setup-chrome


.devcontainer/devcontainer.json
```diff
@@ -1,22 +1,7 @@
...
-	// "customizations": {},
-
-	// Uncomment to connect as root instead. More info: https://aka.ms/dev-containers-non-root.
-	// "remoteUser": "root"
+	"name": "The Interceptor",
+	"build": {
+		"dockerfile": "Dockerfile"
+	},
+	"remoteUser": "node"
 }
```
MicahZoltu, line 10:
```suggestion
{
	"image": "mcr.microsoft.com/devcontainers/javascript-node:1-20-bullseye",
	"features": {
		"ghcr.io/devcontainers-extra/features/bun:1": {
			"version": "1.3.12"
		}
	}
}
```

.devcontainer/Dockerfile
MicahZoltu, line 1: Recommend deleting this in favor of using the bun extra.  In both cases you can't protect against supply chain attacks, sadly, because neither the `bun.sh` installer nor devcontainer features support hashed checks, but I feel like at least the devcontainer feature is simpler and faster.

test/tests/addNewAddress.test.ts
```diff
@@ -11,7 +11,7 @@ const sampleAddressBookEntry = {
 
 export async function main() {
 	describe('add new address save flow', () => {
-		should('waits for the save message to finish before closing the popup', async () => {
+		test('waits for the save message to finish before closing the popup', async () => {
```
MicahZoltu: This is a weird way to write tests.  Normally the pattern is:
```
describe('name', () => {
	test('sub-name', () => {
		// test code goes here
	})
})
```

Why do you have everything wrapped in a `main` function that is then called when the file is run?  See https://bun.com/docs/test/writing-tests for details.

bunfig.toml
```diff
@@ -0,0 +1,3 @@
+[install]
+exact = true
+ignoreScripts = true
```
MicahZoltu: Bun defaults to false for this, so it is unnecessary.  https://bun.com/docs/runtime/bunfig#install-ignorescripts

MicahZoltu (approved):
I strongly recommend making the tests follow standard testing patterns unless there is a really good reason to deviate like this.

bunfig.toml
KillariDev: if it defaults to false, isnt't it necessary so we don't use defaults?

MicahZoltu: Hmm, you are right.  For some reason I really thought that Bun didn't have the same problem as NPM with regards to running scripts automatically, but a quick internet search plus the documentation suggests that isn't the case.  This belief caused me to mentally flip the flag to `--runScripts` where a default of `false` would be desirable.

Though, this has lead me to an even bigger problem.  😖  https://www.bunsecurity.dev/blog/bun-security-vulnerability-insecure-practice-package-manager/

PR #1401, opened 2026-04-25
If rpc method doesnt use simulation results, just use simulation inputs

KillariDev:
## Summary
- split simulation-backed RPC handling between prepared simulation input and full simulation state
- move simulated block, transaction, receipt, log, filter, and `newHeads` behavior onto a consistent execution-block view
- derive deterministic simulated block hashes from prepared input and add regression coverage for subscriptions and gas-split execution blocks

## Testing
- npm test
- npm run setup-chrome


app/ts/simulation/services/SimulationModeEthereumClientService.ts
```diff
@@ -35,6 +36,49 @@ type GroupedEthSimulateV1BlockResult = {
...
+
+type SimulationInputInspection = {
+	success: true
+	base: SimulationInspectionBase
+	groupedEthSimulateV1CallResult: GroupedEthSimulateV1Result
+} | {
+	success: false
+	base: SimulationInspectionBase
+	jsonRpcError: ReturnType<JsonRpcResponseError['serialize']>
+	}
```
KillariDev:
```suggestion
}
```

MicahZoltu (approved):
Nothing looks malicious, but I don't understand what this change is doing/enabling.  Can you expand on what is happening here?

KillariDev:
> Nothing looks malicious, but I don't understand what this change is doing/enabling. Can you expand on what is happening here?

Previously we run simulator and generate everything in that (calculate all pre and post eth balances, same for tokens, calculate simulator results etc), every block. All the RPC methods are waiting until that is run and then we run their queries. That's completely unnecessary, to reply some eth_call, we just run a single eth_simulate to get the results, so there's no need to wait for that big simulation. Also that big simulation is not anyway needed unless the user opens the extension popup, and then we can just update that when it happens

PR #1404, opened 2026-04-27
Handle cleared popup simulation state

KillariDev:
## Summary
- stop the popup visualisation from throwing when clear resets simulation results to an empty state
- keep the existing Dino empty-state UX while safely narrowing results before passing them to simulation detail components
- add a regression test that mounts Home, clears the simulation results, and asserts the popup rerenders instead of crashing

## Testing
- npm test
- npm run setup-chrome


MicahZoltu (approved):

bunfig.toml
```diff
@@ -0,0 +1,2 @@
+[install]
+exact = true
```
MicahZoltu, line 2: ignoreScripts = true

PR #1406, opened 2026-04-27
Tighten TypeScript strictness and agent guidance

KillariDev:
## Summary
- enable stricter TypeScript compiler checks in the app and inpage configs
- update the codebase to satisfy exact optional property and index signature enforcement
- add AGENTS guidance banning TypeScript directive comments, unsafe casting, null-preference, and class-based patterns

## Testing
- npm test
- npm run setup-chrome


app/ts/background/background.ts
```diff
@@ -277,17 +277,25 @@ export async function changeActiveAddressAndChain(
 	websiteTabConnections: WebsiteTabConnections,
 	change: {
 		simulationMode: boolean,
-		activeAddress?: bigint,
-		rpcNetwork?: RpcNetwork,
+		activeAddress?: bigint | undefined,
+		rpcNetwork?: RpcNetwork | undefined,
```
MicahZoltu: These were already optional properties.  Why are you making them optional | undefined (which is *mostly* redundant?

app/ts/components/pages/WebsiteAccess.tsx
```diff
@@ -232,14 +232,16 @@ const WebsiteAccessOverview = ({ websiteAccess, checked }: WebsiteAccessOverview
...
 	switch (status) {
 		case 'blocked':
 			return <span class = 'status-warn' role = 'img' title = 'External Request Blocked' aria-label = 'External Request Blocked'><RequestBlockedIcon /></span>
 		case 'disabled':
 			return <span class = 'status-danger' role = 'img' title = 'Protection Disabled' aria-label = 'Protection Disabled'><InterceptorDisabledIcon /></span>
 		case undefined:
-			return <></>
+			return null
+		default:
+			return null
```
MicahZoltu, line 244: This seems like a bad change.  Why return null instead of empty fragment?  It changes this function from something that returns `Fragment` to `Fragment | null` which seems worse.

app/ts/components/pages/WebsiteAccess.tsx
```diff
@@ -232,14 +232,16 @@ const WebsiteAccessOverview = ({ websiteAccess, checked }: WebsiteAccessOverview
 	)
 }
 
-const SiteStatusIndicator = ({ status }: { status?: 'disabled' | 'blocked' }) => {
+const SiteStatusIndicator = ({ status }: { status?: 'disabled' | 'blocked' | undefined }) => {
```
MicahZoltu: Same question as above, this was already optional, why are you adding `| undefined` to it.  In this case, I think it is *entirely* redundant.

MicahZoltu: Looks like there are a bunch of these, same question for all of them.

app/ts/simulation/compoundGovernanceFaking.ts
```diff
@@ -101,14 +101,14 @@ export const simulateCompoundGovernanceExecution = async (ethereumClientService:
 }
 
 export const parseVoteInputParameters = (ethersResult: Result) => {
-	if (ethersResult.proposalId === undefined) throw new Error('proposal Id missing from vote call')
-	if (ethersResult.support === undefined) throw new Error('support missing from vote call')
+	if (ethersResult['proposalId'] === undefined) throw new Error('proposal Id missing from vote call')
+	if (ethersResult['support'] === undefined) throw new Error('support missing from vote call')
```
MicahZoltu, line 105: Why the style change here?  I'm not sure if there is a functional or performance difference between these two styles, but the original is the more commonly used style.

Same question for all instances of this change.

MicahZoltu (approved):

app/ts/components/pages/WebsiteAccess.tsx
KillariDev: Null is Fragment type. Null makes it that the element doesnt render, <></> renders empty.

KillariDev: https://react.dev/learn/conditional-rendering#conditionally-returning-nothing-with-null](https://react.dev/learn/conditional-rendering#conditionally-returning-nothing-with-null)

PR #1407, opened 2026-04-27
Harden make-me-rich popup replies

KillariDev:
## Summary
- treat raw `null` popup replies as `undefined` instead of trying to parse them as `PopupReply`
- make `popup_requestMakeMeRichData` resilient to one bad rich-list entry by falling back per address instead of failing the whole reply
- recover corrupt persisted `fixedAddressRichList` data by resetting it to `[]`
- add regression coverage for the popup reply path and make-me-rich request handling

## Reproduction
On `v1.0.19`, the popup can surface an unexpected error starting with:

`Unable to assign null to { readonly type: "RequestMakeMeRichDataReply" ... }`

The underlying failure mode is:
- a background failure during `popup_requestMakeMeRichData` can result in Chrome returning a top-level `null` reply to the popup
- the popup currently attempts to parse that `null` as a `PopupReplyOption`
- a single failing `identifyAddress(...)` call in the fixed rich list can also break the whole request

I reproduced both code paths directly against `v1.0.19`:
- mocked `browser.runtime.sendMessage(...)` returning `null` produces the reported parse-family error
- one failing rich-list address causes `requestMakeMeRichList(...)` to throw

## Changes
- normalize raw `null` replies in `sendPopupMessageWithReply(...)`
- resolve rich-list entries independently and emit a fallback `contact` entry when address identification fails
- preserve the real failure by reporting the unexpected error with the failing address context
- reset corrupt stored rich-list data instead of crashing the request path

## Validation
- `npm test`
- `npm run setup-chrome`


MicahZoltu (commented):
## CI Agent Review

No security issues found.

MicahZoltu (commented):
## CI Agent Review

No security issues found.

MicahZoltu (approved):

PR #1409, opened 2026-04-28
Refactor log summarization to a functional API

KillariDev:
## Summary
- replace the `LogSummarizer` class with direct functional summary APIs
- update simulation summary consumers to use the new functions
- keep fallback address metadata behavior from `main` and add/port summarizer coverage


MicahZoltu (approved):
I do like me some functional.

PR #1411, opened 2026-04-28
Refactor simulation service wiring out of simulator

KillariDev:
## Summary
- remove the `Simulator` wrapper and thread `ethereum`/`tokenPriceService` directly
- split simulator responsibilities into parsing, protector, and service lifecycle modules
- update background flows and tests to the direct-service API

## Validation
- npm test
- npm run setup-chrome


MicahZoltu (commented):
## CI Agent Review

This PR replaces the monolithic `Simulator` class with independently managed services (`EthereumClientService`, `TokenPriceService`) and a `ResetSimulationServices` callback, and extracts parsing and protector logic into separate files. While the extraction is clean, the service lifecycle changes introduce several concerns:

1. **Parameter explosion**: Every background function that previously received `Simulator` now accepts three separate arguments (`ethereum`, `tokenPriceService`, `resetSimulationServices`). Many handlers (especially in `providerMessageHandlers.ts`) pass unused parameters, increasing wiring complexity and maintenance burden.

2. **Closure‑based service access**: `onContentScriptConnected` takes a lambda `getCurrentSimulationServices` instead of reading a module variable directly. This indirection adds unnecessary complexity.

3. **Race condition in `newBlockAttemptCallback`**: The guard `if (ethereumClientService !== getSimulationServices().ethereum) return` is checked only at entry; subsequent async work could operate on a stale service after a reset. Either snapshot the service or keep it alive until the callback finishes.

4. **Testing impact**: The separate argument pattern makes tests pass minimal mocks; any parameter list change will require updates across many test files.

**Recommendations**:
- Keep a single service-bundle object (e.g., `SimulationResources`) passed as one argument.
- Eliminate the `getSimulationServices` closure; read the module variable directly.
- Fix the race condition by saving a reference to the service at callback entry.
- Remove unused parameters from provider message handlers.
- Define a helper type to reduce argument-count explosion.

The `parsing.ts` and `protectorRunner.ts` extraction is a good readability improvement and should be retained.

MicahZoltu (commented):
## CI Agent Review

No security issues found.

MicahZoltu (approved):

PR #1412, opened 2026-04-28
Fix subscription removal to only affect the matching socket

KillariDev:
## Summary
- remove only the matching socket subscription instead of dropping unrelated entries
- tighten the success check so it verifies the same socket/id pair
- add a regression test for socket-scoped removal

## Validation
- npm test
- npm run setup-chrome


MicahZoltu (commented):
## CI Agent Review

No security issues found.

MicahZoltu (commented):
## CI Agent Review

No security issues found.

MicahZoltu (approved):

PR #1413, opened 2026-04-28
Fix eth_getFilterChanges polling state and filter payloads

KillariDev:
## Summary
- seed filters from the simulated head instead of assuming a single extra block
- pass the stored filter payload through to simulated log lookup
- preserve non-matching subscriptions and advance filter state correctly
- add regression coverage for repeated filter polling

## Validation
- npm test
- npm run setup-chrome


MicahZoltu (commented):
## CI Agent Review

No security issues found.

MicahZoltu:
That was a lie.  Agents didn't finish running due to a bug (now hopefully fixed to hard fail instead).

PR #1419, opened 2026-04-29
Fix interceptor clear crash from undefined simulation state

KillariDev:
## Summary
- pass a stable simulation results snapshot through the popup results subtree instead of an optional live signal
- update the related transaction and summary components, plus confirm/custom visualizer callers, to consume value-based simulation results consistently
- add the clear-empty-state regression test to the normal Bun test suite and extend the DOM mock support it needs

## Validation
- bun test
- bun run setup-chrome


MicahZoltu (commented):
## CI Agent Review

No security issues found.

MicahZoltu (approved):

MicahZoltu (commented):
## CI Agent Review

No security issues found.

MicahZoltu (approved):

PR #1420, opened 2026-04-29
Fix nonce shifting when removing queued transactions

KillariDev:
## Summary
- only shift later queued nonces for interceptor-assigned `eth_sendTransaction` entries
- leave `eth_sendRawTransaction` entries and same-nonce replacements unchanged when removing an earlier transaction
- add regression coverage for raw, shifted, different-sender, and same-nonce cases

## Validation
- npm test
- npm run setup-chrome


MicahZoltu (commented):
## CI Agent Review

No security issues found.

MicahZoltu (approved):

PR #1422, opened 2026-04-29
Forward interceptor diagnostics through unexpected errors

KillariDev:
## Summary
- serialize `InterceptorError` reports into clone-safe diagnostics before they cross the inpage/content-script boundary
- surface forwarded diagnostics through unexpected error handling as a single verbose popup message
- add regression coverage for forwarded interceptor error reporting and popup rendering

## Validation
- bun test
- bun run setup-chrome


MicahZoltu (commented):
## CI Agent Review

A malicious dapp can inject arbitrary text into the Interceptor's trusted error popup via spoofed InterceptorError messages. The new normalizeUnexpectedError function specifically recognizes InterceptorError objects and extracts params[0] as the message to display in the popup. Before this change, a spoofed message like { method: 'InterceptorError', params: ['phishing text'] } would not have a .message property and would fall through to the safe default ('Please see The Interceptors console for more details on the error.'). After this change, the attacker-controlled string is displayed directly in the extension's error popup — a trusted UI surface where users expect legitimate extension communications, enabling social engineering/phishing attacks.

app/ts/utils/errors.ts
```diff
@@ -41,8 +41,25 @@ export function isFailedToFetchError(error: Error) {
 
 export const isNewBlockAbort = (error: Error) => error.message?.includes(NEW_BLOCK_ABORT)
 
+function getForwardedDiagnostics(error: unknown): string | undefined {
```
MicahZoltu, line 52: Security vulnerability: The new getForwardedDiagnostics and normalizeUnexpectedError functions extract params[0] from any InterceptorError-typed object and display it directly in the popup. A malicious dapp can craft a window.postMessage with interceptorRequest: true, method: 'InterceptorError', and arbitrary text in params[0]. The content script forwards this to the background without origin verification, and the new code path displays the attacker's text in the extension's trusted error popup. Before this change, the old error handling code would not display the attacker's text because the object lacks a .message property and falls through to a safe default. Consider validating the source/origin of InterceptorError messages, or sanitizing/constraining what can be displayed from forwarded diagnostics.

MicahZoltu (changes requested):

MicahZoltu (commented):
## CI Agent Review

No issues found.

MicahZoltu (commented):
## CI Agent Review

No issues found.

MicahZoltu (approved):
This one took forever.  4000 lines of bug hunter thinking/tool calling.

PR #1423, opened 2026-05-04
Preserve interceptor state when RPC stays on same chain

KillariDev:
## Summary
- preserve the interceptor stack when switching RPCs within the same chain
- keep the destructive reset path for cross-chain RPC changes
- add regression coverage for same-chain preserve and cross-chain reset behavior

## Testing
- bun test
- bun run setup-chrome


MicahZoltu (commented):
## CI Agent Review

No security issues found.

MicahZoltu (approved):

PR #1424, opened 2026-05-04
Handle favicon fetch failures explicitly

KillariDev:
## Summary
- make favicon loading best-effort and return explicit failure reasons from `imageToUri`
- simplify `retrieveWebsiteDetails` so favicon failures fall back to `icon: undefined`
- add focused tests for favicon fetch failures and website detail retrieval

## Validation
- bun test
- bun run setup-chrome


MicahZoltu (commented):
## CI Agent Review

The `file:` protocol in `ALLOWED_FAVICON_PROTOCOLS` combined with the null content-type bypass in `imageToUri` creates a local file read vulnerability. When a `file://` URL is fetched, the response has no Content-Type header (returns `null`), which skips the newly-added image content-type validation entirely. This allows arbitrary local files to be fetched, converted to data URIs, and persisted in extension storage.

app/ts/background/iconHandler.ts
```diff
@@ -11,6 +11,8 @@ import { getLastKnownCurrentTabId } from './popupMessageHandlers.js'
 import { checkAndPrintRuntimeLastError, doesTabExist, safeGetTab, silenceChromeUnCaughtPromise } from '../utils/requests.js'
 import { modifyObject } from '../utils/typescript.js'
 
+const ALLOWED_FAVICON_PROTOCOLS = new Set(['http:', 'https:', 'file:', 'data:'])
```
MicahZoltu: The `file:` entry in `ALLOWED_FAVICON_PROTOCOLS` enables a local file read path. When a user opens a malicious local HTML file that sets a `<link rel="icon" href="...">` pointing to a sensitive local file (e.g., `/etc/passwd` or private keys), the extension will fetch that file. Combined with the null content-type bypass in `imageToUri.ts`, the file's contents are read without validation that it is actually an image, converted to a data URI, and stored in extension state (where it persists in `websiteAccess` settings and could be exported). Either remove `file:` from this allowlist, or add explicit image-type validation for `file://` responses (e.g., check file extension or validate the blob).

app/ts/utils/imageToUri.ts
```diff
@@ -1,27 +1,40 @@
...
+	reader.readAsDataURL(blob)
+	return await future
+}
+
+export async function imageToUri(url: string, maxSizeInBytes = 1048576): Promise<ImageToUriResult> {
 	try {
 		const response = await fetch(url)
+		if (!response.ok) return imageToUriFailed(`HTTP ${ response.status }${ response.statusText === '' ? '' : ` ${ response.statusText }` }`)
+		const contentType = response.headers.get('content-type')
+		if (contentType !== null && !contentType.startsWith('image/')) return imageToUriFailed(`response was not an image (${ contentType })`)
```
MicahZoltu, line 30: The condition `contentType !== null && !contentType.startsWith('image/')` skips content-type validation when `contentType` is `null`. This is exactly the case for `file://` URL responses, which typically have no Content-Type header. Since `file:` is explicitly allowed in `ALLOWED_FAVICON_PROTOCOLS`, any local file — regardless of type — will pass this check and be converted to a data URI. Consider rejecting null content-type responses when the URL scheme is `file:`, or change the logic to `if (contentType === null || !contentType.startsWith('image/'))` to require an explicit image content type.

MicahZoltu (changes requested):
Let me know if you think the security vulnerability reported is legitimate or not.  I'm skeptical, but would like you to review it.

MicahZoltu (commented):
## CI Agent Review

No issues found.

MicahZoltu (approved):

PR #1426, opened 2026-05-04
Remove forced gas limits from simulation calls

KillariDev:
## Summary
- stop injecting near-block gas limits into synthetic `eth_simulateV1` estimate and call flows when the caller omitted gas
- route synthetic estimate, call, code lookup, and balance multicall requests through an appended raw block-call helper
- update the Safe `baseGas === 0` fallback to use shared simulated gas estimation and add regression coverage for omitted-gas behavior

## Testing
- bun test
- bun run setup-chrome


MicahZoltu (commented):
## CI Agent Review

No security issues found.

MicahZoltu (approved):

PR #1427, opened 2026-05-06
Handle null popup replies and return visualized simulation replies

KillariDev:
## Summary
- return the reply payload for `popup_requestCompleteVisualizedSimulation`
- treat `null` popup runtime replies as no reply instead of unexpected errors
- remove the generic popup reply cast and add regression coverage for both paths

## Testing
- npm test
- npm run setup-chrome


app/ts/background/backgroundUtils.ts
```diff
@@ -59,9 +59,19 @@ export async function sendPopupMessageToBackgroundPage(message: PopupMessage) {
...
+function parsePopupRequestReply<Method extends PopupRequestMethod>(method: Method, rawReply: unknown) {
+	return PopupRequestsReplies[method].parse(rawReply)
+}
+
+export async function sendPopupMessageWithReply<Method extends PopupRequestMethod>(message: PopupRequestByMethod<Method>): Promise<PopupReplyByMethod<Method> | undefined> {
 	try {
-		return PopupReplyOption.parse(await browser.runtime.sendMessage(PopupMessageReplyRequests.serialize(message))) as PopupRequestsReplyReturn<Request>
+		const rawReply = await browser.runtime.sendMessage(PopupMessageReplyRequests.serialize(message))
+		if (rawReply === null) return undefined
+		return parsePopupRequestReply(message.method, rawReply) as PopupReplyByMethod<Method>
```
MicahZoltu: 😢

MicahZoltu (approved):
Approved because this is a live bug a user is running into.  That typecast bothers me though.

PR #1431, opened 2026-05-19
Fix block cache expiry

KillariDev:
## Summary
- fix the cached block TTL check to use consistent millisecond units
- add a regression test proving stale cached blocks are discarded

## Validation
- bun test
- bun run setup-chrome


MicahZoltu (commented):
## CI Agent Review

No security issues found.

MicahZoltu (approved):

PR #1432, opened 2026-05-19
Continue newHeads delivery after stale subscriptions

KillariDev:
## Summary
- continue iterating subscriptions after removing a stale entry
- add a regression test proving later live subscribers still receive newHeads

## Validation
- bun test
- bun run setup-chrome


MicahZoltu (commented):
## CI Agent Review

No issues found.

MicahZoltu (approved):
The AI spent *way* more processing power on its PR review of this diff than I would have.  😆

KillariDev:
> The AI spent _way_ more processing power on its PR review of this diff than I would have. 😆

dont micromanage it! let it work :P

PR #1433, opened 2026-05-19
Ignore stale address book replies

KillariDev:
## Summary
- match address book replies against the full active view filter instead of only chainId
- add a regression test for stale reply rejection

## Validation
- bun test
- bun run setup-chrome


MicahZoltu (commented):
## CI Agent Review

No security issues found.

MicahZoltu (commented):
## CI Agent Review

No issues found.

MicahZoltu (approved):

PR #1434, opened 2026-05-19
Fix confirm dialog force-send gating

KillariDev:
## Summary
- fix signable message confirm gating to respect the `forceSend.value` signal
- extract the signable-message disable logic into a small helper
- add a regression test covering the unsupported-message force-send path

## Validation
- bun test
- bun run setup-chrome


MicahZoltu (commented):
## CI Agent Review

No issues found.

MicahZoltu (approved):

PR #1435, opened 2026-05-19
Fix interceptor access modal selection

KillariDev:
## Summary
- select modal props from the pending access request matching the stored `accessRequestId`
- keep the modal inactive when the selected request is no longer pending instead of reading from the first queued request
- add a focused regression test for pending-request selection

## Validation
- bun test
- bun run setup-chrome


MicahZoltu (commented):
## CI Agent Review

No issues found.

MicahZoltu (approved):

PR #1436, opened 2026-05-19
Fix common token router warning

KillariDev:
## Summary
- fix `commonTokenOops` so it checks the transferred token contract metadata instead of the router recipient
- treat ERC20, ERC721, or ERC1155 metadata as sufficient for the warning to trigger
- add focused regression coverage for direct transfers of a known token to the Uniswap V2 router

## Validation
- bun test
- bun run setup-chrome


MicahZoltu (commented):
## CI Agent Review

No issues found.

MicahZoltu (approved):

PR #1437, opened 2026-05-19
Fix website access selection state

KillariDev:
## Summary
- bind the website access radio group to the selected domain state instead of defaulting to the first row
- update selection through the radio input so clicks and keyboard changes both drive the URL hash
- add regression coverage for hash-driven selection state

## Validation
- bun test
- bun run setup-chrome


MicahZoltu (commented):
## CI Agent Review

No issues found.

MicahZoltu (approved):

PR #1438, opened 2026-05-19
Fix concurrent requestAccounts replies

KillariDev:
## Summary
- preserve the real signer outcome while a requestAccounts prompt is already pending
- replay the same success or error reply to later concurrent callers instead of always reporting success
- add regression coverage for a rejected concurrent requestAccounts flow

## Validation
- bun test
- bun run setup-chrome


MicahZoltu (commented):
## CI Agent Review

No issues found.

MicahZoltu (approved):

PR #1440, opened 2026-05-19
Fix unsupported chain change copy

KillariDev:
## Summary
- remove the dead "disable and change" promise from the unsupported-chain dialog
- derive the primary action state from the actual supported/unsupported behavior
- add regression coverage for supported and unsupported change-chain states

## Testing
- bun test
- bun run setup-chrome


MicahZoltu (commented):
## CI Agent Review

No security issues found.

MicahZoltu (approved):

PR #1446, opened 2026-05-21
import map refactoring

KillariDev:
## Summary
- Replace the MV3 import-map approach with a build pipeline that vendors runtime dependencies, bundles Chrome runtime entrypoints, and rewrites remaining imports to browser-safe vendored paths.
- Route extension runtime `viem`/ENS usage through a local facade so Chrome no longer hits bare module specifiers like `webextension-polyfill`, `viem/ens`, or `@noble/hashes/sha3` at runtime.
- Register service-worker listeners synchronously for MV3 startup, clean generated output before rebuilds, and update the Bun/Docker setup so local, Windows, and CI builds behave consistently.

## Why
- The previous MV3 build depended on import maps plus a post-compile rewrite pass, which could leave bare specifiers in emitted JS. In Chrome that turned into an iterative runtime failure loop where fixing one missing specifier only exposed the next one.
- The build could also be polluted by stale generated files under `app/js` and `app/inpage/js`, producing false failures locally and in Docker.
- MV3 service workers need event listeners registered at module load, so the background startup flow needed to be made explicit and synchronous at registration time.

## Testing
- `bun test`
- `bun run setup-chrome`
- `bun run setup-firefox`


MicahZoltu (commented):
## CI Agent Review

No security vulnerabilities or architectural concerns found.

MicahZoltu (approved):

PR #1448, opened 2026-05-21
Fix duplicate tab opening

KillariDev:
## Summary
- await existing-tab updates in openWebPage
- return after a successful update so a duplicate tab is not created
- add focused coverage for update, missing-tab, and update-failure paths

## Validation
- bun test
- bun run setup-chrome


MicahZoltu (commented):
## CI Agent Review

No issues found.

MicahZoltu (approved):

PR #1457, opened 2026-05-21
Harden internal provider bridge messages

KillariDev:
## Summary
- replace the public page-wide callback path with a private `MessageChannel` bridge between the inpage script and the content script
- reject reserved internal provider callback methods on the public `window.ethereum` API and only allow them through the bridge-derived internal marker in the background
- inline the inpage payload into `document_start` during the build so the hardened bridge is available at startup, and include `scripts/` in the Docker build context for that build step
- fix approved `eth_requestAccounts` flows so signer rejections and errors wake the pending waiter instead of leaving the request unresolved
- add end-to-end and regression coverage for the hardened bridge, including a real Chrome communication smoke test and benchmark harness updates for MV3 content script registration and large-state reads

## Problem Fixed
Before this change, privileged callback traffic and normal page traffic shared a public `window.postMessage` path. That meant a page could spoof internal callback-shaped messages and claim they were extension-internal.

After this change, the extension decides what is internal:
- the inpage script sends privileged callback traffic over a private `MessagePort`
- the content script only forwards validated bridge messages from that port
- the background rejects reserved callback methods unless the internal marker came from that bridge
- signer account request failures now resolve cleanly instead of hanging

This closes the original issue where an attacker could just say the callback was internal, and it also fixes the follow-on `eth_requestAccounts` waiter bug exposed during validation.

## Validation
- `bun test`
- `bun run setup-chrome`
- `bun run typecheck`
- `bun run lint`
- `bun run test:chrome-communication`


MicahZoltu (commented):
## CI Agent Review

No issues found.

MicahZoltu (approved):
# Architect

Based on my thorough analysis of every changed line and verification against the existing architecture, this diff does not introduce any structural degradation. Here is my reasoning:

**What the diff does:**

1. Replaces the broadcast `window.postMessage` channel between the inpage provider and the content script with a dedicated `MessageChannel`/`MessagePort`, scoping communication to a private channel instead of broadcasting to all page listeners.

2. Adds runtime message validation (`parseInterceptorApprovedMessage` in `inpage.ts`, `isBridgeRequest` in both content scripts) at the trust boundary between the untrusted page context and the extension, replacing unchecked type assertions (`as InterceptedRequestForward`).

3. Introduces an `interceptorInternalRequest` flag that marks internal provider callbacks (`eth_accounts_reply`, `signer_chainChanged`, etc.), then rejects public-page-originated calls to those methods at both the inpage layer and the background service worker layer.

4. Inlines `inpage.js` into `document_start.js` at build time for MV2/Firefox, eliminating a separate file load and its associated race condition.

5. Adds test infrastructure for end-to-end Chrome communication verification.

**Why this is architecturally sound:**

The MessageChannel change **improves encapsulation** — interceptor communication is now point-to-point rather than broadcast, preventing other page listeners from intercepting or interfering with provider messages. The inpage's `onWindowMessage` handler now only deals with coinbase injection detection; all interceptor traffic flows through the private port.

The `parseInterceptorApprovedMessage` type guard replaces an unsafe `as` cast at the trust boundary between the content script and the inpage provider. This is a **proper validation layer** at an architectural boundary, not a degradation.

The internal method filtering is **defense in depth** across two execution contexts (page main world and service worker) that cannot share code by browser platform design. The duplicate lists (`INTERNAL_BACKGROUND_METHODS` in `inpage.ts`, `INTERNAL_PROVIDER_METHODS` in `background.ts`) are architecturally necessary because these modules execute in isolated environments with no import capability between them.

The code duplication between `document_start.ts` and `listenContentScript.ts` is a **pre-existing constraint** of the dual-manifest architecture (MV2 vs MV3), not something this diff introduces. The diff applies identical improvements to both, which is the correct maintenance pattern — it would be an architectural regression to fix the bridge protocol in only one of the two content script entry points.

No architectural degradation found.

# Bug Hunter

After a thorough review of all changes in this diff, I find no runtime bugs that would cause incorrect behavior, crashes, or logic errors. Here is a summary of my analysis:

## What Was Reviewed

This diff implements a **MessagePort-based isolation layer** for the inpage-to-content-script bridge, replacing the previous `window.postMessage` broadcast approach. It also adds an `interceptorInternalRequest` flag to distinguish internal provider callback methods (like `eth_accounts_reply`, `signer_chainChanged`, etc.) from public dapp requests, and blocks public pages from calling these internal methods.

## Key Areas Checked

### MessagePort Bridge Setup
- The inpage script (`inpage.ts`) creates a `MessageChannel` and transfers `port2` to the content script via `window.postMessage`. The content script (`document_start.ts`, `listenContentScript.ts`) receives the port and sets up a dedicated message channel.
- **Order of operations is correct**: `listenContentScript(undefined)` runs before the inpage script is injected, so the listener is in place when the port transfer fires.
- **Queueing is safe**: `MessagePort.postMessage()` queues messages even before the receiving end has set up its handler, so early messages (like `connected_to_signer` sent during `connectToSigner`) are not lost.

### Internal Method Flagging (`interceptorInternalRequest`)
- The flag is consistently added across all layers:
  - **Inpage script** → Bridge messages marked `internal: true` for methods like `eth_accounts_reply`, `signer_chainChanged`, etc.
  - **Content script** → Forwards `interceptorInternalRequest: true` to the background when the inpage message has `internal: true`.
  - **Background startup** → Propagates `interceptorInternalRequest` from `RawInterceptedRequest` to the `InterceptedRequest` object.
  - **Background handler** → `handleInterceptedRequest` checks the flag before blocking internal methods from public access.
- **FunTypes types** (`RawInterceptedRequest`, `InterceptedRequest`) are correctly updated to accept the optional `interceptorInternalRequest` field.

### Public Blocking of Internal Methods
- The `INTERNAL_PROVIDER_METHODS` list in `background.ts` matches `INTERNAL_BACKGROUND_METHODS` in `inpage.ts` exactly.
- Public calls to these methods (via `window.ethereum.request()`) are rejected with the correct JSON-RPC error code.
- Test coverage confirms both public rejection and internal acceptance.

### `eth_accounts_reply` Error Handling (providerMessageHandlers.ts)
- The `socket` variable is now computed once before the if/else block, avoiding duplication.
- The error path now correctly sends `window_signer_accounts_changed` to wake requesters waiting for signer account resolution, even when the signer returned an error. This is a correctness improvement without regression.

### Script Inlining Mechanism
- The new `scripts/inline-inpage-document-start.mts` script correctly replaces the injection marker with the inlined source.
- The `document_start.ts` code handles both inline (`textContent`) and external (`src`) paths via the `_content === '[[injected.ts]]'` check.
- The `Dockerfile` now copies the `scripts/` directory, making the inlining step available in CI.

### Test Harness Additions
- `waitForRegisteredContentScripts` correctly polls `browser.scripting.getRegisteredContentScripts()` until expected content scripts are registered.
- `readExtensionLargeStateValue` handles both IndexedDB and `browser.storage.local` fallback, fixing the popup lifecycle benchmark for migrated state.

# Defender

No backdoor or deliberately malicious code detected in this diff. The changes represent a security hardening of the inpage-to-background communication channel:

- **MessageChannel migration**: Replaces `window.postMessage` (which any page script can inject messages into) with a dedicated `MessageChannel` port that only the extension's inpage script controls.
- **Inlining `inpage.js`**: The build script inlines the inpage script directly into the content script, eliminating the network fetch and shrinking the race window for the MessageChannel handshake.
- **Internal method filtering**: Adds `interceptorInternalRequest` flag enforcement in the background so that sensitive callback methods (`eth_accounts_reply`, `signer_reply`, `connected_to_signer`, etc.) are rejected with `METHOD_NOT_FOUND` when they originate from untrusted page scripts rather than the extension's own inpage bridge.
- **Client-side defense in depth**: The inpage script's `WindowEthereumRequest` prevents page scripts from calling internal methods through `window.ethereum.request()`.
- **Strict message parsing**: Replaces loose `as` casts with proper type-narrowing validation functions (`parseInterceptorApprovedMessage`, `isBridgeRequest`).

All existing tests are updated to validate the new restrictions, and a new end-to-end Chrome communication smoke test is added. No security checks, assertions, or validations are removed or weakened.

# Security

This diff is a net security improvement with no exploitable vulnerabilities introduced. Here is my analysis:

## Architecture Change Summary

The changes replace `window.postMessage` (broadcast to all listeners) with a dedicated `MessageChannel` for inpage↔content-script communication. Three layers of defense are added against malicious page scripts:

1. **MessagePort isolation**: Page scripts can no longer spoof `interceptorApproved` responses via `window.postMessage`, because the inpage script's `onMessage` handler is now attached to `channel.port1.onmessage` (the dedicated port), not to `window.addEventListener('message')`. The `onWindowMessage` handler attached to `window` only processes Coinbase re-injection events — it ignores `interceptorApproved` messages entirely.

2. **Internal method filtering**: Both the inpage script (`WindowEthereumRequest` in `inpage.ts`) and the background (`handleInterceptedRequest` in `background.ts`) now block public calls to internal methods (`connected_to_signer`, `eth_accounts_reply`, `InterceptorError`, `signer_chainChanged`, `signer_reply`, `wallet_switchEthereumChain_reply`). The inpage block is a convenience rejection; the background block is the authoritative gate. Even if a page script monkey-patched the inpage check, the background — running in the isolated service worker context — would still reject the call because the message lacks the `interceptorInternalRequest: true` flag.

3. **Strict message validation**: The new `isBridgeRequest` (content-script side) and `parseInterceptorApprovedMessage` (inpage side) validators replace the old `as InterceptedRequestForward` type assertion. The content script now constructs a whitelisted message object with only the expected fields (`interceptorRequest`, `method`, `params`, `usingInterceptorWithoutSigner`, `requestId`, `interceptorInternalRequest`) rather than forwarding raw untrusted data.

4. **Port establishment race prevention**: The content script (`listenContentScript.js`) runs at `document_start` in the ISOLATED world, and `inpage.js` runs at `document_start` in the MAIN world. The inpage script is inlined and executes synchronously, posting the MessagePort before any page scripts have had a chance to register `message` listeners. The content script accepts only the first `interceptor_bridge_port` message (`inpagePort !== undefined` guard), so even if a page script somehow ran first, it could not hijack the channel after the legitimate handshake.

## No Findings

I found no path for a page script to:
- Forge an `interceptorInternalRequest: true` message (the flag can only be set by the inpage script through the dedicated MessagePort, and the content script validates the source via `isBridgeRequest`)
- Intercept or inject messages on the MessagePort (the port is entangled between the inpage script's private field and the content script running in an isolated world)
- Bypass the internal method filter (the background's `isInternalProviderMethod` check is in the service worker, unreachable from page context)
- Exploit the `parseInterceptorApprovedMessage` validator (it properly rejects messages with both `result` and `error`, invalid `type` values, or malformed error structures)

PR #1459, opened 2026-05-25
Refactor shared helpers and harden website metadata handling

KillariDev:
## Summary
- move popup/tab helpers, signer metadata, address validation, and timestamp formatting into dedicated utility modules so background and simulation code no longer depend on TSX UI helpers
- cache the live block, simulation state, and simulated block generation inside `EthereumSubscriptionService` so repeated `newHeads` subscribers reuse the same work
- move address-book migration into startup and make legacy V1/V2 migration resilient to corrupt stored data
- sanitize persisted website metadata, only background-fetch same-origin favicons for already stored websites that are missing icons, and enforce favicon size/protocol limits to avoid remote icon re-requests and oversized downloads
- add regression coverage for address-book migration, website icon sanitization/imports, favicon privacy and size handling, and subscription caching

## Testing
- `bun test`
- `bun run setup-chrome`
- `bun run typecheck`
- `bun run lint`


app/ts/background/settings.ts
```diff
@@ -168,11 +169,46 @@ export async function changeSimulationMode(changes: { simulationMode: boolean, r
...
 
-export const getWebsiteAccess = async() => (await browserStorageLocalGet('websiteAccess'))?.websiteAccess ?? []
 const websiteAccessSemaphore = new Semaphore(1)
+async function getNormalizedWebsiteAccessFromStorage() {
+	const rawWebsiteAccess = await getParsedStorageValueOrDefault('websiteAccess', [])
+	const sanitizedWebsiteAccess = sanitizeWebsiteAccess(rawWebsiteAccess)
+	return { rawWebsiteAccess, sanitizedWebsiteAccess }
+}
+
+async function persistWebsiteAccessIfChanged(previousWebsiteAccess: WebsiteAccessArray, nextWebsiteAccess: WebsiteAccessArray) {
```
MicahZoltu: Architectural concern: `getWebsiteAccess` now writes sanitized data back to `browser.storage.local` as a side effect of being called. This violates Command-Query Separation — a function named and contracted as a read operation silently mutates persistent state. Every caller becomes an implicit write path, making the system harder to reason about and test. Consider moving this sanitization to a one-time migration (following the `addressBookMigration.ts` pattern) or performing it only inside `updateWebsiteAccess`, where writes are already expected.

MicahZoltu (commented):
## CI Agent Review

Two issues identified: a Command-Query Separation violation in `getWebsiteAccess` that silently writes to storage on every read, and an event listener leak in `retrieveWebsiteDetails` where early returns bypass the required listener cleanup.

app/ts/background/iconHandler.ts
```diff
@@ -103,6 +116,13 @@ export async function retrieveWebsiteDetails(tabId: number) {
 		checkAndPrintRuntimeLastError()
 	}
 
+	const loadedTab = await safeGetTab(tabId)
+	if (websiteOrigin !== undefined) {
+		const { cachedIcon, hasStoredWebsiteAccess } = await getCachedWebsiteIcon(tabId, websiteOrigin)
+		if (cachedIcon !== undefined) return { title: loadedTab?.title, icon: cachedIcon }
```
MicahZoltu, line 129: Event listener leak: When `websiteOrigin` is provided and either a cached icon is found or no stored website access exists, the function returns early without removing the `browser.tabs.onUpdated` listener that was registered earlier. The original code always reached a `try`/`finally` block that cleaned up the listener. While the listener will self-remove on the next tab completion event, it is a resource leak that violates the intended lifecycle and could persist indefinitely in idle sessions.

MicahZoltu (commented):
## CI Agent Review

No security issues found.

MicahZoltu (approved):

PR #1460, opened 2026-05-29
Refresh home state from approved website tabs

KillariDev:
## Summary
- Fix the Home popup state mismatch where MetaMask could still be usable for signing while the UI showed `NOT CONNECTED` and no address.
- Treat a known signer account as available in the Home UI even when the cached `signerConnected` flag is stale.
- Refresh popup home data through the current tab's approved website connections so the popup can proactively resync signer accounts when its cache is empty.
- Update the popup refresh wiring and related tests to cover the extra `websiteTabConnections` argument and the stale-connection regression.

## Testing
- `bun test`
- `bun run setup-chrome`
- `bun run typecheck`
- `bun run lint`


MicahZoltu (commented):
## CI Agent Review

Architectural regression in buildHomePageUpdate: the function now performs a side effect (sending inter-process messages to re-query signer accounts), mixing query and command responsibilities. Additionally, the tab state is re-read immediately after the message is sent but before the async round-trip completes, resulting in stale data being assembled into the home page update. The changeSettings handler passes an empty Map() to silently suppress this side effect, creating a fragile implicit contract.

app/ts/background/popupMessageHandlers.ts
```diff
@@ -885,7 +886,19 @@ async function buildHomePageUpdate(
 	const tabId = await getLastKnownCurrentTabId()
```
MicahZoltu, line 907: Side effect leak and race condition: buildHomePageUpdate was previously a pure data-assembly function. The new block sends a message to approved website ports to re-query signer accounts, then immediately awaits getTabState(tabId). This mixes query and command responsibilities (the function now mutates system state), and is also a correctness bug — the async round-trip through content script → inpage → signer → background has not completed, so getTabState returns the same stale empty-accounts state.

app/ts/background/popupMessageHandlers.ts
```diff
@@ -503,7 +503,7 @@ export async function interceptorAccessChangeAddressOrRefresh(websiteTabConnecti
 export async function changeSettings(ethereum: EthereumClientService, tokenPriceService: TokenPriceService, _resetSimulationServices: ResetSimulationServices, parsedRequest: ChangeSettings, requestAbortController: AbortController | undefined) {
 	if (parsedRequest.data.useTabsInsteadOfPopup !== undefined) await setUseTabsInsteadOfPopup(parsedRequest.data.useTabsInsteadOfPopup)
 	if (parsedRequest.data.metamaskCompatibilityMode !== undefined) await setMetamaskCompatibilityMode(parsedRequest.data.metamaskCompatibilityMode)
-	return await requestNewHomeData(ethereum, tokenPriceService, requestAbortController)
+	return await requestNewHomeData(ethereum, tokenPriceService, new Map(), requestAbortController)
```
MicahZoltu: Fragile implicit contract: changeSettings passes new Map() to requestNewHomeData to suppress the signer-refresh side effect. Whether the function refreshes signer accounts depends on an accidental property of the argument (whether the map contains the current tab's connections), not on an explicit intent flag. This makes the side effect behavior invisible at call sites.

MicahZoltu (commented):
## CI Agent Review

No issues found.

MicahZoltu (approved):

PR #1462, opened 2026-05-29
Handle invalidated background ports during startup

KillariDev:
## Summary
- Attach content-script port listeners during `runtime.onConnect` before waiting for background startup so Chrome cannot invalidate the port first during startup, reload, or navigation churn.
- Centralize ignorable port lifecycle errors in background startup handling, including `Extension context invalidated`, disconnected ports, and closed message-channel cases, so these do not surface as unexpected Interceptor errors.
- Add a regression test that reproduces an invalidated connect port before listener attachment and verifies the background startup path exits cleanly without reporting an unexpected error.

## Testing
- `bun test`
- `bun run setup-chrome`
- `bun run typecheck`
- `bun run lint`


MicahZoltu (commented):
## CI Agent Review

No security issues found.

MicahZoltu (approved):

PR #1472, opened 2026-06-15
Show a structured preview and gas controls for failed transactions

KillariDev:
Failed transaction simulations now render a structured transaction preview instead of a JSON dump, matching the successful transaction UI more closely. The failed state still includes the gas limit editor so users can adjust and retry, and raw transactions fall back to a readable raw-input view when parsed fields are unavailable.


KillariDev:
image

MicahZoltu (approved):
This doesn't solve the problem of gas price being too high.  What gas price do we simulate with and can it be configured as well?

PR #1511, opened 2026-07-03
Fix bundled webextension polyfill loading

KillariDev:
## Summary
- Keep `webextension-polyfill` external when Bun bundles runtime entrypoints so the real polyfill module runs at startup and installs `globalThis.browser` before top-level background code executes.
- Add a bundler-level regression guard that fails if the vendored polyfill asset stops being reachable from runtime imports after rewriting.
- Add focused bundler test coverage for the guard and the bare-import rewrite behavior.

## Testing
- `bun test`
- `bun run setup-chrome`
- `bun run typecheck`
- `bun run lint`
- `bun run test:chrome-communication`


KillariDev:
todo: test in brave

MicahZoltu:
MV2 works in Brave.

MicahZoltu (approved):

PR #1521, opened 2026-07-07
Add `wallet_revokePermissions` support

KillariDev:
## Summary
- Adds background handling for `wallet_revokePermissions` so a site can revoke its `eth_accounts` access.
- Validates the request shape explicitly and rejects unsupported permission payloads with a parse error.
- Updates RPC/type unions so revoke responses are recognized as `null` results.
- Expands background tests to cover revoking access, already-unauthorized sites, disabled interceptor state, and invalid params.

## Testing
- `bun test`
- `bun run setup-chrome`
- `bun run typecheck`
- `bun run lint`


MicahZoltu (approved):
Is this actually a thing that sites do?  I'm mildly surprised.

KillariDev:
> Is this actually a thing that sites do? I'm mildly surprised.

sablier has it

PR #1543, opened 2026-07-21
Prevent metadata probes from masking transaction simulation results

KillariDev:
## Summary

- Prevent optional contract metadata and ABI lookups from masking the real transaction simulation result or leaving the confirmation dialog stuck on `Simulating Transaction...`.
- Bound token metadata probes and recover narrowly from probe-level JSON-RPC failures with chain-aware fallback metadata.
- Derive transaction gas limits from the execution client's fork-aware `maxUsedGas` result instead of maintaining a hardcoded Ethereum gas schedule.
- Adaptively re-simulate with increasing explicit gas limits when an RPC omits `maxUsedGas`.
- Preserve the original simulation error when optional ABI decoding fails, while continuing to propagate cancellation signals.
- Display an omitted `eth_sendTransaction` value as `0 ether` and format nonzero values as decimal Ether.

## Why

The original console error was:

```
err: intrinsic gas too low: have 11332, want 21240 (supplied gas 11332)
```

There were two related failure paths behind the confusing behavior:

1. An optional on-chain metadata probe could fail and escape the enrichment pipeline. Interceptor could then attribute that internal probe failure to the user's transaction and leave the confirmation dialog spinning.
2. `eth_simulateV1.gasUsed` is not always a safe value to reuse as a transaction gas limit. In particular, it can be lower than the peak gas required before refunds or other accounting effects. Submitting that value could make the user's transaction fail immediately with an intrinsic-gas error.

Hardcoding today's intrinsic, calldata, initcode, and creation costs would fix only the current fork and would need maintenance whenever Ethereum's gas schedule changes. Instead, this PR delegates fork-specific gas accounting to the execution client.

### Transaction gas estimation

- Prefer the optional `eth_simulateV1` `maxUsedGas` field, which represents the execution client's peak gas requirement.
- Apply a rounded-up 25% wallet-policy safety buffer to tolerate small state changes between simulation and mining. This is deliberately a wallet policy, not a protocol gas constant.
- Cap the result by the gas still available in the simulated block.
- If the node omits `maxUsedGas`, start from buffered `gasUsed` and re-run the exact simulation with progressively larger explicit limits until one succeeds or the available block gas is reached.
- Apply the same safety buffer to the verified fallback limit.
- Preserve detailed call failures and top-level JSON-RPC errors when no valid estimate can be found.
- Reuse the prepared simulation context across retries.

### Simulation and metadata error handling

- Give each independent token metadata probe a bounded gas limit so malformed estimates cannot make a probe intrinsically invalid.
- Recover only from `JsonRpcResponseError` raised by optional on-chain metadata probes, returning fallback metadata for the active chain.
- Keep unexpected errors and new-block cancellation propagation intact.
- Make simulation-error ABI lookup best effort and skip it for empty revert data.
- Pass the active confirmation abort controller through optional ABI lookup.

### Transaction value display

- Treat an omitted structured transaction value as zero.
- Render failed-transaction values using 18-decimal Ether formatting, including `0 ether` and decimal values such as `1.25 ether`.

## Testing

- `bun test` — 697 passed, 0 failed
- `bun run setup-chrome` — passed
- `bun run typecheck` — passed
- `bun run lint` — passed

Regression coverage includes:

- RPC-provided `maxUsedGas` for both simulation-backed gas-estimation entry points.
- The reported shape where `gasUsed` is `11,332` but peak gas is substantially higher.
- Adaptive retry sequences for nodes that omit `maxUsedGas`, through both estimator entry points.
- Final safety buffering and available-block-gas caps.
- Explicit gas on all metadata probes and recovery from the exact `have 11332, want 21240` probe failure.
- Propagation of unexpected errors and new-block cancellations.
- ABI lookup success, empty revert data, lookup failure, and cancellation.
- Omitted value rendering as `0 ether` and decimal Ether rendering.


MicahZoltu (changes requested):
I don't think we should be hard coding gas prices.  If we do, we fail the walk away test because those are "mutable".

MicahZoltu (approved):

PR #1570, opened 2026-08-02
Fix tab state cleanup startup races

KillariDev:
## Problem

Tab state is transient browser-session data and must be cleared when a background context starts. The existing cleanup was fire-and-forget, which created different failure modes in Firefox/MV2 and Chrome/MV3:

- In MV2, background startup and content-script connection handling could continue while stale tab-state cleanup was still running. A new connection could therefore write fresh `tabState_<id>` data that the late cleanup subsequently removed, leaving the active tab without its signer, website, or icon state.
- In MV3, activation started cleanup without passing its promise to `ExtendableEvent.waitUntil`. Chrome could suspend the service worker before cleanup finished.
- Tab-state discovery used a prefix regex. Keys such as `tabState_12_backup` were treated as live tab state, so similarly named storage could be parsed or deleted accidentally.

### Make tab-state cleanup part of background initialization

A small tab-state lifecycle module now owns browser-version-specific cleanup behavior:

- MV2 returns the cleanup promise and background startup awaits it before migrations and the rest of initialization proceed.
- MV3 attaches cleanup to the activation event with `waitUntil`, keeping the service worker alive until storage deletion settles.

This makes cleanup an explicit lifecycle dependency instead of an untracked asynchronous side effect.

### Gate content-script state writes without delaying listener registration

Waiting only inside `startup()` was not sufficient because `runtime.onConnect` can fire while startup is pending. The connection path now uses one shared initialization promise that:

1. waits for background startup, including MV2 tab-state cleanup;
2. registers the connection and writes its initial tab state;
3. exposes the initialized simulation services to request handling.

Port listeners are still attached immediately so early messages are not dropped. Stateful connection registration, incoming request processing, and disconnect cleanup all await the shared initialization promise. This prevents fresh state from racing with cleanup while preserving prompt port setup.

### Match only real tab-state keys

Tab-state discovery now uses an anchored pattern, `^tabState_[0-9]+$`. Only exact nonnegative numeric tab IDs are read or removed; suffix-bearing storage keys remain untouched.

## Regression coverage

Added tests that verify:

- exact tab-state keys are read while similarly prefixed keys are ignored;
- cleanup removes exact tab-state keys without deleting suffix-bearing storage;
- MV2 initialization does not complete before cleanup;
- MV3 activation receives and waits on the cleanup promise;
- content-script connection initialization remains blocked until background startup completes;
- the concrete `onContentScriptConnected` path keeps listener registration before the startup gate and connection registration plus the first tab-state write behind it.

## Validation

- `bun run test` — 971 passed, 0 failed across 129 files
- `bun run setup-chrome` — passed
- `bun run typecheck` — passed
- `bun run lint` — passed

`bun run test:chrome-communication` was not run because this change does not alter MV3 content-script registration, access-popup approval, injected-provider communication, or the page-to-extension messaging protocol.

## Review

The project reviewer completed three passes:

1. Score 68: identified that the real `runtime.onConnect` path could still write state before startup; fixed with the shared connection-initialization gate.
2. Score 88: requested coverage of the concrete connection entry path; added the architecture-boundary regression test.
3. Score 94: no High, Medium, or Low findings remained.


MicahZoltu (approved):
Nice to see so many bug fixes without having to report bugs!

PR #1582, opened 2026-08-04
Add linter for split prose comments

KillariDev:
## Summary
- add a TypeScript-aware linter and autofixer for prose split across consecutive line comments
- preserve structured provenance, URLs, directives, generated source, and commented-out code
- run the fixer across existing TypeScript comments and wire the check into the aggregate lint command
- add focused heuristic and package-script regression coverage

## Validation
- bun run test
- bun run setup-chrome
- bun run typecheck
- bun run lint

## Review
Project reviewer scores: 64, 74, then 92 after addressing all findings. Final pass reported no High, Medium, or Low findings.


MicahZoltu (approved):
I'm very skeptical this won't break things like diagrams in comments.

KillariDev:
> I'm very skeptical this won't break things like diagrams in comments.

probably would, but I dont think we have any

